PR #170, opened 2020-02-21
Create autoblack.yml

cclauss:
In the first commit, I only changed one file.

Look at the __second__ commit.  That was auto-generated by autoblack.yml.

@vsoch


shahzebsiddiqui:
thanks @cclauss this looks pretty good.  I will close the other PR

shahzebsiddiqui:
at @cclauss  can you please checkout from ``devel`` so we can merge this. It shouldnt be going to master so when i change this PR to merge to ``devel`` i am getting file conflicts and it will undo some of the changes.

vsoch:
I’m not sure I’m comfortable with this recipe - it gives an automated push to someone’s branch which saves the step of running locally but I’m not sure I would trust. You can do the same thing to check with black and then give the user instruction on how to fix - see the Snakemake repo for an example.

vsoch:
It also means another step of having to pull after you push every time, which seems incredibly laborious. Setting up a local hook for the user and then having the check would be more logical, and not give unneeded push access to the action.

shahzebsiddiqui:
i get your point @vsoch  the bot will add a commit after every push.  The user needs to remember to pull and rebase changes which is problematic.

The other option is to have user run black after they push which is not feasible since not everyone will use black to push code. So long as they base they feature branch of devel the commit diff which will be black will be between devel HEAD and feature branch HEAD. So i anticipate the section to be black should only be user changes. 

I also think that when user make change to their branch their subsequent push will fail if they forget to pull since black will commit on top of their changes. Which may be good thing which prevents user to overwrite changes from black. 

The downside of not having black is having the manual labor of cleaning up the entire code-base which I am not a fan of doing.

shahzebsiddiqui (changes requested):
@cclauss can you please base your PR of `devel` there are too many conflicts. Also does it make sense to only run black on ``push`` or should it be on ``pull_request``. I take if its on push it will make all changes every push. Which may be ok.

cclauss:
Closing in favor or #171

shahzebsiddiqui:
@cclauss I think @vsoch has an idea of using a hook. Probably hold off on your PR

cclauss:
Once you have your precommit hooks in place, you can have an GitHub Action auto-runs them even for new users who have not installed precommit.

https://github.com/cclauss/GitHub_Actions/blob/master/.github/workflows/pre-commit.yml

PR #171, opened 2020-02-21
Create autoblack on the devel branch

cclauss:


shahzebsiddiqui:
@vsoch and I were discussing about this in other PR you sent. I want to weigh pro/con of both your PR and what @vsoch has in mind.
I like the auto commit from black but am also worried about when user commits to their feature branch they need to pull changes locally before making any changes which is something user may not be used to.

cclauss:
Commit one was made by me and commit two was made by `github-actions` and commit three was made by me.  As long as the code in the repo is black, the action will be silent.

You could also look at reformatting __only the modified lines__ like 
https://github.com/internetarchive/openlibrary/blob/master/.travis.yml#L19

shahzebsiddiqui:
So when you made your second commit. Did you have to pull and rebase before pushing. 
If you use GitHub editor then I guess it doesn't matter since head will be updated.
I think I will need to get a feel for it.
@vsoch what your thoughts on this?

vsoch:
It doubles the number of rebases that would be needed otherwise - and it's still giving an action immediate push access, no questions ask, to any branch. I am not familiar with a GitHub editor but I don't use it, I use the command line. I'd strongly advocate for an approach that:

 - runs black with --check and (on failure) advises the user how to run black locally
 - and provides a hook to run black before commit, if desired by a user, so the code is formatted going into the PR.

Personally, I wouldn't use the hook, but I'd just run black on my own and use the test as a sanity check that it was correct. I love GitHub actions but I've heard mixed things, so generally I take the most conservative approach when possible.

shahzebsiddiqui:
Ok so I think having a black --check would be good first step. 
Most likely the check will fail and should report how to fix it. User can run black locally and push. This would give more control to user and the maintainer can request this during review process 
Certainly I am also concerned with having something broken let's say regression test fails after some black change.
I can cover this as part of contributing process. I am fine with running black locally on every push so long as code base gets fix and Travis Build is green. If something gets messed up user can reset HEAD. 
In that case @vsoch it would be good idea to not have black autopush where user may run into some issue with some regtest after black applied changes.

PR #172, opened 2020-02-21
Add black checks

vsoch:
This pull request will update the black.yml workflow to run black --check on push or pull requests for all branches. If the update is done via a pull request and the check fails, the user is sent a message to use black to update it. This should be considered separately from adding an optional hook for an interested user, which can be added in another PR.

For this first opening, I've purposefully not run black (and checked that it would fail) so if the action is working, it should fail the linting check and comment here for me to fix it.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,11 +1,36 @@
 name: Black Formatter
-on: [push, pull_request]
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches_ignore: []
+
 jobs:
-  black:
+  formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - name: Black Code Formatter
-        uses: lgeiger/black-action@master
+      - uses: actions/checkout@v2
+
+      - name: Setup black linter
+        run: conda create --quiet --name black black
+
+      - name: Lint python code
+        run: |
+          export PATH="/usr/share/miniconda/bin:$PATH"
+          source activate black
+          black --check buildtest tests/*.py
+
+      - name: Derive organization name
+        if: github.event_name == 'pull_request' && failure()
+        run: echo ::set-env name=IS_FORKED::$(if [ -z "$GITHUB_HEAD_REF" ]; then echo "no"; else echo "yes"; fi)
+        shell: bash
+
+      - name: Comment Pull Request
+        if: github.event_name == 'pull_request' && failure() && env.IS_FORKED == 'no'
+        uses: marocchino/sticky-pull-request-comment@v1.1.0
         with:
-          args: "."
+          message: 'Please format your code with [black](https://black.readthedocs.io) version 19.3b0: `black buildtest tests/*.py`.'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
--- a/buildtest/docgen/main.py
+++ b/buildtest/docgen/main.py
@@ -65,7 +65,14 @@ def introspection_cmds():
         writer(fname, out, cmd)
 
     out = run("buildtest testconfigs view tutorial/compilers/args.c.yml")
-    writer(os.path.join(docgen,"buildtest_testconfigs_view_tutorial_compilers_args.c.yml"),out,"buildtest testconfigs view tutorial/compilers/args.c.yml")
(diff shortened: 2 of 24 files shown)


vsoch:
ah okay - so the message back to comment will only be available on a PR from a branch here (and not a fork) which is logical. So - we can leave as is and the failure above would signal me to run black, or I can remove entirely. I think it's most likely to have PRs from forks, so I think we'd be okay to remove it. I'll run black locally, remove the push back comment, and that should be good. Note that the same issue is happening for the pull request greeting - I ran into this at USRSE too, so it might be best to remove for now because it's going to mean confusing failures.

shahzebsiddiqui:
So should we stick to original workflow, or abandon black all together or fix this if possible to comment in PR . PRs going to come from fork off course so it only makes sense

.github/workflows/black.yml
```diff
@@ -1,11 +1,31 @@
...
-        uses: lgeiger/black-action@master
+
+      - name: Setup black linter
+        run: conda create --quiet --name black black
+
+      - name: Lint python code
+        run: |
+          export PATH="/usr/share/miniconda/bin:$PATH"
+          source activate black
+          black --check buildtest tests/*.py
```
shahzebsiddiqui, line 23: i think ``black --check .`` should be sufficient it will autoformat all python files recursively. So no need to specify the directory paths unless we only want to black specific files. You can test this locally and you will get idea.

vsoch:
I'm doing a few tests now to see if I can determine if it's forked, this should be able to maintain the comment (if it's supported). If that doesn't work, I'll remove it.

.github/workflows/black.yml
```diff
@@ -1,11 +1,36 @@
...
+      - master
+  pull_request:
+    branches_ignore: []
+
 jobs:
-  black:
+  formatting:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
```
cclauss: v2 ?

vsoch:
okay this is what I was going for! The black linter fails, and checks if the repository is forked by way of the `GITHUB_HEAD_REF` environment variable, which is only set for forks. The black format check fails, but we skip trying to comment back. On the other hand, if it's a branch from here, it would comment back to the PR. It's unlikely, but could be useful if the project maintainers builds up, and many work on branches here.

image

.github/workflows/black.yml
vsoch: I hadn't reviewed version 2 yet (notes here) https://github.com/actions/checkout#whats-new but it seems reasonable, we might benefit from a tinier speedup of checkout! I'm happy to update this.

vsoch:
I'm going to run the linter locally now (and also update the checkout version) and push again.

.github/workflows/black.yml
vsoch: Updated.

vsoch:
And here is the successful linting job- since we passed linting the steps to derive if it's the organization branch are skipped over entirely.

image

.github/workflows/black.yml
```diff
@@ -1,11 +1,36 @@
...
+          source activate black
+          black --check buildtest tests/*.py
+
+      - name: Derive organization name
+        if: github.event_name == 'pull_request' && failure()
+        run: echo ::set-env name=IS_FORKED::$(if [ -z "$GITHUB_HEAD_REF" ]; then echo "no"; else echo "yes"; fi)
+        shell: bash
+
+      - name: Comment Pull Request
+        if: github.event_name == 'pull_request' && failure() && env.IS_FORKED == 'no'
```
shahzebsiddiqui, line 31: Don't we want IS_FORKED=yes for the comment action to trigger.

shahzebsiddiqui (approved):
looks good. Thanks @vsoch

.github/workflows/black.yml
vsoch: I think it's good to be explicit - it's a relatively simple structure now, so it makes sense to not opt in everything when we don't know what else could be there. We can change this if there is strong rationale to do so.

.github/workflows/black.yml
vsoch: If the repository is forked, then the action is going to fail with a permissions error, so we want it to only run if it's not forked.

PR #173, opened 2020-02-21
Testing customized greeing to not run on forked pull requests

vsoch:
Currently, if a user opens a pull request on a first, and it's their first PR (so the action is triggered to greet them) the greeting will fail because a forked PR does not have permission to post to the upstream repository. This update to the greeting recipe should modify the logic of greetings to derive if it's a fork, and only greet the user if it's on an organization branch. In practice, I've tended to just remove the PR greeting because it's unlikely that a new contributor will be part of the org, so that's another option. The issue greeting should work as expected.

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -1,27 +1,13 @@
-name: Greetings
+name: First Issue Greeting
 
-on: [pull_request, issues]
+on: [issues]
 
 jobs:
   greeting:
     runs-on: ubuntu-latest
     steps:
-
-      - name: Derive if this is a forked pull request
-        if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


shahzebsiddiqui:
I see the issue now maybe for what its worth. I guess it doesn't make sense to have the greeting action. I got bored and started looking for actions on github marketplace and tested this out. In that case, maybe we can just remove the greeting workflow and remove one check from all the dozen checks buildtest already has. 
I agree this greeting workflow will be rarely triggered so no point in this.
I am fine with you just removing  the greeting workflow in this PR and i can merge it.

vsoch:
The issue one works great (remember it worked for me yesterday!) and it's a nice surprise for a new user. What I'll do is remove the pull request one, which doesn't make a lot of sense / is unlikely to trigger.

shahzebsiddiqui:
I approve your change, i realize that the greeting was triggered on issue creation and i totally agree. User wont jump into creating a PR before submitting an issue. The odds are low and it doesn't really matter.

Issue #174, opened 2020-02-21
Add optional commit hook to run black


vsoch:
A user / developer that wants to run black automatically will likely appreciate having an optional hook. We could either anticipate this and add right away, or wait for someone to explicitly ask for it. If anyone is reading this issue and would like this supported, please share your thoughts!

bkmgit:
Assuming this is for Python code formatting (https://github.com/psf/black), for buildtest itself it would be nice to have this run automatically when people make a pull request to update the code.

vsoch:
We discussed this, and any kind of automated commit or PR would require extra effort on the part of maintainers to review, or in the case of a commit, a need to rebase with every push. It’s also best to take a conservative approach with GitHub actions as there is some risk in giving the bot push access to branches. As an intermediate solution, a hook to run black pre commit would satisfy all parties - the code would be formatted automatically without need for rebase, and it would be installable by those that want it (for example my preference is to not use hooks). The code is tested either way to ensure the formatting is there.

This issue is to get feedback on adding that hook, which is not done yet.

bkmgit:
Not sure if one can run test for Black formatting (black --check) automatically upon trying to merge? If ok, code can be merged, if not ok, committer will need to format code. This saves reviewing time.

Tensorflow has documentation on Style guides:
https://www.tensorflow.org/community/contribute/code_style
There are a number of automated checks that are run before contributions are manually reviewed.

vsoch:
Yes we already added back —check to run before merge is allowed!

shahzebsiddiqui:
@bkmgit I like black, so far there are no issues, the ``black --check`` is in place. I had tried this initially when i first experimented with black using GitHub action. There was a PR related to auto-black on PR by @cclauss https://github.com/HPC-buildtest/buildtest-framework/pull/171 that did what I wanted, but then we realize the auto-black leads to undesired consequences as mentioned by @vsoch. My main concern is 

- User has to pull & rebase after every commit that leads to autoblack
- Black causes some code-base or regression test to break
- Resetting HEAD and making push can lead to auto-black trigger.

So i think the hook option would make sense. I dont have much experience, looking at it are we talking about this link for GitHooks https://githooks.com/

I see there is a pre-commit  hook https://pre-commit.com/. Not sure which ones to look at. @vsoch which one did you have in mind.

cclauss:
[...]

PR #179, opened 2020-02-23
Add pre commit hook for black

vsoch:
This is a first shot to add a pre-commit hook to run for the repository and run black to address #174. I have also added instructions to the docs for "installing it." I have installed it locally, and can test it for a few days to assess if it's working properly (the push to this branch was okay but I hadn't edited much python code).

### DIFF
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -1,0 +1,67 @@
+#!/bin/sh
+#
+# This pre-commit hook will run black linting before finishing the commit.
+# If the linting has any failures (indicating a functional error) we don't commit.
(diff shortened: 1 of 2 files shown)


vsoch:
okay, I originally left the line for trailing white spaces, but I think that's probably overkill for now for the pre-commit hook (unless you want it?) Have you been maintaining that the docs don't have white spaces after lines? We could add in a linter to fix this (and not just check) if it's something important.

shahzebsiddiqui (approved):
Looks good to me. I will install it locally and test it out as well. If it looks good, we can officially put this as part of the contribution workflow.

PR #180, opened 2020-02-24
replacing toolkit/suite with site

vsoch:
This is a first round of changes to replace toolkit/suite with site, and then have references to the tutorials be changed to reference the [HPC-buildtest/tutorials](https://github.com/HPC-buildtest/tutorials) repository. Note that there are white spaces after some of the docs updates I made - and I wasn't sure if we should add some kind of check or lint for this with the pre-commit #179 to ensure that it gets pushed cleaned up. If so, I can update #179 and here without the spaces at end of newlines.


shahzebsiddiqui (approved):
Overall it looks good. My one recommendation is that the changes made to files in ``docs/docgen/*.txt`` were probably done manually and you didn't run the script ``buildtest/docgen/main.py``.

I intended to not run this script right now because some of the commands may not work due to some recent changes to the core functionality and also the recent restructure of directory.

vsoch:
I was aware of the automated generation, but didn't know how to do it, and I suspected that some of the commands would fail, so although it (yes!) was done manually, I wanted to still update the paths.

shahzebsiddiqui:
I could tell it was manual update because if it was run through the automated script, the output of test will be different because the ``buildtest build --config`` has changed its behavior to build and run test automatically. That is the current state of ``devel`` at the moment. 

For your reference the autodoc is covered in https://buildtest.readthedocs.io/en/latest/contributing/build_documentation.html#automate-documentation-examples.

It's not the best way of getting things done but it works for most examples

PR #181, opened 2020-02-24
second addition of module API.

shahzebsiddiqui:
replace subprocess.Popen with BuildTestCommand
raise TypeError in Module and ModuleCollection class when invalid
arguments passed for module or collection name.
Add method save(), describe() that implement saving and showing module
collections.
Change syntax of module load command to use command chaining (&&) which
is important when loading multiple modules and testing for return code.
Add docstring for all methods.
Add one example python to demonstrate how to save and show module
collections.


shahzebsiddiqui:
@vsoch  Let me know your thoughts.

vsoch:
I'll make some time this week to review this carefully!

shahzebsiddiqui:
Take your time @vsoch and feel free to test on your end.  Eventually I'd want to document and add regression test once the API is stable and the scripts need to live somewhere in repo to reference.  I couldn't think of a better name other than ``examples``.

buildtest/modules/util.py
```diff
@@ -1,48 +1,156 @@
...
+     :rtype: list
+     """
+
     collections = "module -t savelist"
-    ret = subprocess.Popen(
-        collections, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-    )
-    out = ret.communicate()[1]
-    tmp = out.decode("utf-8").split()
+    cmd = BuildTestCommand()
```
vsoch: I'm concerned that here we have shell set to True, which I think is considered bad practice:

```bash
        self.ret = subprocess.Popen(
            cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
        )
```
I think there are other ways to handle this (e.g., for sregistry I use Popen, see [here](https://github.com/singularityhub/sregistry-cli/blob/60e2e48d0fbf650f459d99f7b1d28f0c1e44069d/sregistry/utils/terminal.py#L126). Overall, having a consistent function to execute commands is great - but I don't think we want to use shell=True. I'll open an issue to refactor this for a future PR.

vsoch: https://github.com/HPC-buildtest/buildtest-framework/issues/182

buildtest/modules/util.py
```diff
@@ -1,48 +1,156 @@
...
+    cmd.execute(collections)
+    # The output is captured in STDERR stream.
+    out = cmd.get_error().split()
```
vsoch: I very much like the design of this command object - that we can execute, get_error, I'd even add another function to make it easy to print (preview) what command is going to be issued - this would be helpful to hook up to a debug mode.

buildtest/modules/util.py
```diff
@@ -1,48 +1,156 @@
...
 
-    return tmp
+    return out
 
 
 class ModuleCollection:
+    """Class declaration of ModuleCollection."""
+
     def __init__(self, collection):
+        """Initializer method of ModuleCollection class.
```
vsoch: Can you walk me through the difference in just a module / module collection? I'm guessing it's an organized group of modules - is this something created by buildtest, or associated with Lmod? How does this link up to a user request to run just one recipe file:

```bash
buildtest build -c /path/to/mpi.yml
```
Versus an entire folder?

```bash
buildtest build -d /path/to/collection
```
Couldn't it be the case that there is a recipe (defined in a collection) that you would target the individual file for? Or maybe what I'm asking - what's the distinction between a collection and just a folder / subset of recipes, and given that we can organize with folders, why do we need collections?

buildtest/modules/util.py
```diff
@@ -1,48 +1,156 @@
...
         self.collection = collection
         self.module_cmd = f"module restore {self.collection}"
 
     def test_collection(self):
-        ret = subprocess.Popen(
-            self.module_cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-        )
-        ret.communicate()
-        return ret.returncode
+        """Test the module collection by running ``module restore <collection>``.
```
vsoch: I'm wondering if the ModuleCollection should have similar functions to Module, just of course for the collection? Imagine if we had a list, with a Module and a ModuleCollection:

```python
modules = [Module(), ModuleCollection()]
```

With common methods, we could loop through:

```python
for moduleset in modules:
    moduleset.test()
```
But as it currently is, we'd have to do:

```bash
for moduleset in modules:
    if isinstance(moduleset, ModuleCollection):
        moduleset.testt_collection()
    elif isinstance(moduleset, Module):
        moduleset.test_module()
```
Also, if a module collection truly is a collection of modules, why isn't the Module subclass part of the ModuleCollection? For example, let's say I can create a ModuleCollection and add modules based on a uniqueid:

```
collection = ModuleCollection()
moduleExample = Module(...)
collection["example"] = moduleExample
```
And then the module collection has an internal dict that can be looped over / interacted with in various ways to test.

```bash
def test_collection(self):
    for name, module in self.modules.items():
        print("Testing %s" % name)
        module.test_module()
```
And you could even try subclassing a dict if you wanted the ModuleCollection to act like one. Or have it be a variable like in the example above.

buildtest/modules/util.py
```diff
@@ -1,48 +1,156 @@
...
+
+        cmd = BuildTestCommand()
+        cmd.execute(self.module_cmd)
+        return cmd.returnCode()
 
     def get_command(self):
+        """ Get the module command used to restore a collection
+
+        :return: Return the actual command to restore a collection.
+        :rtype: str
```
vsoch: What is an example of this command? I'm not super familiar with Lmod / collections.

examples/saving_collection.py
```diff
@@ -0,0 +1,17 @@
+import os
+import sys
+
+sys.path.insert(0, os.getenv("BUILDTEST_ROOT"))
```
vsoch: We probably don't need this envar anymore, if it's installed on the Python path in site-packages.

examples/saving_collection.py
```diff
@@ -0,0 +1,17 @@
...
+
+from buildtest.modules.util import Module
+
+a = Module(["zlib"])
```
vsoch: Why is this required to be a list? Intuitively, I'd just provide a string.

examples/saving_collection.py
```diff
@@ -0,0 +1,17 @@
...
+# save module collection with name "zlib"
+a.save("zlib")
+# show module collection "zlib"
+a.describe("zlib")
```
vsoch, line 7: I think what we need (actually, what I need) is a very small bash script example that shows me how to set up a small local site with my own modules. And then provide instructions to run these commands. This would give me the entire picture:

 - being on a cluster with Lmod
 - creating a dummy module
 - interacting with buildtest to test it

Without any of the bullets above, a new user like myself won't have the full picture.

vsoch (commented):
Okay - so my review is only going to be superficial, because I don't think we are at a stage where things are properly working as we want, and I haven't experienced a build yet. So I'll review superficially, and do most of the more detailed testing with next step changes.

buildtest/modules/util.py
shahzebsiddiqui: modules and module collection are means to load the modules which are going to be used when defining the YAML schema. Currently the buildtest allowed few ways to load modules, two of them are the module collection and buildtest own module collection. 

See https://buildtest.readthedocs.io/en/latest/build_subcommand/module_collections.html

The schema needs to be changed, before user must specify ``buildtest build -c <config> -co <collection>`` or ``buildtest build -c <config> -mc <buildtest-collection-ID>`` to allow loading modules. If none are specified then active modules are loaded.

The idea was having a Module API which would allow a mechanism to interact with module system. 

If you are new to Lmod user collection check out https://lmod.readthedocs.io/en/latest/010_user.html#user-collections. 

Idea is to allow buildtest to write tests where user can specify how to load modules. Either user can create a Lmod user collection. Or use the buildtest module collection see https://buildtest.readthedocs.io/en/latest/module_operation.html#module-collection-operation-buildtest-module-collection which was implemented to allow an arbitrary size of module collections. The idea was to use collection ID instead of unique names since user who use collection names need to write unique names and it will cause issue when there are alot of user collection.

The ``ModuleCollection`` class is provided to interact with module collection. The ``Module`` class is the main class that user can interact with modules.


I have few examples of the YAML schema in this wiki https://github.com/HPC-buildtest/buildtest-framework/wiki/Yaml-Schema-Draft that i was thinkin about the new schema. This is not implemented.

Hope that answers few questions and maybe leads you to more questions :)

buildtest/modules/util.py
shahzebsiddiqui: let me try to add a method to ``BuildTestCommand()`` to get that information.

buildtest/modules/util.py
shahzebsiddiqui: The ``ModuleCollection()`` at the moment, was designed to take a collection name so that we can retrieve the command ``module restore <collection>`` this command needs to be in the test script  to load a module collection when buildtest generates a script. 

When i was thinking about module specification in YAML file was to avoid hard coding the module logic in tests as it gets too confusing and complex for framework to handle. So i thought let Lmod do most of the work. 

If ``ModuleCollection()`` is a subclass of ``Module()`` I am not sure how user will interact with this class. 

The ``Module().save()`` and ``Module().describe()`` are two methods that save and show module collection. It is assumed that user can specify which modules to load in the class that get saved. The ``ModuleCollection()`` is assumed to work only with collection that are saved. For instance listing all module collection is just one part of it.

buildtest/modules/util.py
shahzebsiddiqui: See https://github.com/shahzebsiddiqui/buildtest-framework/blob/module_api_v2.0/examples/collection.py#L14 this will return the module collection command. This is useful during test creation. Look at https://buildtest.readthedocs.io/en/latest/build_subcommand/module_collections.html#building-test-with-lmod-user-collection-buildtest-build-collection this example shows a CUDA collection loaded in test script to test a saxpy code. This is a way for users to save module collection and specify them in YAML file without having to worry about actual module names. 

Once you read https://lmod.readthedocs.io/en/latest/010_user.html#user-collections you will get idea why user collection are important.

examples/saving_collection.py
shahzebsiddiqui: My thought was around dealing with module load command

```
self.module_load_cmd = [f"module load {x} && " for x in self.modules]
```

If user specify 
```
a = Module(["zlib","GCCcore","CUDA/10.0"]
```

This should generate a module load command as follows:

```
module load zlib && module load GCCcore && module load CUDA/10.0
```

It was easier to process a list since i need to run module load against all modules. Note that running ``module load zlib GCCcore CUDA/10.0`` is also valid but it does not guarantee all the modules are loaded. It is important when it comes to checking return code of the module load command.

You can try this in your system and see output of ``echo $?``  with command chaining like how i propose version doing ``module load <a> <b> <c>``.

Initialliy i had done ``module load <a>; module load <b>; module load <c>`` and this didnt work because it did not guarantee all modules were loaded since exit code will be of the last command.

Also in order for me to implement the ``Module().save()`` method i needed modules to be loaded in order to save them.

So running ``module load zlib && module save zlib`` worked when passing this to ``BuildTestCommand``

buildtest/modules/util.py
vsoch: Interesting - my intuition is that the typical sysadmin would store module "collections" on the filesystem in folders, so this seems like a feature that would be afforded by just having folders instead, and then possibly using an include / exclude filter. 

But actually, since it's integrated with Lmod, this is probably something that is used and should be represented here. I can get a better sense when I am able to create my own little collection, and set of modules! I'll also ask my group about how they use collections. Either way, I think we should be careful with the UI, because there are now several different ways to select some subset of modules (and too many might be confusing for the UI):

 - select a config directly with -c / --config (collections not relevant)
 - select a directory with -d (with some include/exclude filters) - are collections relevant here? They probably shouldn't be.
 - select one or more collections
 - don't select anything, but discover all defined in the config.

Anyway, I am +1 for having collections given that they are a big part of Lmod - I'll need to refresh my learning about how they work to best discuss the implementation.

buildtest/modules/util.py
vsoch: I don't mean that ModuleCollection should be a subclass, but rather that ModuleCollection should ultimately load Module objects into it. It's exactly what it sounds like - a collection of actual Modules. This way, when the Module function is developed more, all of that functionality will be available with the ModuleCollection (because it includes Modules somewhere).

examples/saving_collection.py
vsoch: Sorry, what I mean is to allow:

```python
a = Module("zlib")
```
And then in the init, do something like:

```python
if not isinstance(name, list):
    name = [name]
```
That way, both of these will work:

```python
a = Module("zlib")
b = Module(["zlib", "mpi"])
```
Because it's not super intuitive to provide a list with one item to a function instantiation. The internal stuffs that you detailed looks good to me :)

buildtest/modules/util.py
shahzebsiddiqui: Lmod collection has been around for quite some time. Load a bunch of modules and save them into a collection, then purge your modules and you can restore all your modules using ``module restore <collection>``. 

I think we may be confusing some terminology, the ``--config`` and supposedly ``-d`` (not implemented) are options to deal with which YAML file to process. Inside the YAML you can specify key and one of them could be how to load modules. 

In the backend, buildtest needs to call the Module API to load the appropriate module. 

For example this example is showing that we can use a collection name ``GCC`` to restore the module. Assuming that user has this collection saved. The collections are saved in $HOME/.lmod.d/ 

```
# module restore example for C
hello_ex1:
  type: singlesource
  description: "C++ hello world example"
  compiler: gnu
  source: hello.cpp
  scheduler: local
  # module collection to restore
  moduleload:
    collection:
      name:
       - GCC
  cxxflags:
    - "-O2"
```

There can be other examples such as ones below

```
# module load example with one version
hello_ex1:
  type: singlesource
  description: "C hello world example1"
  compiler: gnu
  source: hello.c
  scheduler: local
  moduleload:
    key:
      name: GCC
      version: ["6.4.0"]
  cflags:
    - "-O1"

# single test with all version
hello_ex2:
  type: singlesource
  description: "C hello world example2"
  compiler: gnu
  source: hello.c
  scheduler: local
  moduleload:
    key:
      name: GCC
      version: ["*"]
  cflags:
    - "-O1"
```


The second example where ``version: ["*"]`` is doing a test against all GCC modules. This is not implemented in the ModuleAPI and it is pretty simple to get this since I have most of the methods implemented already.

examples/saving_collection.py
shahzebsiddiqui: got it. Let me fix this i think i get what you mean.

shahzebsiddiqui:
@vsoch here is some output from the examples run locally

```
ssi29@lsfportal1> python examples/collection.py
Skipping module tree /usr/share/modulefiles/Linux because path does not exist
Skipping module tree /usr/share/modulefiles/Core because path does not exist
module restore Python
[DEBUG] Executing module command: module restore Python
[DEBUG] Return Code: 0
Successfully loaded collection: Python
module restore default
[DEBUG] Executing module command: module restore default
[DEBUG] Return Code: 0
Successfully loaded collection: default
module restore zlib
[DEBUG] Executing module command: module restore zlib
[DEBUG] Return Code: 0
Successfully loaded collection: zlib
module restore invalid_collection
[DEBUG] Executing module command: module restore invalid_collection
[DEBUG] Return Code: 1
Failed to load collection: invalid_collection with exit status: 1
[DEBUG] Executing module command: module restore Python
[DEBUG] Return Code: 0
  File "examples/collection.py", line 26, in <module>
    ModuleCollection(1)
  File "/mxg-hpc/users/ssi29/buildtest-framework/buildtest/modules/util.py", line 37, in __init__
    raise TypeError(f"Type Error: {collection} is not of type string")
TypeError: Type Error: 1 is not of type string

```

```
ssi29@lsfportal1> python examples/moduletest.py
Skipping module tree /usr/share/modulefiles/Linux because path does not exist
Skipping module tree /usr/share/modulefiles/Core because path does not exist
The following modules:  ['GCCcore', 'Python'] were loaded successfully


Command Executed: module purge && module load GCCcore &&  module load Python
Failed to load modules: ['GCCcore', 'invalid']
Command Executed: module purge && module load GCCcore &&  module load invalid
return code: 1
module load GCCcore &&  module load Python
module --force purge &&  module load GCCcore &&  module load Python
[DEBUG] Executing module command: module purge && module load Anaconda3/5.3.0 &&  module load M4/1.4.17
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load Anaconda3/5.3.0 M4/1.4.17
[DEBUG] Return Code: 0

```

```
ssi29@lsfportal1> python examples/saving_collection.py
Skipping module tree /usr/share/modulefiles/Linux because path does not exist
Skipping module tree /usr/share/modulefiles/Core because path does not exist
[DEBUG] Executing module command: module purge && module load zlib   && module save zlib
[DEBUG] Return Code: 0
Saving modules ['zlib'] to module collection name: zlib
Saved current collection of modules to: "zlib"


[DEBUG] Executing module command: module describe zlib
[DEBUG] Return Code: 0
Collection "zlib" contains:
   1) zlib

[DEBUG] Executing module command: module describe default
[DEBUG] Return Code: 0
Collection "default" contains:
   1) bzip2/1.0.8-GCCcore-8.3.0    4) libreadline/8.0-GCCcore-8.3.0    7) XZ/5.2.4-GCCcore-8.3.0        10) Python
   2) zlib/1.2.11-GCCcore-8.3.0    5) Tcl/8.6.9-GCCcore-8.3.0          8) GMP/6.1.2-GCCcore-8.3.0       11) CUDA
   3) ncurses/6.1-GCCcore-8.3.0    6) SQLite/3.29.0-GCCcore-8.3.0      9) libffi/3.2.1-GCCcore-8.3.0    12) GCCcore

  File "examples/saving_collection.py", line 17, in <module>
    a.describe(["zlib"])
  File "/mxg-hpc/users/ssi29/buildtest-framework/buildtest/modules/util.py", line 175, in describe
    raise TypeError(f"Type Error: {collection} is not of type string")
TypeError: Type Error: ['zlib'] is not of type string

```

buildtest/modules/util.py
```diff
@@ -59,24 +70,26 @@ def get_command(self):
...
         """
+        self.debug = debug
+        self.modules = modules
 
-        # raise TypeError exception if collection is not a string type since that is required
-        # when working with module collection
+        # convert input to list if not specified already.
         if not isinstance(modules, list):
-            raise TypeError(f"Type Error: {collection} is not of type list")
+            self.modules = [modules]
```
vsoch: I think you would still need to do:

```
if not isinstance(modules, list):
      modules = [modules]
self.modules = modules
```
No matter what is input (list or string) you want self.modules set.

buildtest/modules/util.py
```diff
@@ -149,8 +177,14 @@ def describe(self, collection="default"):
         module_describe_cmd = f"module describe {collection}"
         cmd = BuildTestCommand()
         cmd.execute(module_describe_cmd)
+        ret = cmd.returnCode()
 
         # For some odd reason, the output of module save gets passed to STDERR stream instead of STDOUT.
         out = cmd.get_error()
 
+        # print executed command for debugging
+        if self.debug:
```
vsoch: This is a lot of redundant code - can this be a shared function for the class? I still think there could be a very basic base class for ModuleCollection and Module, or at least a shared function added to both to essentially run a command and print debug output (or not).

examples/saving_collection.py
```diff
@@ -1,11 +1,11 @@
 import os
 import sys
 
-sys.path.insert(0, os.getenv("BUILDTEST_ROOT"))
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
vsoch: Why is this line needed?

shahzebsiddiqui: when i remove the line i run into import error

```
ssi29@lsfportal1> python examples/moduletest.py
  File "examples/moduletest.py", line 6, in <module>
    from buildtest.modules.util import Module
ModuleNotFoundError: No module named 'buildtest.modules'
```

Pretty sure its the packaging or something else i couldnt figure out
```
ssi29@lsfportal1> pip show -f buildtest-framework
Name: buildtest-framework
Version: 0.7.7
Summary: HPC Application Testing Framework
Home-page: https://github.com/HPC-buildtest/buildtest-framework
Author: Shahzeb Siddiqui
Author-email: shahzebmsiddiqui@gmail.com
License: GPLv2
Location: /u/users/ssi29/.local/lib/python3.7/site-packages/buildtest_framework-0.7.7-py3.7.egg
Requires: argcomplete, PyYAML, distro, termcolor
Required-by:
Files:
Cannot locate installed-files.txt
```

Looking into my egg i dont have the updated files such as ``buildtest.modules``

```
ssi29@lsfportal1> ll /u/users/ssi29/.local/lib/python3.7/site-packages/buildtest_framework-0.7.7-py3.7.egg/buildtest/
total 24
drwxr-xr-x 3 ssi29 hks 4096 Feb 22 17:13 docgen
-rw-r--r-- 1 ssi29 hks   60 Feb 22 17:13 __init__.py
-rw-r--r-- 1 ssi29 hks  739 Feb 22 17:13 main.py
drwxr-xr-x 2 ssi29 hks 4096 Feb 22 17:13 __pycache__
drwxr-xr-x 2 ssi29 hks 4096 Feb 22 17:13 scripts
-rw-r--r-- 1 ssi29 hks 1964 Feb 22 17:13 settings.yml
drwxr-xr-x 6 ssi29 hks 4096 Feb 22 17:13 tools

```

Hence i added that line so that it works with my git repo. I know i am missing something in my environment

buildtest/modules/util.py
shahzebsiddiqui: i tried what you did, and it turns out this breaks badly. When running ``test_modules()`` you get the following output which is not right. For now it needs to be before, assumption was if its a list it is fine, otherwise it converts to a list. This logic can be revisited. I personally like throwing the ``TypeError()`` to be very explicit to on what type of argument type can be passed into `Modules` class.

```
[DEBUG] Executing module command: module purge && module load B &&  module load i &&  module load s &&  module load o &&  module load n &&  module load / &&  module load 3 &&  module load . &&  module load 0 &&  module load . &&  module load 4
```

shahzebsiddiqui:
@vsoch my latest commit is my initial implementation of Spider class. I have an example test ``examples/spider_example.py`` that illustrates how this could work and certainly this is not perfect.

To make any sense of Spider i'd refer to reading https://buildtest.readthedocs.io/en/devel/spider.html which will start to make some sense on how i implemented some of the features.

```
ssi29@lsfportal1> python examples/spider_example.py
Skipping module tree /usr/share/modulefiles/Linux because path does not exist
Skipping module tree /usr/share/modulefiles/Core because path does not exist
Skipping module tree /usr/share/modulefiles/Linux because path does not exist
Skipping module tree /usr/share/modulefiles/Core because path does not exist
Module Trees: ['/mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Core', '/mxg-hpc/users/ssi29/spack/modules/linux-rhel7-x86_64/Core', '/mxg-hpc/users/ssi29/easybuild/modules/all', '/etc/modulefiles', '/usr/share/modulefiles', '/usr/share/lmod/lmod/modulefiles/Core']
Unique Software: ['Anaconda3', 'Autoconf', 'Automake', 'Autotools', 'Bison', 'CMake', 'CUDA', 'Clang', 'EasyBuild', 'FFTW', 'GCC', 'GCCcore', 'GMP', 'M4', 'OSU-Micro-Benchmarks', 'OpenBLAS', 'OpenMPI', 'PyCharm', 'Python', 'SQLite', 'ScaLAPACK', 'Tcl', 'XZ', 'binutils', 'bzip2', 'diffutils', 'flex', 'foss', 'gdbm', 'gettext', 'gompi', 'help2man', 'hwloc', 'libffi', 'libiconv', 'libpciaccess', 'libreadline', 'libsigsegv', 'libtool', 'libxml2', 'lmod', 'm4', 'ncurses', 'numactl', 'pkgconf', 'readline', 'settarg', 'tar', 'util-macros', 'xorg-macros', 'xz', 'zlib']


Parent Modules: ['GCCcore/8.1.0', 'GCCcore/9.2.0']


GCC Versions: ['6.4.0-2.28', '7.1.0-2.28', '9.2.0-2.32', '8.1.0-2.30', '8.1.0-2.30', '8.3.0', '7.4.0-2.31.1']


Module Trees: /mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Core
Unique Software: ['Bison', 'GCC', 'GCCcore', 'M4', 'binutils', 'flex', 'help2man', 'zlib']
Module Names: ['Bison/3.0.4', 'Bison/3.0.4', 'Bison/3.0.5', 'Bison/3.3.2', 'Bison/3.3.2', 'GCC/8.1.0-2.30', 'GCC/9.2.0-2.32', 'GCCcore/8.1.0', 'GCCcore/9.2.0', 'M4/1.4.17', 'M4/1.4.18', 'M4/1.4.18', 'M4/1.4.18', 'binutils/2.30', 'binutils/2.30', 'binutils/2.32', 'binutils/2.32', 'flex/2.6.4', 'flex/2.6.4', 'flex/2.6.4', 'help2man/1.47.10', 'help2man/1.47.4', 'help2man/1.47.6', 'zlib/1.2.11', 'zlib/1.2.11', 'zlib/1.2.11']
[DEBUG] Executing module command: module purge && module load Bison/3.0.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load Bison/3.0.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load Bison/3.0.5
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load Bison/3.3.2
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load Bison/3.3.2
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load GCC/8.1.0-2.30
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load GCC/9.2.0-2.32
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load GCCcore/8.1.0
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load GCCcore/9.2.0
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load M4/1.4.17
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load M4/1.4.18
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load M4/1.4.18
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load M4/1.4.18
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load binutils/2.30
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load binutils/2.30
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load binutils/2.32
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load binutils/2.32
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load flex/2.6.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load flex/2.6.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load flex/2.6.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load help2man/1.47.10
[DEBUG] Return Code: 1
[DEBUG] Executing module command: module purge && module load help2man/1.47.4
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load help2man/1.47.6
[DEBUG] Return Code: 1
[DEBUG] Executing module command: module purge && module load zlib/1.2.11
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load zlib/1.2.11
[DEBUG] Return Code: 0
[DEBUG] Executing module command: module purge && module load zlib/1.2.11
[DEBUG] Return Code: 0

```

If you notice such as this line 
```
[DEBUG] Executing module command: module purge && module load help2man/1.47.6
[DEBUG] Return Code: 1
```

This fails because i have not loaded parent modules before loading requested module. In my spider record i have the following

```
        "/mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Compiler/GCCcore/8.1.0/help2man/1.47.6.lua": {
            "pV": "000000001.000000047.000000006.*zfinal",
            "Description": "help2man produces simple manual pages from the '--help' and '--version' output of other commands.",
            "URL": "https://www.gnu.org/software/help2man/",
            "pathA": {
                "/mxg-hpc/users/ssi29/easybuild-HMNS/software/help2man/1.47.6-GCCcore-8.1.0/bin": 1
            },
            "whatis": [
                "Description: help2man produces simple manual pages from the '--help' and '--version' output of other commands.",
                "Homepage: https://www.gnu.org/software/help2man/",
                "URL: https://www.gnu.org/software/help2man/"
            ],
            "wV": "000000001.000000047.000000006.*zfinal",
            "Version": "1.47.6",
            "parentAA": [
                [
                    "GCCcore/8.1.0"
                ]
            ],
            "hidden": false,
            "fullName": "help2man/1.47.6",
            "help": "\nDescription\n===========\nhelp2man produces simple manual pages from the '--help' and '--version' output of other commands.\n\n\nMore information\n================\n - Homepage: https://www.gnu.org/software/help2man/\n",
            "lpathA": {
                "/mxg-hpc/users/ssi29/easybuild-HMNS/software/help2man/1.47.6-GCCcore-8.1.0/lib": 1
            }
        },

``` 

Note the ``parentAA`` is nested-list  means this module can be loaded with any one set of parent module combination.

You can see this works

```
ssi29@lsfportal1> module purge && module load GCCcore/8.1.0 && module load help2man/1.47.6
ssi29@lsfportal1> echo $?
0
ssi29@lsfportal1> module -t list
GCCcore/8.1.0
help2man/1.47.6
```

Note i have solved this problem in ``buildtest module loadtest``.

buildtest/modules/util.py
shahzebsiddiqui: I think it makes sense to have a shared method for debug purpose.

It will require some code refactor and changes to example program. 

Id be open to changing the ModuleCollection class. I think this PR is already big enough.

I'd eventually like to have this part of regression test. In subsequent PRs.

Also Travis builds a mini eb stack for testing some of these features. That's a whole separate discussion

buildtest/modules/util.py
shahzebsiddiqui: I'd be open to refactor ModuleCollection class in subsequent PRs. The PR is kinda big and complex. Could we do this in subsequent PRs.

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+        """Initialize method for Spider class.
+
+        :param tree: User can specify a module tree to query from spider.
+        :type tree: str
+        """
+
+        #  if user specifies a tree, then run spider command otherwise read from configuration file.
+        if tree:
+            self.tree = tree
+            spider_cmd = f"{os.getenv('LMOD_DIR')}/spider -o spider-json {self.tree}"
```
vsoch: If this isn't defined, it's going to return as None, and then the command will be `None/spider` and should not be run. We need to have a check here that the LMOD_DIR envar is defined, and that it's a path that is valid before executing anything to the system.

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+
+            out = subprocess.check_output(spider_cmd, shell=True).decode("utf-8")
```
vsoch: Keeping in mind we are going to refactor BuidlTestCommand, why not use it here? And then (in the future) when we refactor, we can remove the shell=True.

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+            self.spider_content = json.loads(out)
```
vsoch: I could see issues with loading this here - it's not imminent to put a try except, but let's keep it in mind.

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+
+        else:
+            with open(BUILDTEST_SPIDER_FILE, "r") as fd:
```
vsoch: This assumes that the file exists, which I don't think we can safely do.

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+                self.spider_content = json.load(fd)
+
+            content = load_configuration(BUILDTEST_CONFIG_FILE)
```
vsoch: Is this guaranteed to load something, and is the something guaranteed to have `BUILDTEST_MODULEPATH`? The load_configuration should likely exit cleanly if anything weird happens and then the self.tree should get like:

```
self.tree = content.get("BUILDTEST_MODULEPATH", [])
```

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+        return self.tree
+
+    def get_unique_software(self):
+        """Return all keys from spider. This is the unique software names.
+
+        :return: return sorted list of all spider keys.
+        :rtype: list
+        """
+
+        return sorted(list(self.spider_content.keys()))
```
vsoch: You don't need the list here, the sorted keys already is a list. Here is an example:

```
test = {1:1, 2:2, 3:3}
sorted(test)                                                                                                            
[1, 2, 3]

type(sorted(test))                                                                                                      
list
```

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def get_modules(self):
+        """Retrieve all module names from all module tree.
+
+        :return: returns  a sorted list of all full canonical module name from all spider records.
+        :rtype: list
```
vsoch: Now that I see this - should this (and or the get_unique_software) be an iterator? How / what is the context in which this is used?

buildtest/modules/spider.py
```diff
@@ -0,0 +1,103 @@
...
+        # we only care about unique modules. parentAA is bound to have duplicate modules.
+        parent_set = set()
+
+        for module in self.get_unique_software():
+            for mpath in self.spider_content[module].keys():
+                if "parentAA" in self.spider_content[module][mpath].keys():
+                    for parent_comb in self.spider_content[module][mpath]["parentAA"]:
+                        [parent_set.add(parent_module) for parent_module in parent_comb]
+
+        return sorted(list(parent_set))
```
vsoch: Another sorted ( list , you can remove the list I think.

examples/spider_example.py
```diff
@@ -0,0 +1,36 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
vsoch: We shouldn't need this if buildtest has been installed.

examples/spider_example.py
```diff
@@ -0,0 +1,36 @@
...
+
+from buildtest.modules.spider import Spider
+from buildtest.modules.util import Module
+
```
vsoch, line 3: An organizational question - why is Module imported from util? It would be more intuitive to have like:

```
from buildtest.modules import Module
```
But you don't want it accessible via the init?

buildtest/modules/spider.py
shahzebsiddiqui: I did try using `BuildTestCommand` i had trouble getting content into a dictionary using `json.loads()` when using `get_output()` method so i gave up and used `subprocess` for now. I know this needed to be fixed.

buildtest/modules/spider.py
shahzebsiddiqui: i was not sure about the ``else`` logic. We could simply use `MODULEPATH` instead of this file.

examples/spider_example.py
shahzebsiddiqui: this makes sense to me if the file is `buildtest/modules.py` i can do that. I was not sure if i will be implementing other class so for instance `buildtest/modules/spider.py` was added to avoid having one massive file with all classes and methods.

I can make the file change if it makes sense.

buildtest/modules/spider.py
shahzebsiddiqui: This is used in context of automating ``module load test``. 

Also for some reason the list is not sorted when testing the output

```
Unique Software: ['gompi', 'numactl', 'GCCcore', 'flex', 'ncurses', 'OpenMPI', 'Anaconda3', 'bzip2', 'foss', 'gettext', 'GCC', 'libiconv', 'libtool', 'libffi', 'GMP', 'Tcl', 'libpciaccess', 'Python', 'xz', 'gdbm', 'lmod', 'Automake', 'util-macros', 'Autotools', 'EasyBuild', 'libxml2', 'XZ', 'binutils', 'SQLite', 'tar', 'xorg-macros', 'OSU-Micro-Benchmarks', 'Autoconf', 'settarg', 'libreadline', 'OpenBLAS', 'ScaLAPACK', 'CUDA', 'CMake', 'FFTW', 'Clang', 'PyCharm', 'hwloc', 'Bison', 'm4', 'pkgconf', 'zlib', 'M4', 'libsigsegv', 'readline', 'help2man', 'diffutils']

```

so ``list(self.spider_content.keys()`` is not sorted.

buildtest/modules/spider.py
shahzebsiddiqui: Let's fix it for next time. It will show up during regtest and coverage reports

buildtest/modules/spider.py
vsoch: okay I'll add a note about this in the issue for it.

buildtest/modules/util.py
vsoch: Definitely!

buildtest/modules/util.py
vsoch: This suggests that you are still looping through a string (the module name) and not the list - are you sure about your implementation?

buildtest/modules/util.py
vsoch: For this kind of debugging I usually put the logger in charge of this - you import the logger and then can just do logger.debug / logger.info and it will show (or not) depending on the user selected level.

examples/saving_collection.py
vsoch: Did you install it?

vsoch: Yeah I don't see modules - I don't think it's installed. Is there an __init__.py in the folder?

examples/spider_example.py
vsoch: But you don't want to write it into an __init__.py in the modules folder?

shahzebsiddiqui:
if all looks good we could merge this PR @vsoch and remaining changes can be done in subsequent PR. I think the regtest in #185 will give more confidence to this PR

vsoch:
@shahzebsiddiqui I haven't tested anything, and largely I won't be able to until I can make some time and then have all of these merged, so I'm +1 to merge into devel. If there are bugs we'll find them. :)

shahzebsiddiqui:
No worries @vsoch in that case i can merge this and rely on coverage report for test results.

PR #185, opened 2020-02-25
regression test for  module api

shahzebsiddiqui:
This PR is related to #181.  First #181 needs to be merged before this makes sense.


tests/test_modules.py
```diff
@@ -10,7 +10,8 @@
 from buildtest.tools.modulesystem.module_difference import diff_trees
 from buildtest.tools.modulesystem.tree import module_tree_add, module_tree_rm
 from buildtest.tools.log import BuildTestError
-from buildtest.module import Module, ModuleCollection,get_all_collections
+from buildtest.module import Module, ModuleCollection, get_all_collections
+
```
vsoch, line 14: naming question - is there some reason we would want to name the function `get_all_collections` as opposed to just `get_collections`? My expectation would be that `get_collections` would return all by default, but perhaps there is some variable I could provide to filter, etc. If we name `get_all_collections` and then wanted to add such a variable, it would be weird.

tests/test_modules.py
```diff
@@ -25,16 +26,21 @@ def test_module_deps():
     find_module_deps("GCCcore/8.1.0")
     module_tree_rm(["/mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Core"])
 
+
 @pytest.mark.skip("not working")
 def test_diff_trees():
-    diff_trees("/mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Core,/usr/share/lmod/lmod/modulefiles/Core/")
+    diff_trees(
+        "/mxg-hpc/users/ssi29/easybuild-HMNS/modules/all/Core,/usr/share/lmod/lmod/modulefiles/Core/"
+    )
```
vsoch, line 34: Why is there a hard coded path here?

shahzebsiddiqui:
@vsoch  i will open a new PR, had to rebase my branch

tests/test_modules.py
```diff
@@ -87,18 +92,28 @@ def test_collection(self):
         # show "settarg" collection
         cmd.describe("settarg")
 
-class TestModuleCollection:
+    @pytest.mark.xfail(reason="Collection Name must be string when saving", raises=TypeError)
+    def test_type_error_save_collection(self):
+        cmd = Module(["settarg"])
+        cmd.save(1)
```
vsoch, line 99: Just to clarify the usage here - when we do "save" or "describe" the idea is that we are saving a collection?

And a second question - when I create a Module instance, since I can provide a list, is the idea that I'm actually creating multiple modules? Or it's sets of software that would be defined under one Module on a system?

vsoch:
Did you mean to just close this?

vsoch:
Ah were the commits merged?

shahzebsiddiqui:
Yea i had to branch of #181 to do the regtest. Now that its merged to devel i had conflict not to worry I got my issue resolved

Issue #182, opened 2020-02-24
BuildTestCommand should not use shell=True


vsoch:
Currently, `buildtest.tools.system.BuildTestCommand` issues a command to the host with shell=True

```
        self.ret = subprocess.Popen(
            cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
        )
```
From what I've read, this [should be avoided](https://stackoverflow.com/questions/3172470/actual-meaning-of-shell-true-in-subprocess) - we can likely compensate for expanding envars / other features of having the shell through other measures.

shahzebsiddiqui:
Yea i know ``shell=True`` is a security vulnerability and yes if there is an alternative to running shell commands i would be for it. The code would need to be refactored as i recall ``BuildTestCommand`` is used in several places.

vsoch:
Note: we will also need to update the command in spider.py to use this. Comment here https://github.com/HPC-buildtest/buildtest-framework/pull/181#pullrequestreview-364382563

vsoch:
I'd like to work on this after the class refactor for the types, which is a larger chunk of work.

shahzebsiddiqui:
Sounds good with me.

vsoch:
This is fixed in #219

Issue #168, opened 2020-02-20
proposal buildtest build --config option usage


shahzebsiddiqui:
Current Behavior
-------------------

Usage: ``buildtest build -c /path/to/config.yml`` is path is relative to *toolkit/suite*

The ``--config`` or short option ``-c`` is a choice field auto-populated by buildtest. 

The current behavior few limitation including the following

- Test must be in **toolkit/suite**
- Can only build one test at a time
- There needs to be one sub-directory followed by test so **toolkit/suite/compilers/hello.c.yml**. If yaml file is found more than one directory deep, it causes issue with test generation

Proposal
-----------

The overall objective of this issue to address the current issues by achieving the following

- Capable of building more than one test
- Not be bound to a directory path like **toolkit/suite**
- Change ``--config`` to a string instead of a choice field in argparse 
- Provide means for processing a single configuration or a directory path which may contain multiple configuration files.

When user downloads the test such as the tutorial example by running ``buildtest get https://github.com/HPC-buildtest/tutorials.git`` the test are stored in 

Config usage
----------------------

- Build by test name: ``buildtest build -c hello.yml``

- Absolute Path: ``buildtest build -c ~/.buildtest/site/github.com/HPC-buildtest/tutorials/compilers/hello.f.yml``

- Relative Path to site directory: ``buildtest build -c github.com/HPC-buildtest/tutorials/compilers/hello.f.yml``

- Build by directory path (abspath): ``buildtest build -c ~/.buildtest/site/github.com/HPC-buildtest/tutorials/compilers`` 

- Build by directory path (relative path): ``buildtest build -c github.com/HPC-buildtest/tutorials/compilers``

For scope of PR the above features should be addressed to be significant change from current implementation. It is sufficient to resolve the test configuration correctly and simply print out the full path to YAML file and not actually parse, build or run any test. This can be fixed in subsequent PRs when working on the build component.

Additional options to CLI usage will be addressed in another issue.

vsoch:
I like the ability for buildtest to discover recipes in a directory, but it should also be possible to do

```bash
buildtest -c /path/to/recipe.yml
```
and just have it work. As a comparison, image if Singularity or docker required the user to register a path to Dockerfiles or Singularity recipes before running `docker build` or `singularity build`. It would be hugely not intuitive and frustrating. What seems intuitive is to accept a filename OR a directory, and then allow filtering based on a name (e.g., internally using something fnmatch). So these would be valid:

```bash
[...]

PR #191, opened 2020-02-29
Add/single config argument

vsoch:
This pull request will simplify buildtest so that it is not trying to act as the the total manager / controller of recipes. Akin to how you can build a container with Docker and docker doesn't manage where you have your Dockerfiles, opening, or otherwise editing them, buildtest should not be hyper controlling of the test files, as I learned in trying it out, it actually hinders usage more than helps.

However, we can still take advantage of having a `TESTCONFIG_ROOT` where tests can be discovered, and referenced via relative path. This is the goal of this pull request - to remove the "choices" field that is currently required to specify a config, and allow the user to provide it directly as a string. This means that you can specify the relative or absolute path to a test config, OR you can provide a path relative to the TESTCONFIG_ROOT. Both are shown below:

```bash
$ buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/hello.sh.yml
$ buildtest build -c hello.sh.yml
```
We check the relative path first, and if not found, we check in the test config root. If both fail, we exit cleanly. Along with removing this high level controller of listing these configs, As discussed in #168, this only covers the `--config` and `-c` option. I did this intentionally to start more simply - the specification of a directory, or to include or exclude patterns is important, but is not essential for core functionality and thus can be added later.

Also, I added the ability to "get" a particular branch, since I needed this to get my hello world branch:

```bash
$ buildtest get -b add/hello-world-test https://github.com/HPC-buildtest/tutorials.git
```

And I've started to write the getting started guide to go along with this hello world example. The stopping point (and good scoping for this PR) is that in order to get a simpler test working, we need a simpler model than a BuildTestBuilder that requires a compiler. This will be what I work on next, working on the yaml structures and corresponding classes that use them.


buildtest/tools/menu.py
```diff
@@ -367,40 +363,3 @@ def show_menu(self):
         parser_schema = subparsers_show.add_parser("schema", help="Display YAML schema")
         parser_schema.set_defaults(func=show_schema_layout)
         parser_show.set_defaults(func=func_show_subcmd)
-
-    def testconfigs_menu(self):
```
shahzebsiddiqui, line 371: Makes sense to remove ``buildtest testconfigs`` for now as it relates to how configuration name is set.

buildtest/tools/buildsystem/singlesource.py
```diff
@@ -381,21 +383,26 @@ def __init__(self, file, lmod_collection, buildtest_collection):
...
         self.srcfile = self.test_yaml["program"]["source"]
-        self.execname = "%s.%s.exe" % (
-            os.path.basename(file),
-            hex(random.getrandbits(32)),
-        )
+
+        # Generate a common hex to identify and link exec and script file
+        randhex = hex(random.getrandbits(32))
+
+        self.execname = "%s.%s.exec" % (os.path.basename(config_file), randhex,)
```
shahzebsiddiqui: the comma after ``randhex,``  is not needed. 

I think you meant the following.

```
self.execname = "%s.%s.exec" % (os.path.basename(config_file), randhex)
```

buildtest/tools/build.py
```diff
@@ -141,28 +155,29 @@ def run_tests(build_id):
...
     failed_test = 0
 
-    for test in tests:
-        ret = subprocess.Popen(
-            test, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-        )
-        output = ret.communicate()[0].decode("utf-8")
+    with open(run_output_file, "w") as fd:
+        for test in tests:
+            ret = subprocess.Popen(
```
shahzebsiddiqui, line 213: ``subprocess.Popen`` works for now, eventually once we fix #182 we should refactor all code. Looks good with me for now

docs/getting_started.rst
```diff
@@ -0,0 +1,120 @@
...
+    1 directory, 2 files
+
+
+The yaml (extension ``yml`` above) files in the root of the folder are test configurations.
+This means that we write here the commands, environment, and other variables that are needed
+for the test. Notice that the naming is in the format of ``[name][language].yml``. We know off
+the bat that this is intended to test a bash script. Let's take a look at the file::
+
+
+    testtype: singlesource
```
shahzebsiddiqui, line 83: Overall this documentation on ``Getting Started`` is a great first step. The users will refer to this page which was missing. Overall, the content looks good the only thing is the content of YAML script may need to be changed for future which is not big deal we just need to update docs as we progress through the build feature.

shahzebsiddiqui (commented):
Overall this PR looks good to me, just a question. Were you able to get the example build working. If so what was the content of the test script. If you have it, please share the output. Otherwise i can try to test it locally.

buildtest/tools/build.py
```diff
@@ -72,9 +72,23 @@ def func_build_subcmd(args):
...
-        file = os.path.join(TESTCONFIG_ROOT, args.config)
+        # First try, the user can provide a full path
+        config_file = args.config
 
-        singlesource_test = SingleSource(file, args.collection, args.module_collection)
+        # Second try, the path can be relative to the TESTCONFIG_ROOT
+        if not os.path.exists(config_file):
+            config_file = os.path.join(TESTCONFIG_ROOT, config)
+
+        if not os.path.exists(config_file):
```
shahzebsiddiqui: This looks good. I think we can take it a step further in the checks
- Check 1: Check if file exists as specified by ``args.config`` maybe use ``os.path.isfile()`` as well.
- Check 2: Next check if file exist via ``os.path.join(TESTCONFIG_ROOT,config)`` as you did.
- Check 3: If not, check if ``args.config`` is a directory if it passes use ``walk_tree(config_file,".yml")`` to get all YAML files
- Check 4: if check 3 doesn't work with relative path check if ``os.path.join(TESTCONFIG_ROOT,config)`` is a directory and use ``walk_tree`` again. 

Finally depending on the number of config_files we get we do a for loop and call ``SingleSource`` for time being. 


This would allow for any of the following

```
# find the first hello.sh.yml relative to all TESTROOT_CONFIG
buildtest build -c hello.sh.yml

# relative path for a single configuration file
buildtest build -c buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/hello.sh.yml

# all configuration under directory hello-world. Search for first hello-world directory 
buildtest build -c hello-world

# all configuration under directory specified as relative path
buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/
```

This would merge the directory and single configuration behavior in one option. That would be easier for end-users since they have one option to work with instead of adding a directory option in future. This adds more logic on the framework.

My suggestion is to first check if ``args.config``  is a file or directory. Based on the detection, retrieve one or more test configuration and put them in a list before invoking the build.

buildtest/tools/buildsystem/singlesource.py
vsoch: I think this was added by black, it automatically adds commas after the last item of the list. There should be no change to functionality. I can try removing it and running a commit again to be absolutely sure.

docs/getting_started.rst
vsoch: Yes, this is my intention! If I can work on the different kinds of recipes next, then I'll finish updating this document.

vsoch:
@shahzebsiddiqui to answer your question, I could not get any of the tutorials tests working without loading anything (why I started the hello world example) but I just got one to work loading a compiler and then running, let me update the getting started guide with that for now so it's not totally incomplete. I'll push again to this PR.

vsoch:
@shahzebsiddiqui I've updated the comma and the docs for running a build, but the build wasn't entirely successful:

```
Test Name:/home/users/vsochat/.buildtest/testdir/build_7/hello.f.yml.0x69d83e54.sh
Return Code: 0
---------- START OF TEST OUTPUT ---------------- 
/home/users/vsochat/.buildtest/testdir/build_7/hello.f.yml.0x69d83e54.sh: line 11: hello.f.yml.0x69d83e54.exec: command not found
------------ END OF TEST OUTPUT ---------------- 
```
Either it's the case that:
 - the build didn't generate the exec file 
 - the PWD of the build is not correct
 - the exec file needs to be made executable

Do you want to take a look?

buildtest/tools/buildsystem/singlesource.py
vsoch: This is fixed.

vsoch:
Looks like some issue with easy build being used to install bzip - do we need to have ezbuild as a dependency for testing when it's not related to the library?

shahzebsiddiqui (commented):
> Looks like some issue with easy build being used to install bzip - do we need to have ezbuild as a dependency for testing when it's not related to the library?

I saw that easybuild is not installing bzip2 i have no idea. I needed a few software stack packages to test the module feature but certainly since Lmod and settarg are provided they could be sufficient. I will try to update Travis to see if i can use another easybuild recipe. It was also to test some features like checking easybuild modules ``buildtest module -eb`` which why i decided to build a eb stack.

docs/getting_started.rst
```diff
@@ -250,3 +250,34 @@ have seen this output file instead, along with a failed build message::
...
+or within it. For example, the following command will find either a ``hello.sh.yml``
+that is located in your present working directory, or the first file named ``hello.sh.yml``
+in your testing root at ``$HOME/.buildtest/site``::
+
+	buildtest build -c hello.sh.yml
+
+The following will target a specific file path under your test config root::
+
+
+	buildtest build -c buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/hello.sh.ym
```
shahzebsiddiqui: this looks like a typo, there are two **buildtest** commands in one line. Missing file ext , please change **hello.sh.ym** --> **hello.sh.yml**

buildtest/tools/build.py
vsoch: @shahzebsiddiqui - 

 - the first check is if there is nothing provided, that means we use the entire test config root.
 - the second if else statement is essentially the same as doing isfile first, but it's a more concise way to write out the code. The way I implemented it results in the same result, but we don't need to call the os_walk command twice.

We can't technically check if it's a file or directory _first_ because it could be None, and that would result in an error. 

And yep, we are looping through the files and (for now) assuming single source. In the future this will be derive from the config files themselves.

shahzebsiddiqui (changes requested):
minor fix to the docs but overall i dont mind the changes you made. I like the **discover_configs** method and it looks good for now. We can sort out the testing during regtest.

docs/getting_started.rst
vsoch: Oh I copy pasted from slack, that's my bad for not catching it! Let me fix before the panel starts...

shahzebsiddiqui (approved):
Looks good with me! merging PR now thanks @vsoch

PR #192, opened 2020-03-01
remove ModuleCollection class and merge methods into Module()

shahzebsiddiqui:
add error checking on argument passed to Module class. It's ok
to pass no arguments to Module, this is only required when you want
to work with module collection methods.
Remove tests related to "buildtest testconfigs"
Refactor few regression tests for Module() class.
Add MODULEPATH and module av command to Travis for debugging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,8 @@ script:
   - module use $LMOD_PKG/modulefiles/Core
   - echo $MODULEPATH
   - cd $TRAVIS_BUILD_DIR
+  - echo $MODULEPATH
+  - module av
   - coverage run buildtest_regtest.py
   - coverage report -m
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ script:
   - cd $TRAVIS_BUILD_DIR
   - echo $MODULEPATH
   - module av
-  - coverage run buildtest_regtest.py
+  - coverage run -m pytest -vra tests/
   - coverage report -m
 
 after_success:
--- a/buildtest/module.py
+++ b/buildtest/module.py
@@ -1,5 +1,6 @@
 from buildtest.tools.system import BuildTestCommand
 
+
 def get_all_collections():
     """Get all Lmod user collections that is retrieved by running ``module -t savelist``.
 
@@ -14,6 +15,7 @@ def get_all_collections():
     out = cmd.get_error().split()
 
     return out
+
 
 class Module:
     """Class declaration for Module class"""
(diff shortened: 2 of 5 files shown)


shahzebsiddiqui:
@vsoch some of the tests are failing specific to loading and saving the module. I am not sure if it is related to shell (login  or interactive) not sure how pytest runs test internally. It can't find the module command if you look at the regtest output. 

```
tests/test_modules.py:94: AssertionError
----------------------------- Captured stdout call -----------------------------
Saving modules ['settarg'] to module collection name: settarg
/bin/sh: 1: module: not found
Saving modules ['settarg'] to module collection name: default
/bin/sh: 1: module: not found
/bin/sh: 1: module: not found
/bin/sh: 1: module: not found
```

vsoch:
Taking a look.

vsoch:
Does this work locally?

vsoch:
What sticks out to me as off is that get_module_collection is returning a list with /bin/sh, plus an error code, plus a command:

```
AssertionError: assert 'settarg' in ['/bin/sh:', '1:', 'module:', 'not', 'found']

E        +  where ['/bin/sh:', '1:', 'module:', 'not', 'found'] = get_all_collections()
```
the comment in the code is that "# The output is captured in STDERR stream." that seems off / strange to me.

vsoch:
@shahzebsiddiqui is the high level idea of having modules represented here that you can add them as a listing in a test config, and then the test will know to load the modules before running? But wouldn't this break across sites as it would require the same named modules? I think it's important that modules can be represented in this way for the user, but I don't think they should be required for running a build or test.

shahzebsiddiqui:
> @shahzebsiddiqui is the high level idea of having modules represented here that you can add them as a listing in a test config, and then the test will know to load the modules before running? But wouldn't this break across sites as it would require the same named modules? I think it's important that modules can be represented in this way for the user, but I don't think they should be required for running a build or test.

I think all module load detail need to be specified in the configuration and not on command line. I agree it will break stuff when specifying on command line with new approach to ``-c`` option. This means we need to remove ``--collection`` and ``--module-collection`` option from **buildtest build** and user will need to specify this through some YAML key.

vsoch:
I agree.

shahzebsiddiqui:
> What sticks out to me as off is that get_module_collection is returning a list with /bin/sh, plus an error code, plus a command:
> 
> ```
> AssertionError: assert 'settarg' in ['/bin/sh:', '1:', 'module:', 'not', 'found']
> 
> E        +  where ['/bin/sh:', '1:', 'module:', 'not', 'found'] = get_all_collections()
> ```
> 
> the comment in the code is that "# The output is captured in STDERR stream." that seems off / strange to me.

I think it could be related to the way ``BuildTestCommand`` is implemented. Right now both STDOUT and STDERR are passed to `subprocess.PIPE` anyhow I think this is only failing on few tests and I was able to get this run locally.  I would want to look at this later considering the core functionality of `ModuleCollection` is now migrated to `Module` and fixes #192. If all looks good I think we can merge this PR

shahzebsiddiqui:
@vsoch also note i removed regtest ``tests/test_testconfigs.py`` which relates to your PR #191. I just realize you have removed the file from PR so dont worry if codecov report is dropped. Its mainly due to   that. The `module_difference.py` had drop in coverage because I accidently skipped this test locally since it doesnt work locally but only on Travis. I was suppose to let this run. Which is fixed now.

PR #198, opened 2020-03-03
Remove/lmod dependency

vsoch:
This is a first shot at removing the lmod dependency so that I'm able to better develop and test locally for the next round of changes (supporting different recipe types). Note that I added a parameter to the system to be able to capture if lmod is installed, which can be referenced later. For some future update I'd like to better organize modules to be a subtype of environment (or we can call it something else) but this should work for now. I also:

 - add a run_tests.sh bash script to replace the buildtest_regtests.py file. The reason is because pytest is most understood to be run from the command line, and the user should be able to inspect the file and choose some subset of commands. If you look in the file, we skip over the module tests if lmod isn't installed (and can do this for other future environments). The testing in .travis.ci, since we assume lmod is installed, will of course run all of them. We might want to add a test case here to support trying the same without having lmod, if you think that's necessary.
 - remove module commands from client: as they are redundant. Buildtest is not a wrapper for the module command, this doesn't make sense.
 - update test_config_files so tests for modules are in the test_modules function.


vsoch:
@shahzebsiddiqui I don't think that the module configs should exist by default - nor should a collection config (writing an empty file doesn't make a lot of sense). I think we should remove those checks for now, and when a full test is added that reads in a recipe file with a module, this is when we would need to load the Module class and ensure that the file is added, etc.

vsoch:
https://github.com/z4r/python-coveralls/issues/73

buildtest/spider.py
```diff
@@ -20,15 +20,14 @@ def __init__(self, tree=None):
         if tree:
             self.tree = tree
 
-            if not os.getenv("LMOD_DIR"):
-                raise SystemError(
-                    "Cannot find environment variable LMOD_DIR. Please fix your Lmod configuration!"
+            # Lmod can be optionally installed for using modules
+            self.spider_content = []
+            if os.getenv("LMOD_DIR"):
```
shahzebsiddiqui: Should we be explicit to check if path of LMOD_DIR also exist as valid directory. It is likely a user can bypass this and set as environment variable. The spider command is accessible if you load the Lmod module or use the LMOD_DIR to reference the path. By default any site that has Lmod installed as this environment variable setup as part of startup configuration.

buildtest/tools/modulesystem/tree.py
```diff
@@ -5,25 +5,6 @@
 from buildtest.tools.modules import update_spider_file
 
 
-def func_module_tree_subcmd(args):
-    """ Entry point for ``buildtest module tree`` subcommand """
```
shahzebsiddiqui, line 9: How do we expect buildtest to update spider file when they change BUILDTEST_MODULEPATH. The idea was for user to not update that variable from config file. But if they do from config file we need a history to check what was the value before they update the file. I guess we could capture this detail in some json file maybe even The spider file. The spider command needs to run against all module trees. That's the essence of getting all the module features working

buildtest/tools/menu.py
```diff
@@ -160,157 +145,6 @@ def get_menu(self):
 
         parser_get.set_defaults(func=func_get_subcmd)
 
-    def module_menu(self):
```
shahzebsiddiqui, line 163: Ok if we remove all these from argparse. Where will this be exposed. The module part is important but if it's not part of buildtest then will we have a separate program or will module API be taken out of buildtest into a separate git project.
The module load test is a key part of buildtest. Initially I wanted a framework to automate software stack testing

buildtest/spider.py
vsoch: yes, this is a good idea. I'll add this now. I think I had originally not added so I could try to "trick" buildtest that I had it installed just by exporting the envar. :)

buildtest/tools/modulesystem/tree.py
vsoch: So - let's step back and discuss what the purpose of this spider file is. You want to always keep a total representation of all system modules? Why? The module path is logical to keep if we need to look for modules, but keeping the listing cached is a huge responsibility, and in that the module is a kind of environment, and one that has a specific state at the run of any recipe, I don't think it makes sense to try to always keep track of this. To play devil's advocate, let's say we do keep track of it. Then the user is required to run that command. Are they really going to do that? Probably not. It makes much more sense to interact with spider and derive some state when it's actually needed, for doing a build test.

buildtest/tools/menu.py
vsoch: The modules I think are best entirely controlled via the build test configuration files. If I need a particular module, I add it to the listing there. The state is determined at runtime, so it's not going to be dated or rely on some update to a cache. But I'm not required to use modules if I don't want to. The direction I think makes sense is to move toward a module being a kind of environment that can be used, but not so hard coded around the application, and with equal representation to other kinds of environments that the user might want. In the same way we will add different kinds of test config types, there can also be a family of environment/etc. types that can be specified for use to load for any particular build.

vsoch: >  Initially I wanted a framework to automate software stack testing

You would still have this! But to adjust your statement, what you were creating is _a framework to automate only stack testing if the site uses modules_. To be more true to the statement, it should be possible to support any flavor of test environment. Just like spack, or snakemake, in that you have a base class of "a test configuration case" and "an environment that can be used" I think we should aspire for that for buidltest. Then it will be the most useful / powerful.

vsoch:
hah! So codecov is failing, but travis is at least passing. If codecov is just for test coverage, we will definitely be working on that, albeit likely not in this exact PR.

**Update** codecov is magically green now.  No idea how that is working, but I'm good with it :)

shahzebsiddiqui (approved):

PR #200, opened 2020-03-04
remove all module features from buildtest.

shahzebsiddiqui:
Moved ModuleAPI to lmodule project
https://github.com/HPC-buildtest/lmodule
Remove all doc test examples for unused features and update file buildtest/docgen/main.py
Rebuild API docs
Remove all unused imports from files
Remove module section from default configuration (settings.yml)
Remove regtest for module features
Remove buildtest metadata json files such as BUILDTEST_SPIDER_FILE,
BUILDTEST_MODULE_COLLECTION_FILE
For now Lmod check doesn't have to do much since Spider file is not
needed anymore.
Remove top-level examples folder. This will be moved to lmodule project


vsoch:
It looks like you fixed the issue with the args? I'm going to push fixes for black, then I think we can merge.

vsoch:
We also needed to update the git hook - the examples folder doesn't exist anymore so it would hit an error.

vsoch (approved):
Sometimes the hardest (but highest quality work) is knowing when to prune, and I think you've done a great job here! We will still need to interact with modules using some simple shell commands, but I'm glad that development of this wrapper can move to it's own module, which is most appropriate.

shahzebsiddiqui:
> We also needed to update the git hook - the examples folder doesn't exist anymore so it would hit an error.

Good catch. I forgot about the hook file.

PR #202, opened 2020-03-08
update urlchecker to v0.1.4

shahzebsiddiqui:
See https://github.com/SuperKogito/URLs-checker/issues/8

### DIFF
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -18,4 +18,5 @@ jobs:
 
         # Choose whether to include file with no URLs in the prints.
         print_all: False
-        
+
+        white_listed_urls: "https://github.com/marketplace/trafico-pull-request-labeler,https://github.com/HPC-buildtest/buildtest-framework,https://easybuild.readthedocs.io/en/latest/,https://github.com/HPC-buildtest/buildtest-framework/issues,https://buildtest.readthedocs.io/en/latest/index.html"
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -18,3 +18,5 @@ jobs:
 
         # Choose whether to include file with no URLs in the prints.
         print_all: False
+        
+        white_listed_urls: https://github.com/marketplace/trafico-pull-request-labeler, https://github.com/HPC-buildtest/buildtest-framework, https://easybuild.readthedocs.io/en/latest/, https://github.com/HPC-buildtest/buildtest-framework/issues, https://buildtest.readthedocs.io/en/latest/index.html 
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -19,4 +19,3 @@ jobs:
         # Choose whether to include file with no URLs in the prints.
         print_all: False
         
-        white_listed_urls: https://github.com/marketplace/trafico-pull-request-labeler, https://github.com/HPC-buildtest/buildtest-framework, https://easybuild.readthedocs.io/en/latest/, https://github.com/HPC-buildtest/buildtest-framework/issues, https://buildtest.readthedocs.io/en/latest/index.html 
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -8,7 +8,7 @@ jobs:
 
     steps:
     - name: URLs-checker
-      uses: SuperKogito/URLs-checker@0.1.3
+      uses: SuperKogito/URLs-checker@0.1.4
       with:
         # The project base path.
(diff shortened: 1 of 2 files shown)


vsoch:
I think minimum criteria for evaluating the url checker would be first for the action to run successfully. I can take a look then.

shahzebsiddiqui:
@vsoch i requested the developer to add a feature to the URLchecker see https://github.com/SuperKogito/URLs-checker/issues/8 the output is sanitized to only failures, before it was reporting all links. Now the URLchecker is failing since it found some links that failed, before it was passing. I guess it is a step in right direction.

vsoch:
Cool! Would it make sense to fix the urls that are broken for this PR then? I already read the issue - great response time (under 2 weeks!) to get your request implemented!

shahzebsiddiqui:
> Cool! Would it make sense to fix the urls that are broken for this PR then? I already read the issue - great response time (under 2 weeks!) to get your request implemented!

Sure i will fix these links soon.

shahzebsiddiqui:
@vsoch it looks like the `white_listed_urls` does not work as i expect. All the links are valid when checking them manually. The one in `docs/conf.py` was one that needed to be fixed since it was redirected to another link. Anyhow, I will remove the white_listed_urls since I dont want to keep a track of all these links which are valid. I have reported this to developer to help look into this. Otherwise we may just have to look at this once in a while to see if anything is broken.

vsoch:
Can you show me the version of the action that had `white_listed_urls` and the link to the action in GitHub actions? I'd like to take a look.

vsoch:
I think your problem is that the comma separated list has spaces?

```
        white_listed_urls: https://github.com/marketplace/trafico-pull-request-labeler, https://github.com/HPC-buildtest/buildtest-framework, https://easybuild.readthedocs.io/en/latest/, https://github.com/HPC-buildtest/buildtest-framework/issues, https://buildtest.readthedocs.io/en/latest/index.html 
```
I wouldn't expect a comma separated list to allow spaces. It's set as an environment variable, so when you have spaces without quotes you don't get all ofit.

vsoch:
See how it's grabbed [here](https://github.com/SuperKogito/URLs-checker/blob/master/check.py#L86)

vsoch (changes requested):
Let's try the white listed urls either removing the spaces between commas, OR putting the entire thing in quotes. I suspect it's not working because of the usage here.

shahzebsiddiqui:
> Let's try the white listed urls either removing the spaces between commas, OR putting the entire thing in quotes. I suspect it's not working because of the usage here.

The underlying issue is that all the links are valid but URLchecker is reporting them they are not. I dont want to manage all of these links as white_list_patterns. It requires us to update the workflow as we change docs and some links may not be relevant. 

You can look https://github.com/HPC-buildtest/buildtest-framework/runs/493725470?check_suite_focus=true for more details.

vsoch:
The action link is from 2 hours ago, which was before removing the empty spaces. Can we please test removing the empty spaces?

vsoch:
If you want to remove all white list urls and not white list anything, then that's a different thing. But we probably should have a passing test, still.

shahzebsiddiqui:
@vsoch I think the space and probably quoting everything makes it work. Looking at the links, many if not all are valid links and its reported as false positive. I have reported this to developer for now I guess i will add all of them to make this pass.

vsoch:
Interesting - so let's look at the test cases so we can give feedback to the author:

 -  https://buildtest.readthedocs.io/en/latest I see is failing, but the version you provide has a trailing slash, does the integration need to sanitize to make use / non use consistent?

vsoch:
If you are giving the developer a note, ask him/her about pattern expansion, e.g., so we could provide a prefix with some expander to mean some entire subset of urls.

shahzebsiddiqui:
The issue is https://github.com/urlstechie/URLs-checker/issues/8 the developer hasnt got chance to reply back. I will try to look at some of test this out with few more commits. Stay tuned.

shahzebsiddiqui:
ok let's hold off on this PR, the developer just responded and will look into this. I'd hope this can be fixed in the next version rather than me patching with white list of many urls. Once there is a fix I will update the version. So will keep this PR open.

shahzebsiddiqui:
Closing this PR because of #211 will address URL check with retry

vsoch:
Sounds good - I'm going to wait to hear back about the branch and repo too, because we would absolutely need this added for it to (truly) function correctly I'm hoping that I can have a discussion with the lead maintainer, do a PR to add support for different branches (or just use the repo checkout by default, which makes more sense) and then there can be a versioned release with _both_ fixes for us to check here.

PR #204, opened 2020-03-09
updating docs including section on contribution, CHANGELOG, and some …

shahzebsiddiqui:
…output examples from doc examples


docs/contributing.rst
```diff
@@ -15,123 +13,25 @@ There are many ways you can help contribute to buildtest that may include
...
 
-GitHub Bots
------------
+Maintainers are listed in alphabetical order by Last Name:
 
-Buildtest has a few bots to do various operations that are described below.
+- Shahzeb Siddiqui (**@shahzebsiddiqui**)
 
-- **Stale**  - buildtest is using `Stale <https://github.com/marketplace/stale>`_ to close outdated issues. This is configured in ``.github/stale.yml``. If there is no activity on a issue after certain time period, **probot-stale** will mark the issue and project maintainers can close it manually. For more details on Stale refer to the `documentation <https://probot.github.io/>`_
+- Vanessa Sochat (**@vsoch**)
```
vsoch, line 25: :D

docs/contributing/github_integration.rst
```diff
@@ -0,0 +1,115 @@
+GitHub Integration
+====================
+
+buildtest has several github integration, including automated checks during PR that maintainers will check to
```
vsoch: nit: github integration**s**

vsoch: "including automated checks during PR that maintainers will check during the PR remove" (delete the "to" after check)

docs/contributing/github_integration.rst
```diff
@@ -0,0 +1,115 @@
...
+during the PR review. You should check results from the `buildtest actions <https://github.com/HPC-buildtest/buildtest-framework/actions>`_
+to see if any fixes need to me made.
```
vsoch: maybe we should add that it's linked from the PR too, something like:

> You should check results from the `buildtest actions <https://github.com/HPC-buildtest/buildtest-framework/actions>`_ that are also typically linked as part of the pull request testing suite.

docs/contributing/github_integration.rst
```diff
@@ -0,0 +1,115 @@
...
+
+It's good practice to check Travis `builds <https://travis-ci.com/HPC-buildtest/buildtest-framework>`_ since we use Travis
+to run regression test and Codecov and Coveralls depend on Travis to pass all checks.
+
+You will want to make sure code is formatted via black as we have automated checks for python formatting. If you have not
+setup the black hook check out :ref:`black_hook`
```
vsoch, line 12: Maybe for now we can add something like:

> If you run the black linter and find that the check still fails, check the step of the action that installs black and ensure that you are using the same version locally.

Another thing we can do is to pin a particular black version and then tell the user up front to use it, also spit out a warning in the commit hook if it's a different version.

docs/contributing/github_integration.rst
```diff
@@ -0,0 +1,115 @@
...
+- **GuardRails** - GuardRails provides continuous security feedback for modern development teams
+
+- **Snyk** - Snyk tracks vulnerabilities in over 800,000 open source packages, and helps protect over 25,000 applications.
+
+Links to the following apps
+
+- **Travis**: https://travis-ci.com/HPC-buildtest/buildtest-framework
+
+- **CodeCov**: https://codecov.io/gh/HPC-buildtest/buildtest-framework
+
```
vsoch, line 23: Would it make sense to put the links to the apps alongside the descriptions so it doesn't look like they are listed twice / redundant?

docs/contributing/github_integration.rst
```diff
@@ -0,0 +1,115 @@
...
+    -    x=0+1*3
+    +    x = 0 + 1 * 3
+
+The changes will be shown with lines removed or added via ``-`` and ``+``. For more details refer to black
+`documentation <https://github.com/psf/black>`_.
+
+GitHub Bots
+-----------
+
+Buildtest has a few bots to do various operations that are described below.
```
vsoch: One thing to bring up for general discussion - I've seen buildtest with upper and lower case B, is the convention to always use lowercase unless we are starting a sentence?

docs/contributing/release_process.rst
```diff
@@ -1,6 +1,15 @@
-Release Process
+Maintainer Guide
 ================
 
+Incoming Pull Request
+------------------------
+
+Any incoming Pull Request should be assigned to one or more maintainers for review. Upon approval, the PR should
+be **Squash and Merge**. If it's important to preserve a few commits during PR then **Rebase and merge** is acceptable.
+
```
vsoch: Maybe we should add a note about cleaning up the final squashed message if appropriate?

docs/contributing/setup.rst
```diff
@@ -90,8 +85,60 @@ Once you are ready to push to your fork repo do the following::
...
+Instead of merge from ``devel`` you can rebase your commits interactively when syncing with ``devel``. This can be done by running::
+
+    git rebase -i devel
+
+Once you have synced your branch push your changes and check if file conflicts are resolved in your Pull Request::
+
+    git push origin <feature-branch>
+
+
+General Tips
```
vsoch: How do these tips relate to the CONTRIBUTING.md? Or a better question - what goes here vs. there?

docs/contributing/setup.rst
```diff
@@ -90,8 +85,60 @@ Once you are ready to push to your fork repo do the following::
...
+
+2. If you have an issue, ask your question in slack before reporting issue. If your issue is not resolved check any open issues for resolution before creating a new issue
+
+3. For new features or significant code refactor please notify maintainers and open an issue before working on task to keep everyone informed.
+
+4. If you open an issue, please respond back during discussion, if there is no activity the issue will be closed.
 
-Someone from the **buildtest team** will review the PR and get back to you with the feedback. If the reviewer requests
-some changes, then the user is requested to make changes and update the branch used for sending PR
+5. Please don't open a duplicate issue if their is an existing issue addressing similar problem, instead you can participate in discussion in the issue or contact appropriate individuals directly in slack
```
vsoch: if _there_ is an existing issue

vsoch (changes requested):
Overall looks good! Just a few nits and questions.

docs/contributing/github_integration.rst
shahzebsiddiqui: Good point, i guess i can change it to lower case to be consistent even if grammatically it may make no sense. The name is `buildtest` so it may be confusing with `Buildtest`

vsoch: Great! Yeah as long as we are consistent we shouldn't see derivations popping up like `Buildtest` or even `BuildTest`.

docs/contributing/github_integration.rst
shahzebsiddiqui: I agree, i will fix this

docs/contributing/setup.rst
shahzebsiddiqui: good point i dont know the best place, let me try putting this in another location.

docs/contributing/github_integration.rst
shahzebsiddiqui: If we want to pin a version, it may make sense to add this to `docs/requirements.txt` so that all deps are added in one place and user dont have to do this manually. Looking at https://buildtest.readthedocs.io/en/devel/contributing.html#github-hooks i can see the `pip install black==19.3b0` be something user don't have to worry about if its part of `requirements.txt`. Anyhow I will add some section about your comment in the docs.

shahzebsiddiqui:
@vsoch  just synced this PR with latest changes from `devel`.

docs/contributing/github_integration.rst
vsoch: Yeah this is a good idea. Then we will need to install that version explicitly for the test as well!

docs/contributing/github_integration.rst
```diff
@@ -1,47 +1,35 @@
...
-- **Travis**: https://travis-ci.com/HPC-buildtest/buildtest-framework
-
-- **CodeCov**: https://codecov.io/gh/HPC-buildtest/buildtest-framework
+- `Travis CI <https://travis-ci.com/HPC-buildtest/buildtest-framework>`_ - Test and deploy with confidence. Trusted by over 800,000 users, Travis CI is the leading hosted continuous integration system.
 
-- **Coveralls**: https://coveralls.io/github/HPC-buildtest/buildtest-framework
+- `CodeCov <https://codecov.io/gh/HPC-buildtest/buildtest-framework>`_ - Codecov provides highly integrated tools to group, merge, archive and compare coverage reports
 
-- **CodeFactor**: https://www.codefactor.io/repository/github/hpc-buildtest/buildtest-framework
+- `Coveralls <https://coveralls.io/github/HPC-buildtest/buildtest-framework>`_ - Coveralls is a web service to help you track your code coverage over time, and ensure that all your new code is fully covered.
```
vsoch, line 27: This is much cleaner! One question that comes to mind is that (for some future update) if we can store the repository address / name as a variable, so we don't need to do manual changes for links like this. I'm not sure the extent to which this is possible with sphinx, but maybe? Either way, that's out of scope for here, this looks good.

docs/contributing/maintainer_guide.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+Incoming Pull Request
+------------------------
+
+These are just a few points to consider when dealing with incoming pull requests
+
+1. Any incoming Pull Request should be assigned to one or more maintainers for review.
+
+2. Upon approval, the PR should be **Squash and Merge**. If it's important to preserve a few commits during PR then **Rebase and Merge** is acceptable.
```
vsoch, line 14: I think there is a setting for this under the repo settings, I'll check after this final review.

vsoch (approved):

vsoch:
To follow up on my comment, I think we can enable/disable different merge actions, but not set a default (unless we disable the first and the squash and merge just happens to be first in the list) 

image

Since this is a limited subset of maintainers anyway, likely we can just have awareness about this and update any new contributors that get write access.

shahzebsiddiqui:
> To follow up on my comment, I think we can enable/disable different merge actions, but not set a default (unless we disable the first and the squash and merge just happens to be first in the list)
> 
> image
> 
> Since this is a limited subset of maintainers anyway, likely we can just have awareness about this and update any new contributors that get write access.

I think you bring up a good point I prefer having a linear history and would disable **Merge Commit** button. This could be prime example to put in maintainers guide docs. Even as silly as this may sound, having some standard rules around repository setting is important as it helps keep everyone in sync and avoid having accidental changes to repository setting by any admin. If we decide to change some admin settings it should be full concensus or atleast majority rule and then we update the settings and docs.

PR #205, opened 2020-03-09
Moving install_deps into testing directory, as it is only used there

vsoch:
The script install_deps.sh is only used within the context of testing, so logically this script should be kept with tests. Also, if I'm a new user and I clone the repository, I'm likely to want to run tests early on to test things, and I'm more likely to see the install_deps.sh script that might be useful to me if it's in the tests folder than hidden somewhere in the package. Since the "scripts" folder under buildtest is now empty, this should clean up the top level module organization too.


shahzebsiddiqui (approved):
This makes sense to me. Thanks for fixing this.

PR #207, opened 2020-03-09
Removing argcomplete and termcolor, not essential for buildtest

### LINKED ISSUE #206: Dependency Discussion: termcolor
Where is termcolor used? I don't see it outside of the setup files, is it a dependency we can remove?

```
setup.py:        "termcolor==1.1.0",
buildtest_framework.egg-info/requires.txt:termcolor==1.1.0
docs/requirements.txt:termcolor==1.1.0
```

vsoch:
This PR will remove argcomplete and termcolor, which are either no longer necessary or add extra (unneeded) dependencies. It's a move in the right direction to keep the library simple - the minimum number of dependencies that we need is the way to go! :) 

 - This will close #206 

I also updated the current changelog to be for March (I can't believe it's March already...)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,10 @@
 CHANGELOG
-=================
+=========
 
-v0.8.0 (Feb xxx, 2020)
+v0.8.0 (Mar xxx, 2020)
 -----------------------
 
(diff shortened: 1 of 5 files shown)


shahzebsiddiqui (approved):
Looks good to me. Thanks

PR #211, opened 2020-03-10
testing urlchecker that has retry

vsoch:
This pull request will test an update to URL checker, where we add a retry (3 times for this case) for failed URL checks. See the changes here https://github.com/urlstechie/URLs-checker/pull/17.

### DIFF
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: URLs-checker
-      uses: urlstechie/URLs-checker@add/checkout-optional-git-path
+      uses: urlstechie/URLs-checker@0.1.6
       with:
         # check this subfolder only
         subfolder: docs
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -11,6 +11,9 @@ jobs:
     - name: URLs-checker
       uses: urlstechie/URLs-checker@add/checkout-optional-git-path
       with:
+        # check this subfolder only
+        subfolder: docs
+
         # A comma-separated list of file types to cover in the URL checks
         file_types: .md,.py,.rst
 
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -8,7 +8,7 @@ jobs:
 
     steps:
     - name: URLs-checker
-      uses: SuperKogito/URLs-checker@0.1.3
+      uses: urlstechie/URLs-checker@fix/bug-with-check-response-status
       with:
         # The project base path.
         git_path: https://github.com/HPC-buildtest/buildtest-framework
@@ -17,4 +17,7 @@ jobs:
         file_types: .md,.py,.rst
 
         # Choose whether to include file with no URLs in the prints.
-        print_all: False
+        print_all: false
+
+        # How many times to retry a failed request (defaults to 1)
+        retry_count: 3
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -22,3 +22,6 @@ jobs:
 
         # How many times to retry a failed request (defaults to 1)
         retry_count: 3
+
+        # Cannot check private GitHub settings
+        white_listed_patterns: https://github.com/HPC-buildtest/buildtest-framework/settings
--- a/.github/workflows/urlchecker.yml
+++ b/.github/workflows/urlchecker.yml
@@ -7,12 +7,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: actions/checkout@v2
     - name: URLs-checker
-      uses: urlstechie/URLs-checker@fix/bug-with-check-response-status
+      uses: urlstechie/URLs-checker@add/checkout-optional-git-path
       with:
-        # The project base path.
-        git_path: https://github.com/HPC-buildtest/buildtest-framework
-
         # A comma-separated list of file types to cover in the URL checks
         file_types: .md,.py,.rst
 
--- a/README.rst
+++ b/README.rst
@@ -24,9 +24,9 @@
 buildtest
 ---------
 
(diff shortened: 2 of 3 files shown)


vsoch:
okay we have a bug! I'm going to fix it on the upstream, update the commit here, and continue.

vsoch:
Great! I just had a good run for the URL checker - I cleaned up the printing quite a bit and adding printing of params so it's much more clear, going to open a PR over there and hopefully get a release candidate we can use to update the version we use.

vsoch:
https://github.com/urlstechie/URLs-checker/pull/19

shahzebsiddiqui:
the output looks much better, if i understand the output in green are success after retry. i take it once there is a release may `0.1.4` we can bump the version instead of using ` urlstechie/URLs-checker@fix/bug-with-check-response-status` . 

One thing i found odd is some files are getting check when they dont exist. For instance there is no such page **buildtest-framework/docs/build_subcommand/module_collections.rst**   and its getting picked up from where?

The page was deleted and i can see it is not found in your feature branch https://github.com/researchapps/buildtest-framework/tree/test/URLchecker-update/docs 


```
 buildtest-framework/docs/build_subcommand/module_collections.rst 
 ----------------------------------------------------------------
https://lmod
https://lmod
```

vsoch:
This would likely be a separate bug with the tester (addressed in another issue/PR I think). I think it's likely that we need to add checkout of a branch? I don't see anywhere that a branch is checked out - https://github.com/urlstechie/URLs-checker/blob/master/check.py#L15.

I'll open an issue and see if we can add this, it obviously is super important to be testing the right branch!

vsoch:
Also - the git path is used for the clone, for a PR from a different user this would not work. If we trust the url check to be run for someone else's repo, then it's safe to change this to default to whatever the github repository of the PR is set to be.

shahzebsiddiqui:
> This would likely be a separate bug with the tester (addressed in another issue/PR I think). I think it's likely that we need to add checkout of a branch? I don't see anywhere that a branch is checked out - https://github.com/urlstechie/URLs-checker/blob/master/check.py#L15.
> 
> I'll open an issue and see if we can add this, it obviously is super important to be testing the right branch!

I agree, if its not testing the PR branch then result is useless. My guess is its testing the `master` branch i can see it on line 7 https://github.com/HPC-buildtest/buildtest-framework/edit/master/docs/build_subcommand/module_collections.rst.

shahzebsiddiqui:
> Also - the git path is used for the clone, for a PR from a different user this would not work. If we trust the url check to be run for someone else's repo, then it's safe to change this to default to whatever the github repository of the PR is set to be.

This means each user fork has to change `git_path` https://github.com/HPC-buildtest/buildtest-framework/pull/211/files#diff-2b1385f14466a13c63506ce44d851102L14 not really intuitive. Or we test only devel branch to check URls periodically before merging them to `master`.  It may make sense to go through all the URL check before doing a release as most links would be found in documentation. It would be just one added thing doing release instead of doing once per PR.

vsoch:
I'm not sure why he has a git clone at all - the logical thing would be to checkout the repo and then have everything there. I think the variable should be removed (or optional) but I'll check with him first.

vsoch:
https://github.com/urlstechie/URLs-checker/issues/20

vsoch:
@shahzebsiddiqui I think we have a working URLchecker now! https://github.com/HPC-buildtest/buildtest-framework/pull/211/checks?check_run_id=501529549 I haven't PRed the branch to upstream (so there isn't a release that is ready for use) but I wanted for you to take a look here. Notice that:

 - we don't provide a git_path anymore, the default branch / repo that is already checked out is used
 - we are only checking the docs folder, afforded by adding the "subfolder" variable
 - I added a user agent string so that we don't get 403 errors (the hpc wire failure is now resolved)
 - I whitelisted the settings GitHub path - this won't be viewable publicly and thus would 404.

I'm going to open a PR now, I'll reference the build for review.

.github/workflows/urlchecker.yml
```diff
@@ -7,14 +7,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - uses: actions/checkout@v2
     - name: URLs-checker
-      uses: SuperKogito/URLs-checker@0.1.3
+      uses: urlstechie/URLs-checker@add/checkout-optional-git-path
```
shahzebsiddiqui: Overall it looks good, if the PR https://github.com/urlstechie/URLs-checker/pull/24 is merged and version is bumped in the action we can bump it here.

.github/workflows/urlchecker.yml
```diff
@@ -7,14 +7,21 @@ jobs:
...
       with:
-        # The project base path.
-        git_path: https://github.com/HPC-buildtest/buildtest-framework
+        # check this subfolder only
+        subfolder: docs
```
shahzebsiddiqui, line 15: This makes sense to me and a nice feature to have.

shahzebsiddiqui (commented):
Overall, this looks great. I am glad we got the retry feature and ability to checkout branch so we are testing the PR and not `master` branch. This looks good to me and we can merge this upon new version provided by developer.

.github/workflows/urlchecker.yml
vsoch: In that case we would need to use a commit sha, but likely we can wait a day or two and the author will have a release (he seems speedy for these things). I'd prefer to wait for the release, but that's just me.

vsoch:
hey @shahzebsiddiqui this is ready for merge! I've updated to 0.1.6, and has a much nicer print out to the terminal, and also we can see retry in action!

image

Note that the url above had a timeout, was retried, and then returned green.

shahzebsiddiqui:
@vsoch great work i like it now, i see the retry worked on the Slack Heroku link. This looks really good now and I can see this to be useful when writing docs to check all URLs. One main concern is when URLs change over time and this can cause broken links in the docs, I am going to merge this now.

vsoch:
Great! I'm working on some more bug fixes now, for example, the newlines error and allow for patterns of files (instead of urls) to fix. https://github.com/urlstechie/URLs-checker/pull/30

PR #213, opened 2020-03-12
moving BuildTestError and BuildTestCommand methods to utils 

shahzebsiddiqui:
This PR will do the following
 - Move *BuildTestCommand* and *BuildTestError* to ``buildtest/utils/command.py``
 - Fix import statements as necessary 
 - Add regtest for *BuildTestCommand* now. We know this method will be changed but atleast we should test it


buildtest/utils/file.py
```diff
@@ -12,7 +12,7 @@
 import logging
 
 from buildtest.defaults import logID
-from buildtest.log import BuildTestError
+from buildtest.utils.command import BuildTestError
```
vsoch: hmm, should this be under `buildtest.utils.exceptions`?  Or actually, where I usually expect custom exceptions for a library would be is under `buildtest.exceptions` directly. Do you consider this a command?

shahzebsiddiqui:
@vsoch  I know we discussed this, hope this change makes sense.

buildtest/utils/file.py
shahzebsiddiqui: Sure i can move this over to ``exceptions.py`` give me a second to refactor the code

buildtest/utils/command.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+        if args:
+            msg = msg % args
+        self.msg = msg
+
+    def __str__(self):
+        return repr(self.msg)
+
+
+class BuildTestCommand:
```
vsoch, line 4: Love this organization! It's really intuitive to find the `BuildTestCommand` under utils.commands too.

buildtest/utils/file.py
vsoch: Awesome! And I think it's most intuitive to find at the root (right under buildtest) if that works for you too. I can't tell you how many times I've been writing something to catch an error, and interactively I look for some `<module>.exceptions` to import from.

buildtest/buildsystem/singlesource.py
```diff
@@ -10,7 +10,7 @@
 
 from buildtest.config import config_opts
 from buildtest.defaults import logID
-from buildtest.utils.command import BuildTestError
+from buildtest.utils.exceptions import BuildTestError
```
vsoch, line 14: Can we instead do buildtest.exceptions? That is where I'd expect it.

shahzebsiddiqui: Sure, let me fix it.

vsoch (approved):
LGTM, good for merge when all tests pass (green)

PR #217, opened 2020-03-13
remove buildtest show schema feature. 

### LINKED ISSUE #215: Remove buildtest show schema
The purpose of ``buildtest show schema`` was to show YAML schema on CLI with all the key and description. Since this is being developed and documented in https://github.com/HPC-buildtest/schemas we can remove the feature for now.

I think in future we can add the cli feature to show schema that could take the ``json`` schema and all the YAML examples and print them to screen. Another option could be to have all the schema docs on a GitHub page and use the CLI to pull the page locally. It could be like a man page style. 

I can see the feature useful if designed like a man page which would help users that are developing a test to view docs and examples from command line instead of going to a separate web-page. If this is to work, all the docs should be in the schema repo or some GitHub page so that the framework is not re-implementing the schema docs.

shahzebsiddiqui:
Fix #215.
Refactor code, remove regtest and update docs where reference
to schema. Removed the example docs generated by docgen script


vsoch (changes requested):
This most definitely needs to come after the work that I'm doing to refactor the schema - in that a build schema comes from what I'm working on, there is no reasonable way to write this until that is done. I am adding request changes as this cannot be further worked on until the script schema is added.

shahzebsiddiqui:
@vsoch I am going to close this PR. I realize you added two new options in #219 in ``buildtest show schema`` which conflicts with this PR.

Issue #212, opened 2020-03-12
bug with argparse in buildtest build


shahzebsiddiqui:
There is some bug when running ``buildtest build --help`` there is a TypeError issue which i dont get what happened. I haven't had time to look into this as we need to change the way argparse is setup. For now putting this in to fix.

```
(docs) ssi29@ag-mxg-hulk090> buildtest build -h
Traceback (most recent call last):
  File "/u/users/ssi29/.local/share/virtualenvs/docs-dnVBLcEp/bin/buildtest", line 11, in <module>
    load_entry_point('buildtest-framework', 'console_scripts', 'buildtest')()
  File "/mxg-hpc/users/ssi29/buildtest-framework/buildtest/main.py", line 20, in main
    parser.parse_options()
  File "/mxg-hpc/users/ssi29/buildtest-framework/buildtest/menu/__init__.py", line 76, in parse_options
    args = self.parser.parse_args()
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1749, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1781, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1969, in _parse_known_args
    positionals_end_index = consume_positionals(start_index)
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1946, in consume_positionals
    take_action(action, args)
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1855, in take_action
    action(self, namespace, argument_values, option_string)
  File "/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/lib/python3.7/argparse.py", line 1152, in __call__
    subnamespace, arg_strings = parser.parse_known_args(arg_strings, None)
[...]

PR #219, opened 2020-03-15
Add script schema

vsoch:
This is the first round of changes to add the script schema. The PR includes:

 - the version 0.0.1 of the script schema added, and used to run tests
 - refactored build flow in menu using a new builder class
 - fix to #212 
 - fix to #215 to show a schema (we might still want to remove, but in another PR)
 - will close https://github.com/HPC-buildtest/buildtest-framework/issues/210

I suspect since there are likely over a thousand lines changed not all tests will work, I'll update the PR as needed. **I want to also suggest for this PR that if we do a squash and merge I do the final one, so that I don't lose credit for my work.**

Consider the PR a WIP until tests pass. When tests pass, I'd like to suggest that for nitpicks / further work, we do this in well scoped, additional PRs. The reason is that I want unblock you for the two PRs currently open, and this one is getting a bit large and overwhelming to work with!


buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+
+known_sections = variable_sections + build_sections
+
+
+class BuildConfig:
+    """A BuildConfig is a base class for a build test configuration.
+       The type (e.g., script) and version are derived from reading in
+       the file, and then matching to a buildtest schema, each of which is
+       developed at https://github.com/HPC-buildtest/schemas and added to
+       subfolders named accordingly under buildtest/tools/buildsystem/schemas.
```
shahzebsiddiqui: change **buildtest/tools/buildsystem/schemas** --> **buildtest/buildsystem/schemas**

buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json
```diff
@@ -0,0 +1,57 @@
...
+           "pattern": "^[A-Za-z_][A-Za-z0-9_]*$"
+        }
+      }
+    },
+    "shell": {
+      "type": "string",
+      "default": "bash",
+      "description": "The shell interpreter to use."
+    },
+    "pre_run": {
```
shahzebsiddiqui, line 44: Let's remove ``pre_run`` and ``post_run`` from script schema definition entirely, it doesn't belong here in the first commit. ``run`` should suffice for everything .

buildtest/buildsystem/schemas/utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def get_schema_version(schema):
+    """Given a schema file, use the convention naming of
+       <name>-v<version>.schema.json to derive the version
+    """
+    match = re.search(
+        "v(?P<version>[0-9]{1}[.][0-9]{1}[.][0-9]{1})[.]schema[.]json", schema
+    )
+    if match:
```
shahzebsiddiqui, line 72: just a thought if we don't have a match ``if match`` what do we want to return considering ``get_schema_version`` is used by ``get_latest`` to calculate latest version. If there is no return value then ``get_latest`` won't calculate. I assume it is unlikely since we are populating the schema files. Maybe raise an error would be best.

buildtest/defaults.py
```diff
@@ -10,12 +10,21 @@
 
 logID = "buildtest"
 
+# each has a subfolder in buildtest/buildsystem/schemas/ with *.schema.json
+supported_schemas = ["script"]
+
+# global config sections that are known, added to buildbase.metadata
+variable_sections = ["env"]
+build_sections = ["pre_build", "build", "post_build", "pre_run", "run", "post_run"]
```
shahzebsiddiqui, line 18: for now, let's remove all the extra build sections except for ``run``. I think that should be sufficient for this PR to get ``script`` schema working.

tests/test_buildsystem.py
```diff
@@ -0,0 +1,59 @@
+from buildtest.buildsystem.base import BuildConfig, supported_schemas
```
shahzebsiddiqui: can we swap the import lines as follows:
 
```
import pytest
import os

from buildtest.buildsystem.base import BuildConfig, supported_schemas
```

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+
+A cd to the $TESTDIR is always the first step taken, and any defined
+sections after that are added if not empty.
+The following sections can be defined, and serve different purposes:
+
+ - env
+
+"""
+
+from copy import deepcopy
```
shahzebsiddiqui: can we put this line after all the ``import`` statement to be consistent

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+ - post_build
+ - pre_run
+ - run
+ - post_run
+
+A cd to the $TESTDIR is always the first step taken, and any defined
+sections after that are added if not empty.
+The following sections can be defined, and serve different purposes:
+
+ - env
```
shahzebsiddiqui: I think this section is unnecessary and best if its removed, it confuses the reader.  A brief description of the classes 
``BuildConfig`` and ``BuilderBase`` would be appropriate.

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+    load_recipe,
+    get_schemas_available,
+    here,
+)
+
+known_sections = variable_sections + build_sections
+
+
+class BuildConfig:
+    """A BuildConfig is a base class for a build test configuration.
```
shahzebsiddiqui: change ``build test configuration`` --> ``test configuration``. It can be confusing when using the words 

**build test configuration** or **buildtest configuration**  or **buildtest's configuration**

For now, i am calling the YAML files (test configuration). It's worth documenting this in our terminology section (later) and update all of our docstrings where appropriate. I'd be open to a more catchy name.

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+       The type (e.g., script) and version are derived from reading in
+       the file, and then matching to a buildtest schema, each of which is
+       developed at https://github.com/HPC-buildtest/schemas and added to
+       subfolders named accordingly under buildtest/tools/buildsystem/schemas.
+       The schema object can load in a general test configuration file
+       to validate it, and then match it to a schema available.
+       If the version of a schema is not specified, we use the latest.
+       If the schema fails validation, we also don't continue.
```
shahzebsiddiqui: bit wordy **we also don't continue**. We could just say: **If the schema fails validation check, then we stop immediately.**

buildtest/buildsystem/schemas/utils.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    if not os.path.exists(path):
+        sys.exit("schema file %s does not exist." % path)
+    with open(path, "r") as fd:
+        schema = json.loads(fd.read())
+    return schema
+
+
+def load_recipe(path):
+    """load a yaml recipe file. The recipe is validated against a schema.
```
shahzebsiddiqui, line 31: can we add description of parameters, type,  return, and rtype

buildtest/buildsystem/schemas/utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+here = os.path.dirname(os.path.abspath(__file__))
+
+
+def load_schema(path):
+    """load a schema from file. We assume a json file
+
+       Parameters
+       ==========
+       path: the path to the schema file
```
shahzebsiddiqui: can you add ``type path: str, required``

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+    metadata = ["version"]
+
+    def __init__(self, config_file):
+        """initiate a build configuration file, meaning that we read in the
+           file, match it to a schema provided by buildtest, and 
+           validate it
+
+           Parameters
+           ==========
+           config_file: the pull path to the configuration file, must exist.
```
shahzebsiddiqui, line 68: can we also put the ``:type config_file: str, required``

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+            schema_file = os.path.join(
+                here, section["type"], self.lookup[section["type"]][version]
+            )
+            validate(instance=section, schema=load_schema(schema_file))
+
+    def _validate_global(self, config_file=None):
+        """the global validation ensures that the overall structure of the
+           file is sound for further parsing. We load in the outer.schema.json
+           for this purpose. The function also allows a custom config to be 
+           to extend the usage of the class.
```
shahzebsiddiqui, line 148: can you define the parameter ``config_files`` in docstring

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+        return "[builder-%s-%s]" % (self.type, self.name)
+
+    def __repr__(self):
+        return self.__str__()
+
+    def _create_test_folders(self):
+        """Create all needed test folders on init, and add their paths
+           to self.metadata
+        """
+        testpath = os.path.expandvars(self.metadata["testpath"])
```
shahzebsiddiqui, line 258: should we also account for tilde (``~``) character so also check for ``os.path.expanduser`` https://docs.python.org/3.7/library/os.path.html#os.path.expanduser

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+        testdir = os.path.dirname(testpath)
+        create_dir(testdir)
+        for folder in ["run", "log"]:
+            name = "%sdir" % folder
+            self.metadata[name] = os.path.join(testdir, folder)
+            create_dir(self.metadata[name])
+
+    def _init_logger(self, to_file=True):
+        """initialize the logger. This is called at the end of prepare_run,
```
shahzebsiddiqui, line 267: Can we define the parameter ``to_file`` in the docstring

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+
+        # If we are using a python shell, we need to import os first
+        if "python" in shell:
+            env.append("import os")
+
+        # Parse environment depending on expected shell
+        if pairs:
+
+            # Handles csh and tsch
+            if "csh" in shell:
```
shahzebsiddiqui, line 306: might as well add **tcsh** in the if condition

buildtest/buildsystem/base.py
```diff
@@ -0,0 +1,558 @@
...
+            stat.S_IRWXU | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH,
+        )
+        print("{:<40} {}".format("[WRITING TEST]", "PASSED"))
+        return testpath
+
+    def _get_test_lines(self):
+        """Given test metadata, get test lines to write to file or show.
+        """
+        # Every test starts with cd to TESTDIR
+        lines = ["cd $TESTDIR"]
```
shahzebsiddiqui, line 532: I am now wondering if it makes sense to add ``cd $TESTDIR`` to all test-scripts. I think the purpose was to ensure any output, error or log files are not put in current directory which could be anywhere depending on where user ran the command. I think buildtest should correctly write all .out, .err, and .log and any other files in same or sub-directory where test exists.

shahzebsiddiqui (changes requested):
@vsoch Overall, this looks a lot better than what we have. I made a few comments if you can kindly go over them.

I noticed a few things related to docstrings that we may want to standardize upon.
- Some docstrings first letter started with lower case while some are capital case, we need to fix that to one style
- Some docstrings didn't have a trailing period while some did
- If there are any missing documentation on parameters description, type, return value, it should be added to docstring

Other than that, i have to test this locally which i wasn't able to do but looking at coverage report it looks promising.

vsoch:
@shahzebsiddiqui I agree on having a standard for docstrings, and:

 1. it should be handled in a separate PR
 2. it should be enforced with a tool, for example URL checker used this. https://app.codacy.com/gh/urlstechie/URLs-checker/pullRequest?prid=5180802

So for now, albeit they are imperfect, I'll open up an issue to discuss this further and then a separate PR down the line to add the integration and address the changes it requests.

buildtest/buildsystem/base.py
vsoch: Fixed locally.

buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json
vsoch: This is a reasonable request, it will need to go through in the schemas repo first. Why did you approve there if you didn't ultimately agree?

vsoch: I think it's a mistake to remove these, because we need to be able to support the user and test config being able to set precedence for the order of things. For example, a recipe might have the user put commands to do some git or other preparatory steps in pre_run, and then the run command is generated to be some default for the config, and then post_run might be a custom cleanup desired by the user. If we only provide one section, run, the user is going to give a dump of commands and the parser of those lines will have no way to separate them into logical stages. Akin to how spack has clearly defined stages, we at least need simple pre, and post ones here.

vsoch: It also does not hurt to have them, because there is no extra dependency to having them, the user doesn't need to define them, and minimally it just exposes these sections if there is some pre or post command that the user wants to run.

buildtest/buildsystem/schemas/utils.py
vsoch: A schema that is validated / tested from the schemas repository must conform to that file name pattern.

buildtest/defaults.py
vsoch: I disagree here, I think we need support for different stages, to allow for a custom test config parser to use them as needed  (e.g., when something needs to be compiled, or the user needs to run some pre_build command to obtain a dependency). If we remove them now they will need to be added later, and I've already done all the careful work here and in the schemas repository to support these sections. There is no harm in keeping them as they are just optional sections in a recipe (and note that the build_x sections aren't even defined for script!) and give test config parsers that we write in the future power to support ordering of commands. I added these explicitly after seeing the examples that require several stages of a build, and I see how they will integrate with these new parsers (not written yet).

tests/test_buildsystem.py
vsoch: lol sure why not!

buildtest/buildsystem/base.py
vsoch: Let's handle all docstrings in another PR, there is too much linting / fixes that need to be done and since it's globally across the library it's scoped outside of this work.

buildtest/buildsystem/base.py
vsoch: I think we should wait for this to show up in the wild before implementing it like that.

buildtest/buildsystem/base.py
vsoch: Another PR.

buildtest/buildsystem/base.py
vsoch: See the comment.

buildtest/buildsystem/base.py
vsoch: This would fundamentally change the behavior of the library, and basically eliminate the need to store test results in the buildtest home. I think this is a good idea, but since it's a bit of a refactor of the library, it should be addressed in another PR, with an issue opened first for discussion.

shahzebsiddiqui (approved):
@vsoch this looks good to me. Thanks for your work. I think we can proceed with a merge. Feel free to Squash Commit

PR #218, opened 2020-03-13
reorg regtests based on source file to be tested 

shahzebsiddiqui:
This PR is mapping all the unit tests to the source file to be tested. The structure of the tests looks as follows

```
(docs) ssi29@ag-mxg-hulk090> tree -P test_*.py tests/
tests/
├── menu
│   └── test_config.py
├── test_defaults.py
├── test_exceptions.py
└── utils
    ├── test_command.py
    └── test_file.py

2 directories, 5 files
```

PR #222, opened 2020-03-18
Update Regression Test Docs

shahzebsiddiqui:
This PR does the following
- Update the Regression Testing Docs to address #218 and how user can get started
- Remove unused imports throughout project thanks to PyCharm
- Minor tweak to conf.py

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -9,13 +9,10 @@ expose functions to run builds.
 """
 
 import datetime
-import json
 import os
 import re
 import stat
-import subprocess
 import sys
-import yaml
 
 from copy import deepcopy
 from jsonschema import validate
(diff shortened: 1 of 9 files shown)


vsoch:
@shahzebsiddiqui I can review this when the failed tests in the CI above are addressed.

shahzebsiddiqui:
> @shahzebsiddiqui I can review this when the failed tests in the CI above are addressed.

I dont think there is an issue with CI tests according to Build 294 https://travis-ci.com/github/HPC-buildtest/buildtest-framework/builds/153843814 all Travis tests passed. 
Coveralls is failing but doesn't show anything outstanding why its failing other than decrease on coverage report which is insignificant. 

Codefactor is not important, i probably should remove this autocheck. It was handy but now i rarely look at it.

vsoch:
okay I'm good with merge, but could we please remove tests that are no longer considered important? It feels like noise at this point.

vsoch (approved):

shahzebsiddiqui:
> okay I'm good with merge, but could we please remove tests that are no longer considered important? It feels like noise at this point.

I agree, i will remove some of the regtest that are not needed. It will lower coverage but it's ok. Some of these are overhead and maybe its the feature that needs to be removed. I will look through the tests and see what can be removed.

Issue #227, opened 2020-03-18
missing magic string in test script


shahzebsiddiqui:
Test needs to have magic string ``#!/bin/bash`` or ``#!/bin/csh``. This is important specify when it comes to using modules in Lmod see https://lmod.readthedocs.io/en/latest/040_FAQ.html. 

Here is an example generated file

```
cd $TESTDIR
FILE=1mb_bzip2.txt
TARFILE=1mb_bzip2.txt.bz2
dd if=/dev/zero of=$FILE bs=1M count=1
tar cvjf $TARFILE $FILE
ll $TARFILE
bzip2 -d $TARFILE
```

vsoch:
Haha, I've never heard "magic string" - I think this is called the shebang, or bang line. See https://bash.cyberciti.biz/guide/Shebang.

vsoch:
This would easily be parsed from the `executable` variable derived from the command, since it looks for the full path of a shell. The command could be modified to do the parsing (and get this executable) in a separate function so that it can be run without executing the command. I can do this later this week if you like.

shahzebsiddiqui:
I must have heard this "magic string" from one of the instructor videos I took. Yes shebang is it.

I was going to do it tomorrow, but if you have started then I can wait. I also want to get rid of cd TESTDIR. 

That doesn't work and there is no such variable.

vsoch:
I wouldn't get rid of cd TESTDIR, that will be needed. What will change is what exactly TESTDIR is, it likely will be derived custom based on the relative path of the file to .buildtest. If TESTDIR isn't there it's because the environment exports are missing.

shahzebsiddiqui:
if it's okay let me work on this, i won't remove the TESTDIR for time being.

vsoch:
You'll need to update the BuildTestCommand to extract some executable (the full path) based on a shell, and then use that full path for the shabang. What we cannot do is take a shell preference and assume /usr/local/bin or /usr/bin, etc.

vsoch:
We've addressed this, closing.

PR #229, opened 2020-03-19
Adds shebang in test, change format name of generated testscript.

shahzebsiddiqui:
1. Adds shebang #!/bin/bash and #!/bin/csh in the test
2. Correctly adds shell extension .csh when shell: csh is selected
3. Remove schema name embedded in test, Now test name is just name of YAML file + datetime
4. Fix all single line docstring, all docstring should start with Capital letter and end with a period at end

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -546,9 +546,9 @@ class BuilderBase:
 
         if self.get_shell() in ["sh", "/bin/sh"]:
             lines += [f"#!/bin/sh"]
-            
+
         elif self.get_shell() in ["bash", "/bin/bash"]:
-            lines += [BUILDTEST_SHELL]
+            lines += [f"#!{BUILDTEST_SHELL}"]
 
         # Every test starts with cd to TESTDIR
         lines += ["cd $TESTDIR"]
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -382,7 +382,7 @@ class BuilderBase:
 
         # Derive the path to the test script
         self.metadata["testpath"] = "%s.%s" % (
-            os.path.join(self.options["build"]["testdir"], self.name, self.build_id),
+            os.path.join(self.options["build"]["testdir"], self.name),
             self.get_test_extension(),
         )
         self.metadata["testdir"] = os.path.dirname(self.metadata["testpath"])
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -54,8 +54,7 @@ class BuildConfig:
 
     def __init__(self, config_file):
         """initiate a build configuration file, meaning that we read in the
-           file, match it to a schema provided by buildtest, and 
-           validate it
+           file, match it to a schema provided by buildtest, and validate it
 
            Parameters:
 
@@ -258,7 +257,7 @@ class BuilderBase:
             create_dir(self.metadata[name])
 
     def _init_logger(self, to_file=True):
-        """initialize the logger. This is called at the end of prepare_run,
+        """Initialize the logger. This is called at the end of prepare_run,
            so the testpath is defined with the build id.
         """
         if to_file:
@@ -276,6 +275,8 @@ class BuilderBase:
         shell = self.get_shell()
         if "python" in shell:
             return "py"
+        if "csh" in shell:
+            return "csh"
         return "sh"
 
     def get_environment(self):
@@ -316,8 +317,8 @@ class BuilderBase:
         return env
 
     def get_shell(self):
-        """Return the shell defined in the recipe, or default to bash
-        """
+        """Return the shell defined in the recipe, or default to bash"""
+
         return self.recipe.get("shell", BUILDTEST_SHELL)
 
     def run_wrapper(func):
@@ -339,8 +340,9 @@ class BuilderBase:
 
     def finish_run(self):
         """Finish up the run (not sure what might go here yet, other than
-           honoring a custom subclass function
+           honoring a custom subclass function.
         """
+
         # If the subclass has a _finish_run function, honor it
         if hasattr(self, "_finish_run"):
             self._finish_run()
@@ -348,8 +350,9 @@ class BuilderBase:
     def prepare_run(self):
         """Prepare run provides shared functions to set up metadata and
            class data structures that are used by both run and dry_run
-           This section cannot be reached without a valid, loaded recipe
+           This section cannot be reached without a valid, loaded recipe.
         """
+
         # Generate a unique id for the build based on key and unique string
         self.build_id = self._generate_build_id()
 
@@ -393,6 +396,7 @@ class BuilderBase:
            This parent class handles shared starting functions for each step
            and then calls the subclass function (_run) if it exists.
         """
+
         # Create test directory and run folder they don't exist
         self._create_test_folders()
         self._init_logger()
@@ -403,8 +407,9 @@ class BuilderBase:
     def run_tests(self, testfile):
         """The shared _run function will run a test file, which must be
            provided. This is called by run() after generation of the
-           test file, and it return a result object (dict). The 
+           test file, and it return a result object (dict).
         """
+
         # Keep a result object
         result = {}
         result["START_TIME"] = self.get_formatted_time("start_time")
@@ -466,8 +471,8 @@ class BuilderBase:
         return timestamp
 
     def show_prepare(self):
-        """Print basic run information to the user, if defined, before the run.
-        """
+        """Print basic run information to the user, if defined, before the run."""
+
         start_time = self.get_formatted_time("start_time")
 
         print("{:_<80}".format(""))
@@ -489,20 +494,21 @@ class BuilderBase:
         """Akin to a build preview, we prepare and finish a run, but only
            print the script to the screen without writing or running files.
         """
+
         # Dry run just prints the testing script
         self._init_logger(to_file=False)
         lines = self._get_test_lines()
         print("\n".join(lines))
 
     def _generate_build_id(self):
-        """Generate a build id based on the recipe name, type, and datetime
-        """
-        now = datetime.datetime.now().strftime("%m-%d-%Y-%H-%M-S")
-        return "%s_%s_%s" % (self.name, self.type, now)
+        """Generate a build id based on the recipe name, and datetime."""
+
+        now = datetime.datetime.now().strftime("%m-%d-%Y-%H-%M")
+        return "%s_%s" % (self.name, now)
 
     def _write_test(self, lines=None):
-        """Given test metadata, write test content.
-        """
+        """Given test metadata, write test content."""
+
         # If no lines provided, generate
         if not lines:
             lines = self._get_test_lines()
@@ -527,10 +533,15 @@ class BuilderBase:
         return testpath
 
     def _get_test_lines(self):
-        """Given test metadata, get test lines to write to file or show.
-        """
+        """Given test metadata, get test lines to write to file or show."""
+
         # Every test starts with cd to TESTDIR
-        lines = ["cd $TESTDIR"]
+        lines = ["#!/bin/bash"]
+
+        if self.get_shell() == "csh":
+            lines = ["#!/bin/csh"]
+
+        lines += ["cd $TESTDIR"]
 
         # Add environment variables
         lines += self.get_environment()
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -541,10 +541,16 @@ class BuilderBase:
 
     def _get_test_lines(self):
         """Given test metadata, get test lines to write to file or show."""
+        
+        lines = []
+
+        if self.get_shell() in ["sh", "/bin/sh"]:
+            lines += [f"#!/bin/sh"]
+            
+        elif self.get_shell() in ["bash", "/bin/bash"]:
+            lines += [BUILDTEST_SHELL]
 
         # Every test starts with cd to TESTDIR
-        lines = ["#!/bin/bash"]
-
         lines += ["cd $TESTDIR"]
 
         # Add environment variables
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -11,6 +11,7 @@ expose functions to run builds.
 import datetime
 import os
 import re
+import shutil
 import stat
 import sys
 
@@ -543,12 +544,13 @@ class BuilderBase:
         """Given test metadata, get test lines to write to file or show."""
         
         lines = []
+        shell = shutil.which(self.get_shell())
+        
+        if not shell:
+            shell = BUILDTEST_SHELL
 
-        if self.get_shell() in ["sh", "/bin/sh"]:
-            lines += [f"#!/bin/sh"]
-
-        elif self.get_shell() in ["bash", "/bin/bash"]:
-            lines += [f"#!{BUILDTEST_SHELL}"]
+        lines += [f"#!{shell}"]
+        
 
         # Every test starts with cd to TESTDIR
         lines += ["cd $TESTDIR"]
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -54,7 +54,7 @@ class BuildConfig:
 
     def __init__(self, config_file):
         """initiate a build configuration file, meaning that we read in the
-           file, match it to a schema provided by buildtest, and validate it
+           file, match it to a schema provided by buildtest, and validate it.
 
            Parameters:
 
@@ -69,8 +69,8 @@ class BuildConfig:
         self.load(config_file)
 
     def load(self, config_file):
-        """load a config file. We check that it exists, and that it is valid.
-           we exit on any error, as the config is not loadable
+        """Load a config file. We check that it exists, and that it is valid.
+           we exit on any error, as the config is not loadable.
 
            Parameters:
 
@@ -105,6 +105,7 @@ class BuildConfig:
            to buildtest. If a version is provided, honor it. If not, use latest.
            We also don't allow repeated keys in the same file.
         """
+
         version = self.recipe.get("version", "latest")
         seen = set()
         for name, section in self.recipe.items():
@@ -135,11 +136,12 @@ class BuildConfig:
             validate(instance=section, schema=load_schema(schema_file))
 
     def _validate_global(self, config_file=None):
-        """the global validation ensures that the overall structure of the
+        """The global validation ensures that the overall structure of the
            file is sound for further parsing. We load in the outer.schema.json
            for this purpose. The function also allows a custom config to be 
            to extend the usage of the class.
         """
+
         config_file = config_file or self.config_file
         outer_schema = load_schema(os.path.join(here, "outer.schema.json"))
         self.recipe = load_recipe(config_file)
@@ -158,6 +160,7 @@ class BuildConfig:
            for each based on the type. Each type is associated with a known 
            Builder class.
         """
+
         builders = []
         if self.recipe:
             for name in self.keys():
@@ -180,6 +183,7 @@ class BuildConfig:
         """Return the list of keys for the loaded recipe, not including
            the metadata keys defined for any global recipe.
         """
+
         keys = []
         if self.recipe:
             keys = [x for x in self.recipe.keys() if x not in self.metadata]
@@ -188,8 +192,9 @@ class BuildConfig:
     def get(self, name):
         """Given the name of a section (typically a build configuration name)
            return the loaded section from self.recipe. If you need to parse
-           through just section names, use self.keys() to filter out metadata
+           through just section names, use self.keys() to filter out metadata.
         """
+
         return self.recipe.get(name)
 
 
@@ -199,7 +204,7 @@ class BuilderBase:
     """
 
     def __init__(self, name, recipe_config, config_file=None):
-        """initiate a builder base. A recipe configuration (loaded) is required.
+        """Initiate a builder base. A recipe configuration (loaded) is required.
            this can be handled easily with the BuildConfig class:
 
            bc = BuildConfig(config_file)
@@ -212,6 +217,7 @@ class BuilderBase:
            recipe_config: the loaded section from the config_file for the user.
            config_file: the pull path to the configuration file, must exist.
         """
+
         self.name = name
         self.result = {}
         self.build_id = None
@@ -246,8 +252,9 @@ class BuilderBase:
 
     def _create_test_folders(self):
         """Create all needed test folders on init, and add their paths
-           to self.metadata
+           to self.metadata.
         """
+
         testpath = os.path.expandvars(self.metadata["testpath"])
         testdir = os.path.dirname(testpath)
         create_dir(testdir)
@@ -260,6 +267,7 @@ class BuilderBase:
         """Initialize the logger. This is called at the end of prepare_run,
            so the testpath is defined with the build id.
         """
+
         if to_file:
             self.metadata["logfile"] = os.path.join(
                 self.metadata["logdir"], "%s.log" % self.build_id
@@ -269,9 +277,14 @@ class BuilderBase:
             self.logger = init_log()
 
     def get_test_extension(self):
-        """Return the test extension, which depends on the shell used. We
-           return .sh for all shell types except for python, we return .py.
+        """Return the test extension, which depends on the shell used. Based
+           on the value of ``shell`` key we return the shell extension.
+
+           shell: csh --> csh
+           shell: python --> py
+           shell: bash --> sh (default)
         """
+
         shell = self.get_shell()
         if "python" in shell:
             return "py"
@@ -317,7 +330,7 @@ class BuilderBase:
         return env
 
     def get_shell(self):
-        """Return the shell defined in the recipe, or default to bash"""
+        """Return the shell defined in the recipe, or default to bash."""
 
         return self.recipe.get("shell", BUILDTEST_SHELL)
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -280,7 +280,6 @@ class BuilderBase:
         """Return the test extension, which depends on the shell used. Based
            on the value of ``shell`` key we return the shell extension.
 
-           shell: csh --> csh
            shell: python --> py
            shell: bash --> sh (default)
         """
@@ -288,8 +287,7 @@ class BuilderBase:
         shell = self.get_shell()
         if "python" in shell:
             return "py"
-        if "csh" in shell:
-            return "csh"
+
         return "sh"
 
     def get_environment(self):
@@ -310,12 +308,8 @@ class BuilderBase:
         # Parse environment depending on expected shell
         if pairs:
 
-            # Handles csh and tsch
-            if "csh" in shell:
-                [env.append("setenv %s %s" % (k, v)) for k, v in pairs.items()]
-
             # Handles bash and sh
-            elif re.search("(bash|sh)$", shell):
+            if re.search("(bash|sh)$", shell):
                 [env.append("%s=%s" % (k, v)) for k, v in pairs.items()]
 
             # python interpreter can be a shell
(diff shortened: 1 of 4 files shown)


buildtest/buildsystem/base.py
```diff
@@ -527,10 +533,15 @@ def _write_test(self, lines=None):
...
-        """Given test metadata, get test lines to write to file or show.
-        """
+        """Given test metadata, get test lines to write to file or show."""
+
         # Every test starts with cd to TESTDIR
-        lines = ["cd $TESTDIR"]
+        lines = ["#!/bin/bash"]
+
+        if self.get_shell() == "csh":
+            lines = ["#!/bin/csh"]
```
vsoch: This is what I was saying to avoid - can we always be sure bash and csh are in /usr/bin? We need to find the correct executable before just hard coding one that is an assumption. Take a look at the `executable` variable in the BuildTestCommand to see what I mean. We don't want to hard code something like /bin/csh only to discover that the user doesn't have the binary there.

vsoch (changes requested):

buildtest/buildsystem/base.py
```diff
@@ -276,6 +275,8 @@ def get_test_extension(self):
         shell = self.get_shell()
         if "python" in shell:
             return "py"
+        if "csh" in shell:
```
vsoch: I think the .sh extension is fine for csh as well, it's intended to be portable. See https://unix.stackexchange.com/questions/31760/file-extensions-for-unix-shell-scripts. If we do this, then we would need to honor all the other sub-flavors of shells, which doesn't add value. It's the interpreter at the top that matters.

buildtest/buildsystem/base.py
shahzebsiddiqui: bash and csh are almost always going to be in /bin/csh and /bin/bash. By all means the rpm `tcsh` and `bash` are installed in all systems i have been, without it you can't really do anything. In fact root user is setup with one of the default shell usually bash .

```
(docs) login1.stampede2(1097)$ cat /etc/passwd | grep root
root:x:0:0:root:/root:/bin/bash
```


 Even a bare minimal OS will have these rpms installed.

```
(docs) login1.stampede2(1091)$ rpm -ql tcsh | grep "/bin/*"
/bin/csh
/bin/tcsh

(docs) login1.stampede2(1096)$ rpm -ql bash | grep "/bin/bash"
/usr/bin/bash
/usr/bin/bashbug
/usr/bin/bashbug-64
``` 

Not all systems have tcsh but bash is there all the time. 

I realize that `csh` support can add extra work for buildtest to generate test properly. I have went through this in past, if user doesnt really care then I am all for removing the feature and keeping it simple.

Let me know your thoughts.

buildtest/buildsystem/base.py
shahzebsiddiqui: Probably makes sense to remove .csh feature if you agree. It adds extra work for buildtest to support multiple shells. I knew what we were getting into when adding support for csh

shahzebsiddiqui:
I removed the `csh` support it just adds extra complexity not needed for now. I also fixed #230. For time being.

buildtest/buildsystem/base.py
vsoch: okay, this is reasonable. We can implement like this until it breaks for someone, and address then I suppose.

buildtest/buildsystem/base.py
vsoch: I think it's okay to keep it :) I'm just saying for this function (that returns an extension) we can just return .sh.

shahzebsiddiqui:
I realize there could be one more thing that could be fixed. The testscript are duplicated when repetitive runs because of timestamp in shell script. I think the test names should be same if repetitive runs are done, but output, error and log file should have a timestamp to differentiate files. It helps with when file was run. I can make this quick fix before merging.

vsoch:
okay so I think we need to address one major bug before pushing this through. First, clarify if you really want to remove csh (I don't see why it's important at this point if we can add the path to the csh binary and it's just a string and will support a subset of users). And then if there is really this level of strictness with respect to shells, the shells allowed needs to be added to the schemas here https://github.com/HPC-buildtest/schemas/blob/master/script/script-v0.0.1.schema.json#L42 and then validation of a recipe will fail off the bat if an invalid shell is given. I'd be happy to review that PR if you want to put it into the schemas repo.

vsoch:
I agree on the test script naming - I had thought you wanted them unique, but that is definitely not a great way to go about it.

shahzebsiddiqui:
> okay so I think we need to address one major bug before pushing this through. First, clarify if you really want to remove csh (I don't see why it's important at this point if we can add the path to the csh binary and it's just a string and will support a subset of users). And then if there is really this level of strictness with respect to shells, the shells allowed needs to be added to the schemas here https://github.com/HPC-buildtest/schemas/blob/master/script/script-v0.0.1.schema.json#L42 and then validation of a recipe will fail off the bat if an invalid shell is given. I'd be happy to review that PR if you want to put it into the schemas repo.

I'd say if the feature is needed to support `csh` then we can add it. It's partly my fault having this in first place, i dont even need it and most folks don't care. I am sure i'd not want to add extra complexity and be in favor of keeping things as simple as possible. The scope of this PR was addressing the shebang for bash, there are few other bugs addressed.

shahzebsiddiqui:
> I agree on the test script naming - I had thought you wanted them unique, but that is definitely not a great way to go about it.

Ill make that change now

vsoch:
> I'd say if the feature is needed to support csh then we can add it. It's partly my fault having this in first place, i dont even need it and most folks don't care. I am sure i'd not want to add extra complexity and be in favor of keeping things as simple as possible. The scope of this PR was addressing the shebang for bash, there are few other bugs addressed.

okay, I agree! So what would happen now if the user specified a csh (or other) shell - would it return None (and essentially break?) We need to catch this - and I think the best place to handle it is updating the schema so it doesn't validate and doesn't go beyond there. That's the nice thing about the schemas - we don't have to write custom logic in the python code here, just update the specifications.

shahzebsiddiqui:
@vsoch  its fixed now, the test name is now without timestamp. For now it's still a possibility that we get a name conflict but it is low. It would help maybe having parent directory to categorize tests. It's not clear we can leave it be as for now.

```
(docs) login1.stampede2(1144)$ buildtest build -c examples/script/slurm.yml
                             start time: 2020-03-19 16:06:58.708825
                     configuration file: /work/06908/sms1990/stampede2/buildtest-framework/examples/script/slurm.yml
                                testdir: /home1/06908/sms1990/.buildtest/testdir
                               testpath: /home1/06908/sms1990/.buildtest/testdir/slurm_down_nodes_reason.sh
                                logpath: None



STAGE                                    VALUE
[WRITING TEST]                           PASSED
[RUNNING TEST]                           PASSED
Writing results to /home1/06908/sms1990/.buildtest/testdir/run/slurm_down_nodes_reason_03-19-2020-16-06.out
                             start time: 2020-03-19 16:06:58.913396
                     configuration file: /work/06908/sms1990/stampede2/buildtest-framework/examples/script/slurm.yml
                                testdir: /home1/06908/sms1990/.buildtest/testdir
                               testpath: /home1/06908/sms1990/.buildtest/testdir/slurm_not_responding_nodes.sh
                                logpath: None



STAGE                                    VALUE
[WRITING TEST]                           PASSED
[RUNNING TEST]                           PASSED
Writing results to /home1/06908/sms1990/.buildtest/testdir/run/slurm_not_responding_nodes_03-19-2020-16-06.out
Finished running 2 total tests.
                         Test summary
Executed 2 tests
Passed Tests: 2 Percentage: 100.0%
Failed Tests: 0 Percentage: 0.0%

```

vsoch:
This is what I have in mind: https://github.com/HPC-buildtest/schemas/pull/18

If you want the user to be able to specify also /bin/bash or /bin/sh we can add those too, but I think we should keep it simple and require just the name (bash or python, etc).

shahzebsiddiqui:
@vsoch so i tested it here is the result

```
(docs) login1.stampede2(1151)$ buildtest build -d -c  examples/script/bzip2.yml
                             start time: 2020-03-19 16:11:58.404151
                     configuration file: /work/06908/sms1990/stampede2/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home1/06908/sms1990/.buildtest/testdir
                               testpath: /home1/06908/sms1990/.buildtest/testdir/bzip_compression.sh
                                logpath: None



STAGE                                    VALUE
#!/bin/bash
cd $TESTDIR
FILE=1mb_bzip2.txt
TARFILE=1mb_bzip2.txt.bz2
dd if=/dev/zero of="$FILE" bs=1M count=1
tar cvjf "$TARFILE" "$FILE"
ls -l "$TARFILE"

(docs) login1.stampede2(1152)$ buildtest build -c  examples/script/bzip2.yml
                             start time: 2020-03-19 16:12:12.544319
                     configuration file: /work/06908/sms1990/stampede2/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home1/06908/sms1990/.buildtest/testdir
                               testpath: /home1/06908/sms1990/.buildtest/testdir/bzip_compression.sh
                                logpath: None



STAGE                                    VALUE
[WRITING TEST]                           PASSED
[RUNNING TEST]                           FAILED
Writing results to /home1/06908/sms1990/.buildtest/testdir/run/bzip_compression_03-19-2020-16-12.out
Finished running 1 total tests.
                         Test summary
Executed 1 tests
Passed Tests: 0 Percentage: 0.0%
Failed Tests: 1 Percentage: 100.0%
(docs) login1.stampede2(1153)$ cat examples/script/bzip2.yml
version: 0.0.1

bzip_compression:
  type: script
  env:
    FILE: 1mb_bzip2.txt
    TARFILE: 1mb_bzip2.txt.bz2
  shell: csh
  run: |
    dd if=/dev/zero of="$FILE" bs=1M count=1
    tar cvjf "$TARFILE" "$FILE"
    ls -l "$TARFILE"


```

The test fails which is ok, not an issue with buildtest its the way script (``run``) section is written.

shahzebsiddiqui:
I updated the schema and tested locally with ``csh`` and it looks good. Schema is now updated see https://github.com/HPC-buildtest/buildtest-framework/pull/229/commits/8874aa4ce826c1bb16f5c51d921ec20b3833f6e6

```
(docs) login1.stampede2(1166)$ buildtest build -c examples/script/bzip2.yml
  File "/home1/06908/sms1990/.local/share/virtualenvs/docs-qbK3SfF3/bin/buildtest", line 11, in <module>
    load_entry_point('buildtest-framework', 'console_scripts', 'buildtest')()
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/main.py", line 20, in main
    parser.parse_options()
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/menu/__init__.py", line 81, in parse_options
    args.func(args)
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/menu/build.py", line 88, in func_build_subcmd
    bc = BuildConfig(config_file)
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/buildsystem/base.py", line 69, in __init__
    self.load(config_file)
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/buildsystem/base.py", line 93, in load
    self._validate()
  File "/work/06908/sms1990/stampede2/buildtest-framework/buildtest/buildsystem/base.py", line 136, in _validate
    validate(instance=section, schema=load_schema(schema_file))
  File "/home1/06908/sms1990/.local/share/virtualenvs/docs-qbK3SfF3/lib/python3.7/site-packages/jsonschema/validators.py", line 899, in validate
    raise error
jsonschema.exceptions.ValidationError: 'csh' is not one of ['bash', 'sh', '/bin/bash', '/bin/sh']

Failed validating 'enum' in schema['properties']['shell']:
    {'default': 'bash',
     'description': 'The shell interpreter to use.',
     'enum': ['bash', 'sh', '/bin/bash', '/bin/sh'],
     'type': 'string'}

On instance['shell']:
    'csh'

```

buildtest/buildsystem/base.py
```diff
@@ -527,10 +540,12 @@ def _write_test(self, lines=None):
         return testpath
 
     def _get_test_lines(self):
-        """Given test metadata, get test lines to write to file or show.
-        """
+        """Given test metadata, get test lines to write to file or show."""
+
         # Every test starts with cd to TESTDIR
-        lines = ["cd $TESTDIR"]
+        lines = ["#!/bin/bash"]
```
vsoch: How does this honor the user choice if the choice is /bin/sh?

shahzebsiddiqui: so would you want the following
``shell: bash`` or ``shell: /bin/bash`` --> ``#!/bin/bash``
``shell: sh`` or ``shell: /bin/sh`` --> ``#!/bin/sh``. 

Is that right?

vsoch:
I mentioned this before, but the `BuildTestCommand` is structured to take the shell

https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/base.py#L415

And then derive the real executable from it

https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/utils/command.py#L42

So my (same) suggestion is to use this same technique to find the correct executable path based on get_shellI(), and use that for the interpreter line (with added `#!` of course). I think this is cleaner than hard coding these paths that 1) might be wrong and 2) might be the wrong binary/shell to begin with.

shahzebsiddiqui:
Sorry i may have misunderstood and it took me some time to figure this out. I am not sure if i did what you were expecting so I was testing this out.

I am running this locally. One thing i dont understand is why we need ``shell: /bin/bash`` and ``shell: /bin/sh`` when we have ``shell: bash`` and ``shell: sh``. Anyhow, if there is some issue I will look at it tomorrow.

```
gitpod /workspace/buildtest-framework $ buildtest build -d -c examples/script/bzip2.yml 
                             start time: 2020-03-20 01:08:45.511038
                     configuration file: /workspace/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home/gitpod/.buildtest/testdir
                               testpath: /home/gitpod/.buildtest/testdir/bzip_compression.sh
                                logpath: None



STAGE                                    VALUE
#!/bin/bash
cd $TESTDIR
FILE=1mb_bzip2.txt
TARFILE=1mb_bzip2.txt.bz2
dd if=/dev/zero of="$FILE" bs=1M count=1
tar cvjf "$TARFILE" "$FILE"
ls -l "$TARFILE"

                             start time: 2020-03-20 01:08:45.512081
                     configuration file: /workspace/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home/gitpod/.buildtest/testdir
                               testpath: /home/gitpod/.buildtest/testdir/bzip_version.sh
                                logpath: None



STAGE                                    VALUE
#!/bin/sh
cd $TESTDIR
bzip2 --version
                             start time: 2020-03-20 01:08:45.512629
                     configuration file: /workspace/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home/gitpod/.buildtest/testdir
                               testpath: /home/gitpod/.buildtest/testdir/bzip_help.sh
                                logpath: None



STAGE                                    VALUE
#!/bin/sh
cd $TESTDIR
bzip2 --help
                             start time: 2020-03-20 01:08:45.513036
                     configuration file: /workspace/buildtest-framework/examples/script/bzip2.yml
                                testdir: /home/gitpod/.buildtest/testdir
                               testpath: /home/gitpod/.buildtest/testdir/bzip2_help.sh
                                logpath: None



STAGE                                    VALUE
#!/bin/bash
cd $TESTDIR
bzip2 -h
```

vsoch:
> One thing i don't understand is why we need shell: /bin/bash and shell: /bin/sh when we have shell: bash and shell: sh

It's giving the user flexibility to use either as the shell parameter in the test configuration. If I write bash and it tells me it's invalid (requiring /bin/bash) this would be annoying. If I write /bin/bash and the same happens, that's also an un-needed annoyance. The reason that bash or sh alone would be preferable is because we can discover the executable on the path with `which`, and for recipes that want to be run with higher levels of security, the full path `/bin/bash` could be provided. Both would work, and both have use cases that should be supported.

buildtest/buildsystem/base.py
```diff
@@ -546,9 +546,9 @@ def _get_test_lines(self):
 
         if self.get_shell() in ["sh", "/bin/sh"]:
             lines += [f"#!/bin/sh"]
-            
+
         elif self.get_shell() in ["bash", "/bin/bash"]:
-            lines += [BUILDTEST_SHELL]
+            lines += [f"#!{BUILDTEST_SHELL}"]
```
vsoch, line 554: I think you would just want to do simply:

```bash
# Get the default shell executable
shell = which(self.get_shell())

# But if it doesn't exist, default to buildtest default
if not shell:
    shell = BUILDTEST_SHELL
lines += [f"#!{shell}"]
```
I don't see why we need all the custom logic for the shell type - we know it's one of bash or sh from the schema, and then it come down to finding an executable on the path. I'd even say that we should quit with an error if it's _not_ found because how would the script be run?

vsoch (changes requested):

shahzebsiddiqui:
> > One thing i don't understand is why we need shell: /bin/bash and shell: /bin/sh when we have shell: bash and shell: sh
> 
> It's giving the user flexibility to use either as the shell parameter in the test configuration. If I write bash and it tells me it's invalid (requiring /bin/bash) this would be annoying. If I write /bin/bash and the same happens, that's also an un-needed annoyance. The reason that bash or sh alone would be preferable is because we can discover the executable on the path with `which`, and for recipes that want to be run with higher levels of security, the full path `/bin/bash` could be provided. Both would work, and both have use cases that should be supported.

I see your reasoning, the thing is you don't have to worry about getting full path to `bash` or `sh` it is understood that these are part of $PATH for all linux distros. I dont see any way user can run into issue where `bash` or `sh` is command not found. 

We don't even need to do ``bash script.sh`` we can just run the script as is ``script.sh`` since the file is executable.

vsoch:
With the way that subprocess is set up, it's best to provide the absolute path to the executable. I've had experience doing this before and this is what I've concluded. We could be dishy and just execute this mysterious script or use a weird shabang, but I think we should aspire to do it correctly.

buildtest/buildsystem/base.py
shahzebsiddiqui: there is no ``which`` method i can use either i use BuildTestCommand or we can leave as is. If we use the class something like:

```
cmd = BuildTestCommand(f"which {self.get_shell}")
cmd.execute()
cmd.get_output()
```
We are expecting `which` is installed in system which is generally not. I'd prefer hard coded path. or if you like a predefined shebangs in defaults.py.

I can look at this tomorrow or if you have a better alternative, feel free to merge into my PR.

vsoch: BuildTestCommand is already using which:

https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/utils/command.py#L42

:)

shutil.which is supported after python 3. So, there is a which method you can use. And it wouldn't make sense to give to the build command, because it already uses it internally. Instead, you'd do as I showed you, a which() directly to the shell, which is what _the BuildTestCommand is already doing_...

vsoch: I already provided the exact code, if you want me to update the PR with it tomorrow I'd be happy to. Cheers!

vsoch:
Let me walk you through how I originally was designing this, because I still think it might be the cleanest design. The schema I think best should _not_ have any limitation on what can be set as an executable (e.g., the update we made for this PR should be undone). For example, I could want my shell to be python because I want to write python code for the run:

```yaml
run-python:
    type: script
    shell: python
    run: |
       import sys
       sys.exit(0)
```
This is fundamentally different from what you suggested above, which requires having this script written in advance:

```yaml
run-python:
    type: script
    shell: bash
    run: |
        python script.py
```
Because in the first, I can have my entire test code written in one file. The second doesn't tell me much about the content of script.py, and it requires me to provide some script.py somewhere (more confusing).

Then when it comes to the functions here, we also don't need to think too hard about the executable, because we find it with shutil.which(). The same would be done for the shabang - use the which to derive the executable.

But now with the changes, we've completely removed this flexibility, both making it harder for the user to write the test, and for us to implement _all the possible cases_ here. So my suggestion is to return the schema to be flexible as it was (no constraint) and only have issue when the executable desired by the user (Rscript, python, julia, bash, whatever!) is not found. The same which() to derive the shabang will work too. Let me know if you have any questions.

buildtest/buildsystem/base.py
shahzebsiddiqui: okay i think got it working as you mentioned. See https://github.com/HPC-buildtest/buildtest-framework/pull/229/commits/f88efa716344e93fd5c6936e9f82c5e2977d5e8f

I tested locally the shebangs were different but its okay some where ``/usr/bin/bash`` and some were ``/bin/bash``. It doesn't matter since recent distros including RHEL7 and Ubuntu have /bin symlinked to /usr/bin

```
gitpod /workspace/buildtest-framework $ ll -l /bi*
lrwxrwxrwx 1 root root 7 Jan 14 09:43 /bin -> usr/bin/
```

```
gitpod /workspace/buildtest-framework $ head -n 1 ~/.buildtest/testdir/{bzip2_help,bzip_compression,bzip_help,bzip_version}.sh 
==> /home/gitpod/.buildtest/testdir/bzip2_help.sh <==
#!/bin/bash

==> /home/gitpod/.buildtest/testdir/bzip_compression.sh <==
#!/usr/bin/bash

==> /home/gitpod/.buildtest/testdir/bzip_help.sh <==
#!/bin/sh

==> /home/gitpod/.buildtest/testdir/bzip_version.sh <==
#!/usr/bin/sh
```

shahzebsiddiqui:
> Let me walk you through how I originally was designing this, because I still think it might be the cleanest design. The schema I think best should _not_ have any limitation on what can be set as an executable (e.g., the update we made for this PR should be undone). For example, I could want my shell to be python because I want to write python code for the run:
> 
> ```yaml
> run-python:
>     type: script
>     shell: python
>     run: |
>        import sys
>        sys.exit(0)
> ```
> 
> This is fundamentally different from what you suggested above, which requires having this script written in advance:
> 
> ```yaml
> run-python:
>     type: script
>     shell: bash
>     run: |
>         python script.py
> ```
> 
> Because in the first, I can have my entire test code written in one file. The second doesn't tell me much about the content of script.py, and it requires me to provide some script.py somewhere (more confusing).
> 
> Then when it comes to the functions here, we also don't need to think too hard about the executable, because we find it with shutil.which(). The same would be done for the shabang - use the which to derive the executable.
> 
> But now with the changes, we've completely removed this flexibility, both making it harder for the user to write the test, and for us to implement _all the possible cases_ here. So my suggestion is to return the schema to be flexible as it was (no constraint) and only have issue when the executable desired by the user (Rscript, python, julia, bash, whatever!) is not found. The same which() to derive the shabang will work too. Let me know if you have any questions.

Okay so i tried doing a python example i dont think its practical due to the way YAML multiline works, it was my first attempt and maybe there is a way around. What i observed is due to python syntax including indentation, and quotes, it causes issues. I think its much easier in this case writing a python script and invoking. I do understand that it doesn't provide value when you just do ``python script.py`` and the file could be on local machine.

So i have an example of area of circle 

```
version: 0.0.1

circle_area:
  type: script
  shell: python
  description: "Calculate circle of area given a radius"
  run: |
import math
radius = 2
area = math.pi * radius * radius
print("Circle Radius ", radius)
print("Area of circle ", area)
```

Running this i ran into issue with variable assignment 

```
itpod /workspace/buildtest-framework $ buildtest build -c examples/script/python.yml 
  File "/workspace/.pip-modules/bin/buildtest", line 11, in <module>
    load_entry_point('buildtest-framework', 'console_scripts', 'buildtest')()
  File "/workspace/buildtest-framework/buildtest/main.py", line 20, in main
    parser.parse_options()
  File "/workspace/buildtest-framework/buildtest/menu/__init__.py", line 81, in parse_options
    args.func(args)
  File "/workspace/buildtest-framework/buildtest/menu/build.py", line 88, in func_build_subcmd
    bc = BuildConfig(config_file)
  File "/workspace/buildtest-framework/buildtest/buildsystem/base.py", line 70, in __init__
    self.load(config_file)
  File "/workspace/buildtest-framework/buildtest/buildsystem/base.py", line 91, in load
    self._validate_global()
  File "/workspace/buildtest-framework/buildtest/buildsystem/base.py", line 148, in _validate_global
    self.recipe = load_recipe(config_file)
  File "/workspace/buildtest-framework/buildtest/buildsystem/schemas/utils.py", line 40, in load_recipe
    content = yaml.load(fd.read(), Loader=yaml.SafeLoader)
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/__init__.py", line 114, in load
    return loader.get_single_data()
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/constructor.py", line 41, in get_single_data
    node = self.get_single_node()
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 133, in compose_mapping_node
    item_value = self.compose_node(node, item_key)
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/parser.py", line 428, in parse_block_mapping_key
    if self.check_token(KeyToken):
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/scanner.py", line 115, in check_token
    while self.need_more_tokens():
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/scanner.py", line 152, in need_more_tokens
    self.stale_possible_simple_keys()
  File "/home/gitpod/.pyenv/versions/3.7.7/lib/python3.7/site-packages/yaml/scanner.py", line 292, in stale_possible_simple_keys
    "could not find expected ':'", self.get_mark())
yaml.scanner.ScannerError: while scanning a simple key
  in "<unicode string>", line 8, column 1:
    import math
    ^
could not find expected ':'
  in "<unicode string>", line 9, column 1:
    radius = 2
    ^
```

When i copy paste this into a python script and run it was working fine. 

```
(docs) login1.stampede2(1190)$ python circle.py
Circle Radius 2
Area of circle 12.566370614359172
```

I think the feature support should be complemented with practical tests, I do intend to write some tests in python but i dont think this would be the right way. Now i have to troubleshoot YAML on top of python.

shahzebsiddiqui:
@vsoch  i think python is out of scope of this PR, we should discuss this in another issue. This PR is addressing several minor issues including #230 #227

vsoch:
It’s relevant because you are removing the ability to freely specify any executable. Do you want me to PR to this PR today to revert back?

shahzebsiddiqui:
Be my guest.

vsoch:
@shahzebsiddiqui please see https://github.com/HPC-buildtest/buildtest-framework/pull/233 which does the changes that I've suggested, and adds the working python example. It includes all of your commits, so merging there will close the PR here. There are issues with how you are saving the files (the paths) for the testdir, but since #232 is addressing this, I'm ignoring for now.

shahzebsiddiqui:
this is now merged in `devel` in PR #233. This PR is not relevant anymore.

PR #231, opened 2020-03-19
Remove clear build option

shahzebsiddiqui:
I don't see any point of having this feature so i removed it for now.


vsoch (approved):
Total agree with this - it doesn't make sense to keep some global massive json of build metadata. I'd even say to go further and remove that, but happy to look at another PR scoped to that.

Issue #158, opened 2020-02-11
check buildtest configuration file for errors


shahzebsiddiqui:
buildtest has significant changes to configuration file, for now there is very few check on configuration file. This issue is prevalent in `devel` branch and buildtest needs to check key/value pair to ensure user does not corrupt buildtest configuration.

vsoch:
Since we are using json schema now, and we have merged #237 into devel, I am going to close the issue.

PR #236, opened 2020-03-23
json schema for buildtest configuration

shahzebsiddiqui:
This PR makes first attempt to do the following:
- Provides a json schema for buildtest configuration (``config_schema.json``)
- buildtest configuration is now changed to json file (``settings.json``)
- Add regtest for schema check
- Remove ``BUILDTEST_VERSION`` from config_opts, it is not needed
- Update ``config_opts["EDITOR"]`` to ``config_opts["editor"]``  that relates to ``buildtest config edit``
- Remove option ``buildtest show --config`` see https://github.com/HPC-buildtest/buildtest-framework/pull/236/commits/8b0f4f303c4a40e7d60c2bcfbc8bd46f77e328ec
- Initial support for validating scheduler and launcher types. If launcher is not found in $PATH, we exit immediately.


buildtest/config_schema.json
```diff
@@ -0,0 +1,56 @@
+{
+    "$id": "https://raw.githubusercontent.com/HPC-buildtest/buildtest-framework/devel/buildtest/config_schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Buildtest Configuration Schema Definition",
+    "type": "object",
+    "defs": {
```
vsoch: Are any properties required (they would go up here).

buildtest/config_schema.json
```diff
@@ -0,0 +1,56 @@
...
+                }
+            }        
+        }
+    },
+    "properties": {
+        "editor": { 
+            "type": "string",
+            "enum": ["vi","vim","nano","emacs"]
+        },
+        "queues": {
```
vsoch: Possibly we should call this executors? Queues limits to job managers, while executors also extends to cloud "queues." For example, snakemake uses executors, and SLURM/SGE/ other standard HPC queues are supported. If you are modeling a queue as explicitly different than an executor, then I'd be curious to know how an executor would be defined. I think a cleaner solution would be to have an upper level executor that could support either. Take a look at snakemake/executor.py to see how it works. Akin to how we have a BuilderBase for script, we probably could have an ExecutorBase that a custom executor extends.

buildtest/config_schema.json
```diff
@@ -0,0 +1,56 @@
...
+                "properties": {
+                    "name": {"$ref": "#/defs/alphanumeric"},
+                    "description": {"type": "string"},
+                    "scheduler": {
+                        "type": "string",
+                        "enum": ["local", "slurm"]
+                    },
+                    "launcher": {
+                        "type": "string",
+                        "enum": ["local", "mpirun", "mpiexec", "sbatch", "srun"]
```
vsoch: This launcher variable I think would be specific to an executor (and not represented in a global config validation per say, but validation for that particular executor).

buildtest/defaults.py
```diff
@@ -31,11 +31,12 @@
 
 # dictionary used for storing status of builds
 BUILDTEST_BUILD_HISTORY = {}
-BUILDTEST_CONFIG_FILE = os.path.join(BUILDTEST_ROOT, "settings.yml")
-BUILDTEST_CONFIG_BACKUP_FILE = os.path.join(BUILDTEST_ROOT, "settings.yml.bak")
+BUILDTEST_CONFIG_FILE = os.path.join(BUILDTEST_ROOT, "settings.json")
```
vsoch, line 38: Why did you change the settings file to be json instead? Why should we not be consistent?

buildtest/defaults.py
```diff
@@ -31,11 +31,12 @@
...
+BUILDTEST_CONFIG_BACKUP_FILE = os.path.join(BUILDTEST_ROOT, "settings.json.bak")
 
 # TESTCONFIG_ROOT is the root directory where test configurations are found
 # configs can be specified as full paths or relative to this path
 TESTCONFIG_ROOT = os.path.join(BUILDTEST_ROOT, "site")
-DEFAULT_CONFIG_FILE = os.path.join(root, "settings.yml")
+DEFAULT_CONFIG_FILE = os.path.join(root, "settings.json")
+DEFAULT_CONFIG_SCHEMA = os.path.join(root, "config_schema.json")
```
vsoch: I don't think we should start storing schemas in the root. I'd suggest a subfolder under buildtest.

buildtest/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+    "editor": "vi",
+    "queues": [
+        {
```
vsoch: This config file answers my question that a "queue" is not necessarily a standard HPC queue, it's serving more as an executor, and should be called as such (and validated with subtypes similar to the Config File schemas).

tests/config_schema_examples/example1.json
```diff
@@ -0,0 +1,23 @@
+{
+    "editor": "vi",
+    "queues": [
+        {
```
vsoch: These should be updated with the executor format, and the files should be named to give more meaningful insight about their content. "Example1" tells me nothing, but `tests/config_examples/local-example.json` or `tests/config_examples/slurm-example.json` is meaningful.

vsoch (changes requested):

tests/config_schema_examples/example1.json
shahzebsiddiqui: I will fix this in next commit

buildtest/defaults.py
shahzebsiddiqui: I found it easier to write configuration in json. It was a personal choice and buildtest setting does not have to be bound to YAML.

buildtest/config_schema.json
shahzebsiddiqui: I purposely named them as `queues` to confirm to the HPC workload managers. Most if not all except Slurm refers to queues as `partition` which is their way of calling queues. See https://slurm.schedmd.com/rosetta.pdf

I'd prefer to stick to `queue` and not introduce `executors`. Currently, cloud is out of scope of buildtest. I foresee buildtest to integrate with job-scheduler so buildtest can automate job submission with default behavior is to run test locally if no scheduler is available. 

The purpose of `queues` is to allow user to specify one or more queue including `localhost` which could be default where jobs can be run. In subsequent PR what I had in mind was to have a key to select one or more queues to run the test.  Something like 

```
version: 0.0.1
hostname:
  type: script
  queues: [ "localhost", "admin" ]
  run: "hostname"
```

Again this example is a bit too simplistic, but first one ``localhost`` which is running test locally, the ``admin`` queue is of type ``scheduler: slurm``  with a ``launcher: sbatch`` which will run same test using ``sbatch``. 

I think there needs to be **keys** inside one of the schemas (script, etc...) which allows tweaking of scheduler configuration when writing test. This requires a bit more thought.


ReFrame does something very similar to site configuration when building test. It's covered in https://reframe-hpc.readthedocs.io/en/latest/configure.html and schema is defined in https://github.com/eth-cscs/reframe/tree/master/schemas

buildtest/config_schema.json
shahzebsiddiqui: Yes so far there are only two required properties `editor` and `queues` see https://github.com/shahzebsiddiqui/buildtest-framework/blob/configuration_jsonschema/buildtest/settings/settings.schema.json#L54

Is it best practice to put `required` above the `object` field. When i was learning the json schema https://json-schema.org/understanding-json-schema/reference/object.html#required-properties i saw the `required` key defined at bottom.

buildtest/defaults.py
shahzebsiddiqui: I put the files under ``buildtest/settings/`` to imply this is for buildtest settings and named the two files ``settings.json`` and ``settings.schema.json``.

buildtest/config_schema.json
vsoch: Fair enough! I would still like to see validation for queue specific types, because it's always going to be the case that you see srun with slurm, for example. Having validation on the level of the queue also ensures that it's easy to add "other queue-like things" (ahem executors) that you might not think are important, but not everyone agrees with you about that.

buildtest/defaults.py
vsoch: I can't argue with that, although it does seem inconsistent.

buildtest/config_schema.json
vsoch: I just missed it, the order is fairly trivial! I've also seen it at the top, as more of a "heads up this is required." However, I'm indifferent to order, bottom is fine too.

buildtest/config_schema.json
shahzebsiddiqui: `executors` could be something we can put on roadmap, but will need some use-case and how it's intended to work. Right now this was first attempt and will require more revisions to schema. This also addresses the fact we didn't have any check on buildtest configuration (``settings.yml``) see #158. 

On a unrelated note, I foresee no point in having ``buildtest show --config`` when ``buildtest config view`` is showing the content in json.

I intended ``buildtest show --config`` to represent data like key/value pair that can be easily interpreted by user instead of reading a json or yaml file. The output style was inspired by ``eb --show-config`` see https://easybuild.readthedocs.io/en/latest/Configuration.html#overview-of-current-configuration-show-config-show-full-config

The output of this mess looks like 

```
gitpod /workspace/buildtest-framework $ buildtest show --config
editor                                             = vim
queues                                             = [{'name': 'localhost', 'description': 'submit job locally', 'scheduler': 'local', 'launcher': 'local'}, {'name': 'admin', 'description': 'Admin Queue', 'scheduler': 'slurm', 'launcher': 'sbatch', 'options': ['#SBATCH -n 1'], 'vars': {'FOO': 'BAR'}}]
``` 

If it makes sense, i can remove this option in this PR as well.

vsoch: Sure thing, removing show config is okay with me!

I would still like to see validation for queue specific types. If we are going to implement this, it makes sense to do it right the first time.

shahzebsiddiqui: ok i will remove ``buildtest show --config``. 

I think i understand the validation that is missing I will get that done. The ``launcher`` need to be compatible with ``scheduler`` type so that user can't say ``launcher: bsub`` and specify ``scheduler: slurm``.

Ill get to this tomorrow hopefully.

vsoch: Awesome! I'm recording a podcast this morning followed by meetings, but I'll be around later today to review. This is pseudo code, but I am thinking something like:

```json
{
    "type": "array",
    "items": {
        "oneOf": [
            {"slurm":  "$ref": "#slurm" }, 
            {"sge": [  "$ref": "#sge" ] }
        ]
    }  
}
```
(Note sure if the ref is correct, but the gist is that you have a reference for an item under definition, and each of these items can be specific to the scheduler!  There are a bunch of examples here https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref. This way, the properties that are decided on for within some scheduler don't limit us for future development of other schedulers, and if an institution really wanted to, they could modify the config to add a custom scheduler (for example TACC has a tool on one of their clusters called the launcher that is specific to TACC).

Gotta run!

buildtest/config.py
```diff
@@ -108,33 +111,43 @@ def load_configuration(config_path=None):
 # Run on init, so we only load once
 config_opts = load_configuration()
 
-supported_launcher_dict = {
-    "local": ["local", "mpirun", "mpiexec"],
-    "slurm": ["mpirun", "mpiexec", "srun"]
-}
 
-def validate_queues(config_queues)
+def validate_queues(queues):
```
vsoch, line 115: These validation functions are defeating the purpose of having the jsonschema package - it's like going back to the previous code that had to manually check all these fields, etc. We want the validation to happen with the jsonschema, not via hard coded if/else statements.

shahzebsiddiqui: i understand, i had some trouble redesigning the schema again, if i were to do it again, I would and by no means this PR is perfect. I didn't want to break my existing work so happy to get second revision

vsoch: I think implementing something complex that would need to be removed is not the right way to address this. If you need help with the schema, then you should ask and I can put it on my queue to take a look. But adding and merging something that will be removed because it isn't what we would consider the best solution? That doesn't make sense, and it's not good use of your time.

shahzebsiddiqui: I dont anticipate to get the schema correct the first time, It was a learning experience so not a complete waste of time. 

What i had in mind was having ability to define one or more local scheduler followed by other workload managers (slurm). The way i was thinking was something along the lines of this.

```
{
    "editor": "vi",
    "queues": 
     {
       "local": 
        {
          "localhost": {                      
            "description": "submit job locally",            
            "launcher": "local"
            "vars:" {
              "FOO": "BAR"
            }
            "modules": {
              "purge": true
              "name": ["GCC"]                
            }
              
          },
          "local_mpirun": {
            "description": "submit job locally via mpirun",
            "launcher": "mpirun",            
          }
        }
        "slurm": {
          "admin": {
              "description": "admin partition",
              "launcher": "srun",
              "options": [
                "#SBATCH -n 1"
                ]
          }
        }
     }
}
```

vsoch: can you explain the choice of localhost? localhost usually refers to 127.0.0.1 which can be confusing in this context.

shahzebsiddiqui: That's probably a bad name convention. The ``localhost``  and ``local_mpirun`` are queues of type ``local`` which means they are meant to run tests locally without a scheduler. Note that grouping queues by `local` allows me to get rid of ``scheduler: local`` and i can now have unique keys that represent local queues.  

Currently, buildtest supports running test locally, the idea was to allow test-configuration reference one or more queues to run job through a scheduler which could be ``local`` or an actual scheduler. I also had in mind a ``ssh`` launcher which would run test using ``ssh`` on ``localhost`` but it could be configured to use another host. This assumes passwordless-ssh and is kinda out-of scope of this PR. The only value i see with ``ssh`` is ability to run job in a clean environment on any node for user that don't have a scheduler. 

There were few others things i had in mind which were not addressed
- It could be nice to allow user to specify path to launcher types, or buildtest can auto-detect  which is what it is doing now
- Currently, i am checking that slurm queue exists with one defined in configuration file, i think that should be changed by allowing a ``partition`` key and the name (``admin`` in example above) could be arbitrary. Reason is a user could want more than queue configuration using one partition without getting a name conflict in json. Hope that makes sense.

We got few options we could do:
1. If you have a proposed buildtest site configuration (Stanford) you would like to use please share your proposal and I  can try to implement the schema and update this PR
2. I can temporarily close the PR, work on my branch until this is complete and reissue PR for review, though we will lose all the reviews so you would have to do it again.
3. We can PR merge as is or with a few minor modification if you have any. 
4. You can branch from my PR and merge into mine, or re-issue PR to upstream. 

Let me know what your thoughts.

vsoch: Let me take a shot sometime this week at playing around with the schema, I think I understand what you are potentially trying to do. I definitely don’t want to lose all the work you’ve done thus far for this PR, it’s good and we just need this final detail.

shahzebsiddiqui: Sounds good, in that case i can keep my branch and PR open for time being. I kinda like your proposal of introducing json to validate schema using **jsonschema**, makes life easy in terms of schema validation and allow for more elaborate configurations. I am planning on start developing a site configuration for stampede2 with some tests in a repo, i think practically using buildtest at a site will help build and refine the features. It's bit hard to do this at my local cluster.

vsoch:
okay I'm starting work on this today. No time ETA at the moment, but please hold off on additional changes.

shahzebsiddiqui:
> okay I'm starting work on this today. No time ETA at the moment, but please hold off on additional changes.

No worries, i wont be making any changes going forward.

buildtest/defaults.py
```diff
@@ -31,11 +35,16 @@
...
 # TESTCONFIG_ROOT is the root directory where test configurations are found
 # configs can be specified as full paths or relative to this path
 TESTCONFIG_ROOT = os.path.join(BUILDTEST_ROOT, "site")
-DEFAULT_CONFIG_FILE = os.path.join(root, "settings.yml")
+DEFAULT_CONFIG_FILE = os.path.join(root, "settings", "settings.json")
+DEFAULT_CONFIG_SCHEMA = os.path.join(root, "settings", "settings.schema.json")
 EDITOR_LIST = ["vim", "emacs", "nano"]
+supported_launcher_dict = {
+    "local": ["local", "mpirun", "mpiexec"],
+    "slurm": ["mpirun", "mpiexec", "srun"],
```
vsoch, line 49: @shahzebsiddiqui these configs aside, could you please provide me examples of recipes that would be run in the context of some scheduler? For example, I should be able to take the same "script" schema recipe and then hand it to one of two executors - local or slurm - and my choice should not impact the test config file. What does that look like? E.g.,:

```
buildtest build -c my-script.yml -e slurm 
```
and the default would be local
```
buildtest build -c my-script.yml
```
What is confusing to me is that if a user wants some particular executable (e.g., mpirun) wouldn't they just write this into their script (or other type of test config)? Why does this need to be represented on the level of buildtest? For something like srun this makes sense because we need to launch a job (and then run something from it) but for the rest it isn't clear that this is the right level of representation.

vsoch: It might be helpful if you can point me to a configuration file for each of the examples:

 - a config that defaults to just local
 - a config that expects slurm
 - a config that defaults to using mpi (something)

So I can better understand this modeling. Thank you!

shahzebsiddiqui: that dictionary was intended to check supported launcher type ``launcher: srun`` in configuration with ``scheduler: slurm`` but depending if you change the json format this section won't be needed.

In regards to the test-configuration i foresee ability for test to be run in one or more queues if possible. 

Let's say the test https://github.com/HPC-buildtest/buildtest-framework/blob/devel/examples/script/zlib.yml could have a key to specify one or more queues to run the test. Let's assume a default queue `local` is provided by buildtest we could have test written something like 

``` 
version: 0.0.1

zlib_compression:
  type: script
  queues: ["local"]
  run: |
    dd if=/dev/zero of=1mb.txt bs=1M count=1
    tar cvzf 1mb.tar.gz 1mb.txt
    ls -lh 1mb.tar.gz
    gzip -l 1mb.tar.gz
```

Most general scripts that are shell script for sanity check on system won't need to be run locally. I think some useful examples will come as i start testing,

For instance in stampede2 we have the following queues 

```
login4.stampede2(1001)$ sinfo -o '%R' -h
development
normal
large
long
flat-quadrant
skx-normal
skx-large
skx-dev
```

I may want to run let's say stream test https://github.com/HPC-buildtest/buildtest-framework/blob/devel/examples/benchmarks/stream.yml on one or more queue 
something like. Note the names ``normal`` and ``sky-normal`` would be queue names in buildtest configuration.

```
version: 0.0.1

stream_uniprocess_c:
  type: script
  queues: ["normal", "sky-normal"]
  description: "STREAM Microbenchmark C test on uniprocessor"
  run: |
    wget https://www.cs.virginia.edu/stream/FTP/Code/stream.c
    gcc -O2 stream.c -o stream
    ./stream
```

I think some sites may define a queue for ``gpu`` workload, it makes sense to have GPU centric tests be assigned to those queues or buildtest provides a queue that is configured to setup a GPU environment.

I may want a queue configuration for GPU workload to define let's say a gpu_volta queue to select volta gpus with 5G and automatically do ``module load CUDA/10.0``.  Note that queue is "gpu_volta" but actual partition is called **gpu**. It may allow user to define multiple queues for same partition. 
 
```
 "slurm": {
          "gpu_volta": {
              "description": "gpu queue for volta",
              "launcher": "srun",
              "partition": "gpu",
              "options": [
                "#SBATCH -n 1",  
                "#SBATCH --gres=gpu:volta:1",
                "#SBATCH --mem-per-gpu 5G"
                ]
             "modules": 
                purge: True
                names: ["CUDA/10.0"]
          }
```

Then if i have some gpu test i could just reference the queue to prepoluate the information, and allow some configuration to be overridden in test-configuration.

Let's say a test for saxpy as follows:

```
version: 0.0.1
saxpy:
  type: script
  queues: ["gpu_volta"]
  run: |
  nvcc saxpy.cu -o saxpy
  ./saxpy
```

I can foresee the test have content 

```
#!/bin/bash
#SBATCH -n 1
#SBATCH -p gpu
#SBATCH --gres=gpu:volta:1
#SBATCH --mem-per-gpu 5G
module load CUDA/10.0
nvcc saxpy.cy -o saxpy
./saxpy
```

These are just some ideas and it may take some time to flesh out all the details. I can see myself have to work with site-admins to run benchmark during maintenance period, so a suite of benchmark that are configured to use a high priority queue such as ``admin`` could be configured on the site and all tests are configured with that tag.

vsoch: okay, so you validated one of my concerns - the build recipes should be robust to run _across environments_ without requiring a hard coded executor. So this example:

```
version: 0.0.1

zlib_compression:
  type: script
  queues: ["gpu_queue"]
  run: |
    dd if=/dev/zero of=1mb.txt bs=1M count=1
    tar cvzf 1mb.tar.gz 1mb.txt
    ls -lh 1mb.tar.gz
    gzip -l 1mb.tar.gz
```
I would need to delete or edit the "queues" key to use for myself, given that I don't want whatever is defined as the user "gpu_queue." Instead, the recipe file should be agnostic to the executor:

```
version: 0.0.1

zlib_compression:
  type: script
  run: |
    dd if=/dev/zero of=1mb.txt bs=1M count=1
    tar cvzf 1mb.tar.gz 1mb.txt
    ls -lh 1mb.tar.gz
    gzip -l 1mb.tar.gz
```
But then specified to run under some environment by the user, or by the default configuration file enforced for the user. For example:

```bash
buildtest build -c zlib.yml -e gpu_local
```
This makes the recipes portable, which is essential if you want people to share.

shahzebsiddiqui: I foresee it would be difficult to have fully portable tests when it comes to writing elaborate tests. One use-case in HPC is around having buildtest to automate acceptance testing for each site. Certainly this requires site-specific test which can't be shared It was discussed in SC19 https://olcf.github.io/hpc-system-test-wg/events/sc19bof.html . It would make sense to have a repository that be used by sites to share tests they run for acceptance test for their sites. 

It gives HPC a common framework to write test, even if tests are site specific but allows anyone to understand the tests and adapt them to their site.

Some basic acceptance test could be simple, but not entirely portable 
- Check login nodes are reachable
- Check slurmctl and slurmdb
- Check filesystem mounts
- Run a series of dummy slurm jobs for sanity (batch, job-array, multi-nodes, gpu jobs)

vsoch: Yes I agree, but hard coding a test specification (executor) in a test recipe is a different thing. If anything needs to be hard coded, it should be in the site-specific (customized) configuration and the test would still be agnostic to that.

PR #237, opened 2020-03-25
Add config jsonschema

vsoch:
This will add the start of work for the executor base classes, which are connected to the different kinds of executors that can be defined in the user's default config schema, OR specified by a user on the command line (not developed yet). 

Buildtest needs to be flexible to allow specification of more than one kind of executor, whether that be local, slurm, or something else we are not considering yet (e.g., cloud executor). For this purpose, I am adding a BuildExecutor class that is able to read in one of a number of build executor types, which currently include a LocalExecutor and SlurmExecutor. The variables and customizations should live on the level of the schemas, read in by the classes, but not need to be validated by them. I do not have a good understanding of how the executor interacts with the builder step, and I will need this understanding to implement the class into the library.


vsoch:
@shahzebsiddiqui I've done basic work to create a base configuration, validation files and examples, and then the BuildExecutor, BaseExecutor, and example LocalExecutor and SlurmExecutor classes. It is very flexible to allow the user to name executors whatever they like, but then define a `type` that is matched to a particular sublass of `BuildExecutor` under buildtest/executors. Each can have any custom set of variables that are needed for it (e.g., you would want a launcher for slurm, maybe that defaults to sbatch but the user could define a one off change either via a configuration:

```yaml

editor: vi
executors:
  slurm:
    type: slurm
    vars:
      launcher: srun
```

or (not developed yet, but will be!) via a command line run to select a specific executor:

```yaml
buildtest build -c slurm.yml -e slurm
```
And you can imagine having a flag to customize some variable, etc. 

To finish these classes, I need to understand how you imagine buildtest executors to interact with the BuildTestConfig. For example, right now we load the builders and run tests:

```
        # And builders parsed through for each
        for builder in bc.get_builders(testdir=args.testdir):

            # Keep track of total number of tests run
            total_tests += 1
            if not args.dry:
                result = builder.run()

                # Update results
                if result["RETURN_CODE"] == 0:
                    passed_tests += 1
                else:
                    failed_tests += 1
            else:
                result = builder.dry_run()
```
And I intuit that those same tests would need to be run _in context of an executor_. Could you show me some pseudo code for what you had in mind? I'm also done working for today, so feel free to leave notes here so I can finish up this development work.

vsoch:
I haven't put a lot of thought into this, but my general thinking is that we will have loaded the default executor (and any modifications to it) by the time we get to running tests, and there should be a load of the `BuildExecutor()` right above that loop. And then perhaps we would get the executor defined for each recipe (or use the default for the config) and then run some basic steps that you would expect, something like?

```
be = BuildExecutor(config_opts)
# for each builder
executor = be.get(name)  # or default
executor.setup()
executor.up()
# run build
executor.down()
executor.cleanup()
```
The verbs match what you would expect for a service architecture, but of course could be mapped to different environments. For example, for an ssh executor, the setup would likely include testing / making a connection, and then preparing some wrapper to allow for connecting to the remote. It might even be the case that we provide the builder to the executor to manage the run. Thoughts?

.github/workflows/urlchecker.yml
```diff
@@ -24,4 +24,4 @@ jobs:
         retry_count: 3
 
         # Cannot check private GitHub settings
-        white_listed_patterns: https://github.com/HPC-buildtest/buildtest-framework/settings
+        white_listed_patterns: https://github.com/HPC-buildtest/buildtest-framework/settings,https://mlbot.net
```
shahzebsiddiqui: i'd would not add the ``https://mlbot.net`` i was going to remove this from docs

buildtest/config.py
```diff
@@ -123,27 +89,9 @@ def load_configuration(config_path=None):
...
-            tree_list = []
-
-            # check each directory in MODULEPATH and add it to BUILDTEST_MODULEPATH
-            for tree in os.getenv("MODULEPATH", "").split(":"):
-                if os.path.isdir(tree):
-                    tree_list.append(tree)
+    config_opts = load_schema(config_path)
 
-            config_opts["BUILDTEST_MODULEPATH"] = tree_list
+    # config_opts["BUILDTEST_VERSION"] = BUILDTEST_VERSION
```
shahzebsiddiqui: i would be fine with removing line 94, ``BUILDTEST_VERSION`` is now defined in ``buildtest/__init__.py`` https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/__init__.py#L1 which can be imported as variable.

buildtest/defaults.py
```diff
@@ -37,5 +37,6 @@
 # TESTCONFIG_ROOT is the root directory where test configurations are found
 # configs can be specified as full paths or relative to this path
 TESTCONFIG_ROOT = os.path.join(BUILDTEST_ROOT, "site")
-DEFAULT_CONFIG_FILE = os.path.join(root, "settings.yml")
+DEFAULT_CONFIG_FILE = os.path.join(root, "settings", "default.yml")
+DEFAULT_CONFIG_SCHEMA = os.path.join(root, "settings", "default-schema.json")
```
shahzebsiddiqui: This just a comment, maybe we should be consistent with all schema files be something like **file.schema.json**. Thats how the other schema are.

buildtest/executors/base.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class BuildExecutor:
+    """A BuildExector is a base class some type of executor, defined under
+       the buildtest/settings/default-config.json schema. For example,
+       the types "local" and "slurm" would map to `LocalExecutor` and
+       `SlurmExecutor` here, each expecting a particular set of
+       variables under the config options. If options are required
+       and not provided, we exit on error. If they are optional and not
+       provided, we use reasonable defaults.
```
shahzebsiddiqui, line 20: It would be nice to have a list of public and private methods for each class in a docstring. 

According to https://www.python.org/dev/peps/pep-0257/

**The docstring for a class should summarize its behavior and list the public methods and instance variables**

I tried my initial attempt to do this something like https://github.com/HPC-buildtest/lmodule/blob/master/lmod/module.py#L30

buildtest/executors/base.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    def __init__(self, config_opts):
+        """initiate executors, meaning that we provide the config_opts
```
shahzebsiddiqui, line 24: Can we capitalize first letter in the docstring and finish last sentence with period.

buildtest/executors/base.py
```diff
@@ -0,0 +1,102 @@
...
+           that are validated, and can instantiate each executor to be available
+           
+           Parameters:
+
+           config_opts: the validated config opts provided by buildtest.
```
shahzebsiddiqui, line 29: Can we make use of ``param`` and ``type``. See https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html for list of docstring format supported by Sphinx. You can see how i did this for lmodule at https://github.com/HPC-buildtest/lmodule/blob/master/lmod/module.py#L51 

It makes for some nice API docs

buildtest/executors/base.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get(self, name):
+        """Given the name of an executor return the executor for running 
+           a buildtest build.
+        """
+        return self.executors.get(name)
+
+
+class BaseExecutor:
+    """The BaseExecutor is an abstract base class for all executors
```
shahzebsiddiqui, line 54: Can we add public methods in docstring and end last sentence with a period

buildtest/executors/base.py
```diff
@@ -0,0 +1,102 @@
...
+class SlurmExecutor(BaseExecutor):
+    type = "slurm"
+
+    def load(self, name):
+        """Load the executor preferences from the provided config, which is
+           added and indexed with "name." For slurm we look for the following
+           in vars:
+
+           launcher: defaults to sbatch
+        """
```
shahzebsiddiqui, line 103: Can we add newline after end of each docstring.

buildtest/settings/README.md
```diff
@@ -0,0 +1,47 @@
+# Settings
```
shahzebsiddiqui, line 1: Thanks for documenting this for now its okay, we would eventually want this under buildtest docs under Configuring Buildtest https://buildtest.readthedocs.io/en/devel/configuring_buildtest.html

buildtest/settings/default-schema.json
```diff
@@ -0,0 +1,80 @@
...
+               },
+               "options": {
+                   "type": "array",
+                   "items": {
+                       "type": "string"
+                   }
+               },
+               "vars": {"$ref": "#/definitions/vars"} 
+           },
+          "required": ["type", "description"]
```
shahzebsiddiqui: i don't think **description** should be required field, it's just placeholder text field and not used for any purpose

buildtest/settings/default.yml
```diff
@@ -0,0 +1,11 @@
+
+editor: vi
+executors:
+  local:
+    type: local
+    description: submit jobs locally
+  slurm:
+    type: slurm
+    description: submit jobs with slurm using srun instead of sbatch
+    vars:
```
shahzebsiddiqui, line 10: i think of ``vars`` where you can declare variables  that get passed in to all test that use the executor.
The launcher should be its own key part of slurm and local that should be of type `string` with `enum` to validate launcher type. Initially for slurm we can use `srun`, `mpirun`, `mpiexec`. For local it could be `mpirun`, `mpiexec`, `local`. The way i intended launcher to be used  is to figure out how job should be dispatched.  

The `local` executor should target atleast `mpirun` so that a user who wants to run MPI code in a system without scheduler can run it.

shahzebsiddiqui (changes requested):

shahzebsiddiqui:
@vsoch i think the way i had in mind for having the configuration is something like this


image

The method allows for multiple definition of local and slurm executors. Note we dont need ``type: slurm`` since all slurm executors are grouped in ``slurm`` key. Every slurm executor is defined as a key and it is associated to a partition. That would allow multiple slurm executor to be defined to one or more partitions. 

There could be more work to be done to allow ``modules`` and ``options`` which predefine a set of options passed to each executor. That could be done at a later stage.

shahzebsiddiqui:
> I haven't put a lot of thought into this, but my general thinking is that we will have loaded the default executor (and any modifications to it) by the time we get to running tests, and there should be a load of the `BuildExecutor()` right above that loop. And then perhaps we would get the executor defined for each recipe (or use the default for the config) and then run some basic steps that you would expect, something like?
> 
> ```
> be = BuildExecutor(config_opts)
> # for each builder
> executor = be.get(name)  # or default
> executor.setup()
> executor.up()
> # run build
> executor.down()
> executor.cleanup()
> ```
> 
> The verbs match what you would expect for a service architecture, but of course could be mapped to different environments. For example, for an ssh executor, the setup would likely include testing / making a connection, and then preparing some wrapper to allow for connecting to the remote. It might even be the case that we provide the builder to the executor to manage the run. Thoughts?

I like this idea, it makes sense to me and defines a proper control sequence of events run by each executor.

For slurm i could see the stages such as 
- setup: buildtest will write slurm job scripts
- check: check if slurm partition is available for accepting jobs, if not skip job 
- dispatch:  dispatch jobs to scheduler
- poll: wait for jobs to finish, periodically run ``squeue`` or ``sacct`` to see if job is in one of its end state see https://slurm.schedmd.com/squeue.html#lbAG
- gather: gather all job data, exit code, output and error to print to stdout and store data for post-processing (convert output to json)
- close: clean up any generated files

.github/workflows/urlchecker.yml
vsoch: okay, if you want me to address this here, I'll also remove from the docs so the urls don't need to be whitelisted.

buildtest/config.py
vsoch: Gotcha. It was commented out, so I wasn't sure.

buildtest/defaults.py
vsoch: Good thinking.

buildtest/executors/base.py
vsoch: Docstrings is out of scope here, there is too much other work to focus on it. I can do my best to add summary, parameters, etc., but nit picking about this here is out of scope.

buildtest/executors/base.py
vsoch: Sure. But I think this check needs to be automated.

.github/workflows/urlchecker.yml
shahzebsiddiqui: Yea might as well and get that out of way. I already closed #238 but didnt clean docs.

buildtest/executors/base.py
vsoch: Sure I can change that.

buildtest/executors/base.py
vsoch: Please add an automated check for this, so the CI tells me.

buildtest/executors/base.py
vsoch: Automated check.

buildtest/settings/README.md
vsoch: Sure, that works. Right now for development I'm putting it where the developer might expect to find it.

buildtest/settings/default-schema.json
vsoch: I think this was required on your PR, hence why it's here? Glad to change it.

buildtest/settings/default.yml
vsoch: The launcher is a slurm specific variable, so the user can define it under variables and it will be parsed given that `type` is slurm. I'm doing a simple implementation now so I'm not adding any of the mpi tools, but they could be other options for that variable as well. If there is some derivative of local that uses MPI, that's also another customization / variable (not a default).

vsoch:
@shahzebsiddiqui the current approach is flat, and allows for the same (multiple specification of various executors for any base, slurm vs local) just by way of the `type` variable. E.g.,

```yaml
editor: vi
executors:
  my-local-1:
    type: local
  my-local-2:
    type: local
    description: submit jobs locally but with another launcher
    vars:
      launcher: mpirun
  my-slurm-1:
    type: slurm
    vars:
      launcher: srun
  my-slurm-2:
    type: slurm
    description: submit jobs with slurm using sbatch
```
It's a simpler structure that allows for the same specifications without the unnecessary nesting.

buildtest/executors/base.py
shahzebsiddiqui: I understand, just try your best. I

buildtest/executors/base.py
shahzebsiddiqui: do you have an automated check in mind, i would like to have something like this.

vsoch: The only place I've seen this run is for urlstechie, and possibly apache airflow. But I've never used this integration so I can't comment on the kind of setup, etc. required. https://app.codacy.com/gh/urlstechie/urlchecker-action/commits?bid=16994166

shahzebsiddiqui (approved):
Looks good to me

shahzebsiddiqui:
> @shahzebsiddiqui the current approach is flat, and allows for the same (multiple specification of various executors for any base, slurm vs local) just by way of the `type` variable. E.g.,
> 
> ```yaml
> editor: vi
> executors:
>   my-local-1:
>     type: local
>     description: submit jobs locally
>   my-local-2:
>     type: local
>     description: submit jobs locally but with another launcher
>     vars:
>       launcher: mpirun
>   my-slurm-1:
>     type: slurm
>     description: submit jobs with slurm using srun instead of sbatch
>     vars:
>       launcher: srun
>   my-slurm-2:
>     type: slurm
>     description: submit jobs with slurm using sbatch
> ```
> 
> It's a simpler structure that allows for the same specifications without the unnecessary nesting.

Okay that makes sense with this example, in previous example one that is provided ``default.yml`` https://github.com/researchapps/buildtest-framework/blob/add_config_jsonschema/buildtest/settings/default.yml seem to be bit misleading. 

Anyhow, I think we can take this as a starting point and make iterative adjustments. I'll use the template as starting point for making a site configuration for stampede2.

vsoch:
@shahzebsiddiqui the default.yml is purposefully simple because it's the tabula rasa, so to speak - we can't assume the user has anything other than a host (no slurm, etc.) What we will want to do is add other config examples that show adding each executor, so a user can just copy paste the entire file or sections that they like.

The additional changes to integrate the stages - should we do that in a separate PR (and merge here so I don't block you?) I like your suggested stages, btw, they make a lot of sense.

shahzebsiddiqui:
@vsoch this looks fine, i realize the PR was not rebased from my branch. it's okay, there were few changes that were not added from my PR that could be addressed in subsequent changes. We can be good with merging

The changes i had in mind which were missed
- Remove `buildtest show --config`
- Add details to system class variable in BuildTestSystem https://github.com/HPC-buildtest/buildtest-framework/pull/236/files#diff-da5c353202a52de72277768aebc926b6

shahzebsiddiqui:
Feel free to Merge, g2g for lunch. Thanks @vsoch great work 👍

vsoch:
@shahzebsiddiqui I cherry picked the relevant commits (see [here for a rundown](https://stackoverflow.com/questions/9339429/what-does-cherry-picking-a-commit-with-git-mean) which is something that you can do when you want to grab commits/work, but not necessarily the entire branch. In my case I wanted the minimum required to work with, but not the extra ones). But if you look at commits at the top you'll see that your work is represented here (see the commit history with your icon). I'd be happy to open some quick PRs to also integrate these changes that I didn't cherry pick.

PR #242, opened 2020-03-27
Updating Documentation

shahzebsiddiqui:
This PR updates a few things:
- We need to rewrite Configuring Buildtest, for now removing everything 
- Add GitHub bots in GitHub Integration page and minor format changes to page 
- Add buildtest publication in References
- Remove ``buildtest show --config`` from docs


docs/references.rst
```diff
@@ -36,7 +36,11 @@ Conference
   
   - Slides: [`pdf <https://github.com/HPC-buildtest/buildtest-framework/blob/devel/docs/slides/buildtest_hpckp17.pdf>`_]
   
- 
+Publications
+--------------
+
+- Siddiqui S. (2020) `Buildtest: A Software Testing Framework with Module Operations for HPC Systems <https://doi.org/10.1007/978-3-030-44728-1_1>`_ . In: Juckeland G., Chandrasekaran S. (eds) Tools and Techniques for High Performance Computing. HUST 2019, SE-HER 2019, WIHPC 2019. Communications in Computer and Information Science, vol 1190. Springer, Cham
```
vsoch, line 42: You should add some kind of hash here to refer to the current master. I would suggest connecting the repository to zenodo https://zenodo.org/ and then triggering a release, and adding the badge to master here. Then when devel gets merged, the next release will be linked to that version of the code. Importantly. zenodo also makes it easy to link everything up with orcid, etc., because you can provide a .zenodo.json (here is an example https://github.com/datalad/datalad/blob/master/.zenodo.json). I'll open an issue for this.

vsoch (approved):

PR #249, opened 2020-03-28
Add launcher key in slurm object in buildtest configuration

### LINKED ISSUE #246: Print useful information when buildtest configuration fails
If there is a configuration issue, buildtest just prints the following: 
```
(docs) ssi29@ag-mxg-hulk090> buildtest build -c examples/script/bzip2.yml -d
Buildtest Configuration Check Failed!
Configuration File: /u/users/ssi29/.buildtest/settings.yml failed to validate against schema: /mxg-hpc/users/ssi29/buildtest-framework/buildtest/settings/default.schema.json
```

This is not useful in troubleshooting and turns out this can be fixed when invoking the ``validate`` method. This is addressed in #249

### LINKED ISSUE #251: Method for selecting default executor
So here are few ideas on how we can select a default executor as per configuration.

Option 1: Add key ``default`` in executor class
--------------------------------------------------

It would be something like this

```
editor: vi
executors:
  default: runlocal
  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  runlocal:
    type: local
```

We would need to check value of ``default`` after the validate step against all executor objects since we won't know the executor names after the validate step. Json can't use ``enum`` in this case since the executor object name is not defined yet

Option 2: First executor object is default
----------------------------------------------
 
In this case, we don't need to add a separate key (``default``) to make the change. it's just a matter of order. In both case, the default executor does not have to be of ``type: local``. This is true, because we have not made strict check to ensure we have a local executor, if we wanted to we would have to change the json schema.

Option 3: Default per type
-----------------------------

Not sure if this would even make sense but having a default per executor type (``local``, ``slurm``). Here would be an example of two ``local`` and ``slurm`` executor definition with a default executor for each type

```
editor: vi
executors:
  default: 
    local: runlocal
    slurm: normal

  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  admin:
    type: slurm
    description: run jobs on admin partition
    launcher: sbatch
    options:
      - "-p admin"
  runlocal:
    type: local
  
  run_root:
    type: local
    description: run jobs that require root priviledges (root)
```

The behavior of buildtest can change as user change the default executor, it going to be important that their test are compatible when they make change to default executor.

shahzebsiddiqui:
update default.yml
Add slurm-example.yml as reference configuration for sites that using
Slurm
Add executor key in script schema, this refers to name of executor in
configuration to used when dispatching job. This is a required key
Update an example script (examples/script/slurm.yml) to use executor key
Removed the try/except clause in buildtest configuration check, this was
done because validate method prints useful output for troubleshooting
schema, the error message we print to screen was not showing anything
useful.


buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json
```diff
@@ -3,7 +3,7 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "BuildTest Test Configuration for Script",
   "type": "object",
-  "required": ["type", "run"],
+  "required": ["type", "executor","run"],
```
vsoch: This is breaking portability. A test recipe should not have a hard coded executor, it’s determined by the runners config or command line specification at runtime. Also, changes to the schema don’t go here, they are reviewed first in the schemas repository.

shahzebsiddiqui:
@vsoch i have not made any changes to the executor class, this PR is focusing on getting the schema ready on the configuration and script schema so that buildtest can understand how to dispatch the job. The slurm dispatch by default can be ``sbatch`` but it could be others if needed hence ``launcher`` was introduced. The ``executor`` key in ``script`` is needed to specify which launcher to use. Hope that makes sense. Right now a test can pick only one executor.

buildtest/settings/slurm-example.yml
```diff
@@ -0,0 +1,14 @@
+editor: vi
+executors:
+  normal:
+    type: slurm
+    description: submit jobs with slurm using srun instead of sbatch
+    launcher: sbatch
```
vsoch, line 6: Launcher needs to be under vars as it is an executor specific variable.

vsoch (changes requested):

buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json
shahzebsiddiqui: i would prefer having ``executor`` in recipe otherwise it will be problematic especially when it comes to doing directory build (`buildtest build -c dir``) where more than one yml file is processed that may use two different executor. For example, there is one executor for GPU partition, and another for MPI. I have no way of doing this via cli other than running two separate builds. 

Sure, i will make change to same change to schema repo as well.

buildtest/settings/slurm-example.yml
shahzebsiddiqui: ``vars`` is an array that can have any key/value pair. The ``launcher`` needs to be of enum restricted to few types.  I may misunderstood, the ``vars`` are for defining shell variables, if its for executor specific vars, how do you intend on using it.

buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json
vsoch: This would be akin to hardcoding an executor into a Snakefile, meaning that if I move my Snakemake to some other environment I can't just run it as is. The entire purpose of having the user define executors for their host and then specify to use them either with default config or a flag is to make these recipes portable. If executor is added, it should be an *optional* value and then the client here needs to have a flag to ignore it so the recipes are still portable.

shahzebsiddiqui: I understand, in that case i would be fine with making executor as an optional argument and have default executor be one of the local executors. I can make that change

vsoch: The default executor should be whatever the use sets it to be in their local configuration (which I agree would be a local executor when first installed).

shahzebsiddiqui: The change has been made in https://github.com/HPC-buildtest/buildtest-framework/pull/249/commits/a5ab2e7d1f47c99fa936c646c4051f4280964b71 so executor is no longer required. I fixed the Travis issue so it should be all green now.

buildtest/settings/slurm-example.yml
vsoch: Okay that makes sense, but you'll also need to update the executor class [here](https://github.com/HPC-buildtest/buildtest-framework/blob/a5ab2e7d1f47c99fa936c646c4051f4280964b71/buildtest/executors/base.py#L105) that is expecting launcher under vars. I'm not sure if your recipes will validate, but if you report that they do, it definitely is easier to have the slurm type validate for launcher. Perhaps there should be a default for sbatch if it's not defined, and then don't require the user to specify it in the config?

And in this case of defining specific variables, it's not clear to me what purpose "vars" and "options" serve?

vsoch: If "vars" refers to shell variables then they would better be called environment or something like that. Vars is way too general to be clear to the user what it's referring to (I still don't get it).

shahzebsiddiqui: How about let's remove ``vars`` it may be confusing both of us and I dont have a clear idea how it will be used yet

shahzebsiddiqui: the ``options`` is a list of ``SBATCH`` directives that can get injected into the test script. And there can be way to override some of the options in the test configuration. buildtest should add ``#SBATCH`` in front of each item, so that it is used as intended, user shouldn't put any arbitrary value in here.

vsoch: Removing vars sounds good to me! Environment is fairly straight forward, but we have it represented in the [test config](https://github.com/HPC-buildtest/schemas/blob/master/script/script-v0.0.1.schema.json#L27). We'll need to think more about this setup - for example, hard coding environment variables into the testing file isn't the best idea if you want to put them on GitHub. It might also be the case that the runner would want to specify a file name instead, or have a config environment section to override the local one...

Anyway I agree we don't have a good sense of these things, so let's remove vars for now, both for the default config, and the get_variables function.  So you probably want something like:

```python
    def load(self, name):
        """Load the executor preferences from the provided config, which is
           added and indexed with "name." For slurm we look for the following
           in vars:
           :param launcher: defaults to sbatch
           :type launcher: string
        """
        self.launcher = self._settings.get("launcher", "sbatch")
```

shahzebsiddiqui: Let's see if Travis passes all regtest. It worked locally. I made all the changes you requested.

shahzebsiddiqui:
All checks are green :)

tests/config_schema_examples/slurm-example.yml
```diff
@@ -1,8 +1,13 @@
...
 executors:
-  slurm:
+  normal:
     type: slurm
     description: submit jobs with slurm using srun instead of sbatch
+    launcher: sbatch
     vars:
-      launcher: srun
+      FOO: BAR
+    options:
```
vsoch, line 7: Where are options parsed, and why are there vars here if we removed them?

vsoch: If vars was in the example, why did tests still pass before?

shahzebsiddiqui: it's bit odd since ``additionalProperties: False`` in slurm object https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_schema/buildtest/settings/default.schema.json#L48 i have no idea

shahzebsiddiqui: I didn't parse ``options`` yet, that was something i plan to do in later PR. The #SBATCH directives would be read from the options key and injected into the test, so this would require testing on my end at stampede2.

vsoch: I think it's an issue with the nested ref, this looks like something that might work! https://stackoverflow.com/questions/58170135/how-to-use-additionalproperties-from-a-ref-file/58171051#58171051

shahzebsiddiqui: I looked at the link, it's bit confusing according to the docs https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.5.6 

This looks like a silent bug that doesn't show up during testing, and could be something to address in a separate PR. If i understand, we dont want any extra properties passed to any object   whether it be a child object like executor or the parent objects.

vsoch: Basically, the properties serve as booleans to reference if a definition is present or not (and according to the author) this fixes the bug that nesting of schemas doesn't honor the attribute. It's worth a try.

shahzebsiddiqui: so should i remove the following ``additionalProperties`` from ``local`` and ``slurm``. 
https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_schema/buildtest/settings/default.schema.json#L48
https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_schema/buildtest/settings/default.schema.json#L29

I guess the metric for success would be to get the validate method fail when ``vars`` is specified in executor object. Is that right?

vsoch: Yes we want the additionalProperties: false to be honored - in the case of vars it should have triggered an invalid error, but it didn't. It will be important beyond vars, because it could be that a user has a spelling mistake (or something like that) that should be known.

shahzebsiddiqui: Let me do some testing and see what i can fix.

shahzebsiddiqui: so i did a few tests

Test on outer level
---------------------
Expected Result: Fail on ``x: 1``
```
x: 1
editor: vi
executors:
  runlocal:
    type: local
```

```
On instance:
    {'editor': 'vi',
     'executors': {'runlocal': {'description': 'submit jobs locally',
                                'type': 'local'},
                   'x': {'type': 'local'}},
     'x': 1}
```
Test inside executor
-----------------------

Expected Result: Fail on ``vars: 1`` not an object 

```
editor: vi
executors:
  runlocal:
    type: local
  x:
    type: local
  vars: 1
```

```
Failed validating 'type' in schema[0]:
    {'additionalProperties': False,
     'properties': {'description': {'type': 'string'},
                    'options': {'items': {'type': 'string'},
                                'type': 'array'},
                    'type': {'enum': ['local'], 'type': 'string'}},
     'required': ['type'],
     'type': 'object'}

On instance:
    1
```

Test inside executor object
------------------------------

Expected Result: Fail on ``x: 1``

```
editor: vi
executors:
  runlocal:
    type: local
    x: 1
```

What i dont understand is why validate is thinking it needs to match with ``slurm`` object when a new key is being added. Maybe it resorts to the last object since it is using ``anyOf``

```
jsonschema.exceptions.ValidationError: 'local' is not one of ['slurm']

Failed validating 'enum' in schema[1]['properties']['type']:
    {'enum': ['slurm'], 'type': 'string'}

On instance['type']:
    'local'
```

I get same result when i change above schema from ``type: local`` to ``type: slurm``.

Now trying the combined schema from regtest but swap the two executor names and put ``vars: 1`` and it gets same result

```
editor: vi
executors:
  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  local:
    type: local
    vars: 1
```

```
jsonschema.exceptions.ValidationError: 'local' is not one of ['slurm']

Failed validating 'enum' in schema[1]['properties']['type']:
    {'enum': ['slurm'], 'type': 'string'}

On instance['type']:
    'local'
```

Now this last one is a bit interesting, apparently I can define two executors of same name ``local`` and the validate passes

```
editor: vi
executors:
  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  local:
    type: local

  local:
    type: slurm
    launcher: sbatch
```

For now i removed the two lines as i referenced in previous message.

shahzebsiddiqui: all my test were conducted with the following commit https://github.com/HPC-buildtest/buildtest-framework/pull/249/commits/3c1b5d8a7386d0aaeedf059e9523f6112bcbc299

shahzebsiddiqui: @vsoch what your thoughts regarding my results? Do you think the issue is resolved?

vsoch: I don't see a case where you add additional variables to an executor and have the expected failure, so I don't think so. If you want me to give it a shot I can make time for it this week.

shahzebsiddiqui: I think that would be good use-case of testing bad configuration examples as part of regression test. Maybe put them in this directory https://github.com/HPC-buildtest/buildtest-framework/tree/devel/tests/config_schema_examples 

That can be handled in a separate PR.

vsoch: Yes, but the issue we were addressing is having undefined fields in the executors, you asked me if that is resolved, and it is not.

vsoch: @shahzebsiddiqui for the last case, when you load the schema it will actually only have *one* of those executors because it's read as a dictionary. There might be a way to prevent duplicate keys but I haven't found it yet - we couldn't check for it in python because it would only load one of the sections.

vsoch:
I'm testing locally, and this config:
```
editor: vi
executors:
  local:
    type: local
  normal:
    type: slurm
    launcher: sbatch
    extraOption: this-is-invalid
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
```
should be invalid, but it's not.

vsoch:
Also why did you copy the slurm-example.yml into the testing folder, if it's already provided as an example in the executors directory? I'm totally confused...

tests/config_schema_examples/slurm-example.yml
shahzebsiddiqui: If I may let's document and handle testing of configuration check in separate issue. It's out of scope and not the intent when I was issuing the PR.

shahzebsiddiqui:
Let me add #251 to this PR that should be relatively simple and related to this. I will update PR in afternoon once I am back at desk.

shahzebsiddiqui:
> Also why did you copy the slurm-example.yml into the testing folder, if it's already provided as an example in the executors directory? I'm totally confused...

The file ``buildtest/settings/slurm-example.yml`` was added, it meant as starting point and used for documentation purposes as well. For file ``tests/config_example_schemas/slurm-example.yml`` I just updated the content so that it matches new fields such as ``options``, ``launcher`` and check that the test passes. Quickest way was to copy all content from one provided in ``buildtest/settigms/slurm-example.yml``. The slurm configuration used for regtest can diverge in future as serves the purpose of  validating slurm configuration. Hope that makes sense.

Also we should store some invalid test configuration that should fails check from ``jsonschema.validate`` some of which we are discussing already. That can also be stored in ``config_example_schemas`` repo.

vsoch:
I'm actually doing the default now, so please hold off on adding more to this PR.

vsoch:
> Also we should store some invalid test configuration that should fails check from jsonschema.validate some of which we are discussing already. That can also be stored in config_example_schemas repo.

Great idea! Since it's getting confusing, let's have an "invalid" and "valid" folder to go from, e.g.,

```
config_test_examples/
   invalid/
   valid/
      local-example.yml
      slurm-example.yml
```
Looking in the file for test_config.py, it's not clear to me why the two tests (for the files mentioned above) are structured differently? One checks for validation error, and the other just runs the validate function. Could we be consistent here?

shahzebsiddiqui:
ook that's fine, thx 4 heads up

shahzebsiddiqui:
> > Also we should store some invalid test configuration that should fails check from jsonschema.validate some of which we are discussing already. That can also be stored in config_example_schemas repo.
> 
> Great idea! Since it's getting confusing, let's have an "invalid" and "valid" folder to go from, e.g.,
> 
> ```
> config_test_examples/
>    invalid/
>    valid/
>       local-example.yml
>       slurm-example.yml
> ```
> 
> Looking in the file for test_config.py, it's not clear to me why the two tests (for the files mentioned above) are structured differently? One checks for validation error, and the other just runs the validate function. Could we be consistent here?

Yea i see what you mean yea the one that runs ``validate`` is useful as pose to running a try block in ``test_config_slurm`` example. What i found is that ``validate`` prints useful information for debugging but wrapping in ``try`` block makes it useless. That was the same issue with configuration check (#246)

I am +1 for having ``invalid`` and ``valid`` folder for storing configs. 

I can see that testing different configuration templates may not increase coverage report if all we are doing is running ``validate`` . It's just a binary pass/fail. Unless we have other checks after jsonschema pass validation, but addition checks in buildtest to see validation of configuration file. Right now we are not doing anything fancy in the ``check_configuration`` stage.

shahzebsiddiqui:
@vsoch i have addressed both issues discussed in this PR in #253 #254 

If there is anything else missed, let's open an issue and come to resolution on this PR.

vsoch (approved):

PR #250, opened 2020-03-29
remove option buildtest build report.

shahzebsiddiqui:
This feature doesn't work based on current state since we dont write any result to build.json.
There is no example of buildtest build report in docs so removed the example
file docs/docgen/buildtest-report.txt and update docgen script.


vsoch (approved):
LGTM

PR #252, opened 2020-03-30
adding basic steps for a local (base) executor

### LINKED ISSUE #251: Method for selecting default executor

vsoch:
This pull request adds the basic interaction of an executor with the builders produced by the BuildTestConfig class. Basically, we create the BuildExecutors (executor) based on the config_opts
(the users loaded configuration file) and then parse through the builders and hand each one off to it. In the case of defining an --executor on the command line, we honor and use that choice. In the case of finding the executor as a metadata value in the config schema, we also honor that. In the case of no command option or no default, we either use the first in the file OR the base class (which is basically a local executor). This example should be sufficient to show how to set up the same for slurm - each of steps and dryrun_steps and the subsequent functions would need to
be defined. I'll probably need to update this PR with fixes to testing (didn't try locally).

### DIFF
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -103,6 +103,9 @@ class BuildExecutor:
         # Run each step defined for dry run
         for step in executor.steps:
             if getattr(executor, step, None):
+                executor.builder.logger.debug(
+                    "Running %s for executor %s" % (step, executor)
+                )
                 getattr(executor, step)()
         return executor.result
 
@@ -135,7 +138,7 @@ class BaseExecutor:
        executors must have a listing of steps and dryrun_steps
     """
 
-    steps = ["setup", "build"]
+    steps = ["setup", "run"]
     dryrun_steps = ["setup", "dry"]
     type = "base"
 
@@ -156,7 +159,7 @@ class BaseExecutor:
         self._settings = settings
         self.load(name)
         self.builder = None
-        self.result = None
+        self.result = {}
 
     def load(self, name=None):
         """Load a particular configuration based on the name. This method
@@ -169,6 +172,7 @@ class BaseExecutor:
         """Setup the executor, meaning we check that the builder is defined,
            the only step needed for a local (base) executor.
         """
+        print(self.builder)
         if not self.builder:
             sys.exit("Builder is not defined for executor.")
 
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -7,7 +7,6 @@ import os
 import sys
 
 from buildtest.config import config_opts
-from buildtest.log import init_logfile, init_log
 
 
 class BuildExecutor:
@@ -20,13 +19,15 @@ class BuildExecutor:
        provided, we use reasonable defaults.
     """
 
-    def __init__(self, config_opts):
+    def __init__(self, config_opts, default=None):
         """initiate executors, meaning that we provide the config_opts
            that are validated, and can instantiate each executor to be available
            
-           Parameters:
-
-           config_opts: the validated config opts provided by buildtest.
+           :param config_opts: the validated config opts provided by buildtest.
+           :type config_opts: dictionary
+           :param default: the name of the default executor. If not set (or undefined)
+           we fall back to buildtest defaults (see set_default).
+           :type default: str
         """
         self.executors = {}
 
@@ -37,6 +38,8 @@ class BuildExecutor:
             elif executor["type"] == "slurm":
                 self.executors[name] = SlurmExecutor(name, executor)
 
+        self.set_default(default)
+
     def __str__(self):
         return "[buildtest-executor]"
 
@@ -45,14 +48,96 @@ class BuildExecutor:
 
     def get(self, name):
         """Given the name of an executor return the executor for running 
-           a buildtest build.
+           a buildtest build, or get the default.
         """
-        return self.executors.get(name)
+        return self.executors.get(name) or self.executors.get("default")
+
+    def _choose_executor(self, builder):
+        """Choose executor is called at the onset of a run or dryrun. We
+           look at the builder metadata to determine if a default
+           is set for the executor, and fall back to the default.
+
+           :param builder: the builder with the loaded test configuration.
+           :type builder: buildtest.buildsystem.BuilderBase (or subclass).
+        """
+        executor = builder.metadata.get("executor", "default")
+
+        # The executor (or a default) must be define
+        if executor not in self.executors:
+            builder.logger.warning(
+                "executor %s is not defined in default.yml" % executor
+            )
+            sys.exit(1)
+
+        # Get the executor by name, and add the builder to it
+        executor = self.executors.get(executor)
+        executor.builder = builder
+        return executor
+
+    def dry_run(self, builder):
+        """A dry run typically includes all of the steps up to run
+
+           :param builder: the builder with the loaded test configuration.
+           :type builder: buildtest.buildsystem.BuilderBase (or subclass).
+        """
+        # Choose the executor based on the builder provided
+        executor = self._choose_executor(builder)
+
+        # Run each step defined for dry run
+        for step in executor.dryrun_steps:
+            if getattr(executor, step, None):
+                getattr(executor, step)()
+        return executor.result
+
+    def run(self, builder):
+        """Given a buildtest.buildsystem.BuildConfig (subclass) go through the
+           steps defined for the executor to run the build. This should
+           be instantiated by the subclass. For a simple script run, we expect a 
+           setup, build, and finish.
+
+           :param builder: the builder with the loaded test configuration.
+           :type builder: buildtest.buildsystem.BuilderBase (or subclass).
+        """
+        executor = self._choose_executor(builder)
+
+        # Run each step defined for dry run
+        for step in executor.steps:
+            if getattr(executor, step, None):
+                getattr(executor, step)()
+        return executor.result
+
+    def set_default(self, name=None):
+        """Set a particular name as the default executor (defaults to default).
+           In the case that no executors are defined, we return a base (local)
+           executor. If executors are defined, we return the first in the list.
+
+           :param name: the name for the key to set as default, if it exists.
+           :type name: string
+        """
+
+        # If a default is not defined
+        if "default" not in self.executors:
+
+            # If no executors are defined, return a base (local)
+            if not self.executors:
+                self.executors["default"] = BaseExecutor("default", {})
+
+            # Otherwise return the first in the list
+            else:
+                if name in self.executors:
+                    self.executors["default"] = self.executors[name]
+                else:
+                    self.executors["default"] = list(self.executors.values())[0]
 
 
 class BaseExecutor:
-    """The BaseExecutor is an abstract base class for all executors
+    """The BaseExecutor is an abstract base class for all executors. All
+       executors must have a listing of steps and dryrun_steps
     """
+
+    steps = ["setup", "build"]
+    dryrun_steps = ["setup", "dry"]
+    type = "base"
 
     def __init__(self, name, settings):
         """Initiate a base executor, meaning we provide a name (also held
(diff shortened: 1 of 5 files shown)


vsoch:
@shahzebsiddiqui this provides the example of the local executor, which should be enough for you to write the SlurmExecutor class to follow suit. I also wrote notes in the docstring for that class for the steps that you suggested - you can look at the LocalExecutor to see how steps work.

buildtest/menu/build.py
```diff
@@ -9,11 +9,14 @@
 import sys
 
 from buildtest.defaults import (
+    DEFAULT_CONFIG_FILE,
```
shahzebsiddiqui: This line is not needed, its not being used in this file

buildtest/executors/base.py
```diff
@@ -45,15 +48,97 @@ def __repr__(self):
...
+           :param builder: the builder with the loaded test configuration.
+           :type builder: buildtest.buildsystem.BuilderBase (or subclass).
+        """
+        executor = self._choose_executor(builder)
+
+        # Run each step defined for dry run
+        for step in executor.steps:
+            if getattr(executor, step, None):
+                getattr(executor, step)()
+        return executor.result
```
shahzebsiddiqui, line 110: I am not understanding this ``run`` method from BuildExecutor that is returning value to variable ``result`` in https://github.com/researchapps/buildtest-framework/blob/add/executor-steps/buildtest/menu/build.py#L108 

I tested it locally it is returning ``None`` and getting the following error.

```
  File "/workspace/.pip-modules/bin/buildtest", line 11, in <module>
    load_entry_point('buildtest-framework', 'console_scripts', 'buildtest')()
  File "/workspace/buildtest-framework/buildtest/main.py", line 20, in main
    parser.parse_options()
  File "/workspace/buildtest-framework/buildtest/menu/__init__.py", line 80, in parse_options
    args.func(args)
  File "/workspace/buildtest-framework/buildtest/menu/build.py", line 111, in func_build_subcmd
    if result["RETURN_CODE"] == 0:
TypeError: 'NoneType' object is not subscriptable
```

buildtest/executors/base.py
```diff
@@ -45,15 +48,97 @@ def __repr__(self):
...
+
+    def set_default(self, name=None):
```
shahzebsiddiqui, line 112: i am a bit confused at how ``set_default`` method is working to get default executor. It's not clear to me yet and would like to understand this with example yaml configuration. 

Right now if i recall we don't have a key ``default`` in the schema under ``executors``. We assume we get the first entry in the list or am i wrong. Looking at your branch configuration https://github.com/researchapps/buildtest-framework/blob/add/executor-steps/buildtest/settings/default.schema.json i don't see any changes made so i take it there is no default executor.

shahzebsiddiqui (commented):
I think one thing that is confusing me is the name ``BuildExecutor`` and ``BaseExecutor``. I get BaseExecutor, it's base class and there would be a subclass per executor type (local, slurm). 

I think BuildExecutor is getting the list of executors from the configuration file ``~/.buildtest/settings.yml``. It may be helpful to change the name ``BuildExecutor`` to another name.

What also confuses me is why does ``BuildExecutor`` has methods for ``dry_run`` and ``run`` when its also part of ``BaseExecutor`` class and could be overridden in subclass. I think it's bit hard wrapping my mind around the scope of BuildExecutor class.

buildtest/menu/build.py
vsoch: Fixed locally.

buildtest/executors/base.py
vsoch: Can you please provide enough to reproduce this?

vsoch: okay reproduced with
```
$ buildtest build -c tests/testdir/slurm-hello.yml
```

vsoch: okay fixed the bug - one tiny typo! The step is called "run" and not "build" and I forgot to change it :)

buildtest/executors/base.py
vsoch: Okay, so I'll walk through this. The function there shows the following:

1. If the user has an executor named "default" in their config, this is treated as the default. There is no magic there, it can be of any type.
2. If the user has _no_ executors defined, then we instantiate a BaseExecutor, which is basically a LocalExecutor (take a look at LocalExecutor,it just adds type "local". This could probably easily be LocalExecutor too if you think that is more clear.
3. If the user has at least one executor, the first is taken as the default.

However that flow is agnostic to the command line, and any executor settings in the file! If (at runtime) the user specifies the `--executor` this overrides the default choices above, e.g., this part:

```python
        parser_build.add_argument(
            "-e",
            "--executor",
            help='specify a named executor to use (defaults to "default" key, first in list, then local (base) executor',
        )
```
The argument goes into the instantiation of the BuildExecutor

```
    # Load BuildExecutors
    executor = BuildExecutor(config_opts, default=args.executor)
```
and then that key goes into the same function above:

```
        self.set_default(default)
```
And the same logic persists - the default is set to be that key (or the other options if it doesn't exist).

I set it up exactly like this so that the default executor for most users will bet the first in their list of defined, which (based on our starting config) should be local.

vsoch:
They are actually different - the BuildExecutor is the _one_ class that holds all the executors, determines logic for default, etc. If you look in the build menu/build.py setup, we instantiate a _single_ BuildExecutor and then iterate over it's builders. The builders all have base class BuilderExecutor, since that is what they are. A BaseExecutor is _not_ a child of a BuildExecutor, it's another class entirely - the _base_ for a build executor. If you have another naming setup that you've prefer, this would be easiest enough to change. Sorry it's confusing for you!

buildtest/executors/base.py
vsoch: Ah, one more detail I forgot (and I know this is important to you) when the BuildExecutor (holding all the executor subtypes) is created, it further chooses the executor based on a setting in the test config file:

```
    def _choose_executor(self, builder):
        """Choose executor is called at the onset of a run or dryrun. We
           look at the builder metadata to determine if a default
           is set for the executor, and fall back to the default.

           :param builder: the builder with the loaded test configuration.
           :type builder: buildtest.buildsystem.BuilderBase (or subclass).
        """
        executor = builder.metadata.get("executor", "default")

        # The executor (or a default) must be define
        if executor not in self.executors:
            builder.logger.warning(
                "executor %s is not defined in default.yml" % executor
            )
            sys.exit(1)

        # Get the executor by name, and add the builder to it
        executor = self.executors.get(executor)
        executor.builder = builder
        return executor
```
I think instead of hard coding "default" I should use the same variable that was possibly set on the command line with `--executor`.

vsoch: Actually scratch that, the "default" executor key is already set to be whatever they specified.

shahzebsiddiqui: Thank you for the detailed explanation, i recall from #251 we decided to have a key ``default`` in executor section (Option 1) so i take it that would be an optional key for user but required for schema. 

When you say 

> If the user has an executor named "default" in their config, this is treated as the default. There is no magic there, it can be of any type.

Does that mean we have something like 

```
editor: vi
executors:
  default: runlocal
  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  runlocal:
    type: local
```

Or is it we have a default executor something like 

```
editor: vi
executors:
  normal:
    type: slurm
    launcher: sbatch
    options:
      - "-p normal"
      - "-t 00:10"
      - "-N 1"
      - "-n 1"
  default:
    type: local
```

Or something else?

shahzebsiddiqui: I can see value of having ``--executor`` option to run a test with any executor, practically i dont know if test will be written in way to handle on the fly change to executor.  The ``executor`` key is one per ``subtest`` https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json#L39

I think ``--executor`` should override if ``executor`` key is not defined in YAML file (test), but if defined in file i would rather not have an option change that. Reason is if i process one or more subtest in a single YAML file or multiple YAML files (directory based) then the behavior can be unpredictable.

What i mean is if i run the following: ``buildtest build -c examples/script -e admin`` this means all test run with ``admin`` executor, it could be that this breaks many test.
 
Again it out of scope for this discussion and I dont have clear idea how it will work, but it doesn't hurt having the option for now.

vsoch: @shahzebsiddiqui during the implementation I realized it would get very hard to validate a "special" key value pair for default in the list, and it made more sense to have default just as a key. If you'd like it otherwise please feel free to figure out how to write the schema validator to allow it.

The second example is what I had in mind!

I think command line should generally override all things - it's the highest level of specifying how you want something to run, so I'm not sure I agree with you. I could easily have an entire site worth of test configs that I want to test (with my specific executor) and then I would have to manually change the files to make it work, instead of just using `--executor`. For your example case, the user would be best advised to use different subfolders.

```
buildtest build -c examples/my_executor -e admin
buildtest build -c examples/default_executor
```
Honestly I think we're in a situation where no single solution is going to be perfect for everyone. But generally, when you have config files and command line arguments, the command line args are at the highest level and generally override. It would be very confusing to set a command line `--executor` and have it ignored.

shahzebsiddiqui: > @shahzebsiddiqui during the implementation I realized it would get very hard to validate a "special" key value pair for default in the list

Makes sense to me and would be fine with ``default`` as name of an executor. 

Yea i agree with you the command line supersedes all things and that's what i have observed.   I think practically using buildtest to write test would help determine how this will work out.

shahzebsiddiqui:
> They are actually different - the BuildExecutor is the _one_ class that holds all the executors, determines logic for default, etc. If you look in the build menu/build.py setup, we instantiate a _single_ BuildExecutor and then iterate over it's builders. The builders all have base class BuilderExecutor, since that is what they are. A BaseExecutor is _not_ a child of a BuildExecutor, it's another class entirely - the _base_ for a build executor. If you have another naming setup that you've prefer, this would be easiest enough to change. Sorry it's confusing for you!

I realize the difference i dont have an objection on how things are setup right now, i do like the ``BaseExecutor``, ``LocalExecutor`` and ``SlurmExecutor`` to implement executor specific logic such as the ``steps`` section specific to each executor.

I think it makes sense to call ``BuildExecutor`` once which is a list of executors , we would need to determine the executor list by checking the buidltest configuration  and also ``executor`` defined in test configuration. 

I can foresee for slurm execution i'd want to submit all the jobs once and poll them afterwards. It's more efficient than going through the for loop and poll after each job https://github.com/researchapps/buildtest-framework/blob/add/executor-steps/buildtest/menu/build.py#L96 I am just thinking ahead, and would need to think about the logic we have setup in menu/build.py

vsoch:
> I think it makes sense to call BuildExecutor once which is a list of executors , we would need to determine the executor list by checking the buidltest configuration and also executor defined in test configuration.

Yep that's exactly what it's doing.

shahzebsiddiqui (approved):

vsoch:
Woohoo!

shahzebsiddiqui:
@vsoch  i see with latest change fixed the bug I reported earlier with ``result["RESULT_CODE"]`` and test were running to completion. Thanks for your work! :)

PR #255, opened 2020-04-01
Reorg valid config schema for regtest

### LINKED ISSUE #254: Provide a set of valid and invalid configuration files for regression testing
Currently we don't have distinction between valid and invalid files, all the files are in ``tests/config_schema_examples`` we can introduce ``valid`` and ``invalid`` sub-directories to store valid and invalid configuration files for testing purposes. This allows for proper organization on where to store configuration examples and write tests for intended purpose. Both valid and invalid tests need to be written to ensure json schema is working properly.

```
config_schema_examples/
   invalid/
   valid/
      local-example.yml
      slurm-example.yml
```

shahzebsiddiqui:
Create directory examples/config_schemas/valid and store all valid configuration here.

Remove test so we have one test to validate all yaml files. Its sufficient to run ``validate``
against all examples since we expect this to pass. If it fails it will break on the first validate
stage and test will be marked fail.
Change logic on file extension when loading schema. Ensure that .yml, .yaml, and .json are searched
otherwise any arbitrary file can be loaded with extension (ml, or json) without a period.

Fix #254


buildtest/buildsystem/schemas/utils.py
```diff
@@ -25,9 +25,9 @@ def load_schema(path):
     if not os.path.exists(path):
         sys.exit("schema file %s does not exist." % path)
     with open(path, "r") as fd:
-        if path.endswith("json"):
+        if path.endswith(".json"):
             schema = json.loads(fd.read())
-        elif path.endswith("ml"):
+        elif path.endswith(".yaml") or path.endswith(".yml"):
```
vsoch: Nit: this would be more elegant with:

```python
if re.search("(yaml|yaml)^", path):
```
Generally when you have to do more than one if this, if that, it's good to try and look for a regular expression that can represent both.

tests/menu/test_config.py
```diff
@@ -21,26 +22,10 @@ def test_config_restore():
     os.remove(BUILDTEST_CONFIG_BACKUP_FILE)
     func_config_restore()
 
+def test_valid_config_schemas():
 
-def test_config_local():
-    example_schema = os.path.join(
-        pytest_root, "config_schema_examples", "local-example.yml"
-    )
+    valid_schema_dir = os.path.join(pytest_root,"examples","config_schemas","valid")
```
vsoch: Does this validate with black?

vsoch (changes requested):

tests/menu/test_config.py
shahzebsiddiqui: i ran black against all files, seems like line 27 ``valid_schema_dir`` didn't change so i guess it is valid with black.

buildtest/buildsystem/schemas/utils.py
shahzebsiddiqui: thats good suggestion. Regex are very powerful and this would be good example. I need to get more comfortable with regex, i have some understanding but practically haven't used it much in the codebase as you can see.

shahzebsiddiqui:
is this good to merge @vsoch ?

buildtest/buildsystem/schemas/utils.py
```diff
@@ -25,9 +25,9 @@ def load_schema(path):
     if not os.path.exists(path):
         sys.exit("schema file %s does not exist." % path)
     with open(path, "r") as fd:
-        if path.endswith("json"):
+        if re.search(".json$", path):
             schema = json.loads(fd.read())
-        elif path.endswith("ml"):
+        elif re.search("(.yaml|.yml)$", path):
             schema = yaml.load(fd.read(), Loader=yaml.SafeLoader)
```
vsoch, line 31: This isn't matching what you think - a period means "match anything" For example:

```python
re.search("(.yaml|.yml)$", 'ayaml')                                     
<re.Match object; span=(0, 5), match='ayaml'>
```
What you want is an explicit period, and since it's shared between the two, put it outside of the or:

```python
re.search("[.](yaml|yml)$", 'ayaml')                                    
# No match (correct)
re.search("[.](yaml|yml)$", '.yaml')                                    
<re.Match object; span=(0, 5), match='.yaml'>
re.search("[.](yaml|yml)$", 'config.yaml')                              
<re.Match object; span=(6, 11), match='.yaml'>
```
The same is true for the json above - it needs to be
```python
 re.search("[.]json$", path):
```

vsoch (changes requested):
Almost! One small bug with the regular expressions.

buildtest/buildsystem/schemas/utils.py
shahzebsiddiqui: ahh that's good catch! i totally forgot about the period since it's special character in regex. I will fix this.

vsoch (approved):

PR #260, opened 2020-04-01
remove backup configuration settings.yml.bak.

### LINKED ISSUE #259: decide what to do with restore config option (buildtest config restore) or add a save option
Currently, we have option to restore site configuration from backup file, this is via ``buildtest config restore``

The first time, buildtest is issued, it will copy the default configuration to ``settings.yml`` and ``settings.yml.bak`` which are duplicate files. User can  make change to site configuration and this may mean ``settings.yml`` can diverge from original file.

Currently, i dont see a use-case of restore from backup, since i can just delete the file and buildtest will copy it.

Another option would be to have a mechanism to save site configuration something like (``buildtest config save``) which will save the site configuration by making a backup copy. Then user can make change as needed and if they screw up, then can restore the file via ``buildtest config restore``. This only makes sense from command line perspective.

If it doesn't make sense to have ``save`` feature we can just remove the ``restore`` operation

shahzebsiddiqui:
Rename ``buildtest config restore`` --> ``buildtest config reset``
Remove unused imports
Refactor regression test for restore test. We remove file and check if
file exist after restore.

### DIFF
--- a/tests/menu/test_config.py
+++ b/tests/menu/test_config.py
@@ -1,10 +1,9 @@
 import os
-import pytest
 from jsonschema import validate, ValidationError
-from buildtest.defaults import BUILDTEST_CONFIG_BACKUP_FILE, DEFAULT_CONFIG_SCHEMA
+from buildtest.defaults import DEFAULT_CONFIG_SCHEMA, BUILDTEST_CONFIG_FILE
 from buildtest.menu.config import (
     func_config_view,
-    func_config_restore,
+    func_config_reset,
 )
 from buildtest.buildsystem.schemas.utils import load_schema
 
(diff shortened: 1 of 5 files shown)


tests/menu/test_config.py
```diff
@@ -16,10 +15,11 @@ def test_view_configuration():
...
 def test_config_restore():
-    func_config_restore()
-    # removing backup file and testing of restore works
-    os.remove(BUILDTEST_CONFIG_BACKUP_FILE)
-    func_config_restore()
+
+    # removing config file and testing if reset works
+    os.remove(BUILDTEST_CONFIG_FILE)
+    func_config_reset()
+    assert True == os.path.exists(BUILDTEST_CONFIG_FILE)
```
vsoch: I think here you could just put
```python
assert(os.path.exists(BUILDTEST_CONFIG_FILE))
```

shahzebsiddiqui: good catch!

shahzebsiddiqui: Fixed now

vsoch (approved):
When travis is green, this LGTM.

tests/menu/test_config.py
shahzebsiddiqui: seems like black changes ``assert(os.path.exists(BUILDTEST_CONFIG_FILE))`` to ``assert os.path.exists(BUILDTEST_CONFIG_FILE)`` either way it still passes check

PR #261, opened 2020-04-02
Add support for specifying buildtest settings from cli 

### LINKED ISSUE #258: option to specify a different site configuration
Currently there is no support for user to specify a different site configuration, it must be stored in ``$HOME/.buildtest/settings.yml``. 

Logically, it makes sense for ``buildtest build`` to have an option to specify path to site configuration they want to use instead of copying file everytime, this would make sense when a file sits in source control that user wants to make change instead of copying or moving file back-forth.

The option can be ``buildtest build -s`` or long option ``--site`` or ``--site-config``. I am fine with with either long name. This could be relatively easy to do, currently we don't have ``-s`` so it could be a viable short option to use.

```
(docs) login4.stampede2(1035)$ buildtest build -h
usage: buildtest [options] [COMMANDS] build [-h] [-c TEST CONFIGURATION] [-t TESTDIR] [-e EXECUTOR] [-d]

optional arguments:
  -h, --help            show this help message and exit
  -c TEST CONFIGURATION, --config TEST CONFIGURATION
                        Specify test configuration file.
  -t TESTDIR, --testdir TESTDIR
                        specify a custom test directory. By default, use .buildtest in $PWD.
  -e EXECUTOR, --executor EXECUTOR
                        specify a named executor to use (defaults to "default" key, first in list, then local (base)
                        executor
  -d, --dry             dry-run mode, buildtest will not write the test scripts but print content of test that would
                        be written

```
 
This does mean that ``buildtest config view`` and ``buildtest config edit`` won't work with custom path to site config since they are looking at the default location.

Let me know your thoughts @vsoch. I would use this feature once i put site configuration for Stampede in https://github.com/HPC-buildtest/buildtest-stampede2

shahzebsiddiqui:
This PR will fix #258 
- It will add ``--settings`` option
- I had to move the Check Configuration (``check_configuration``)  after we got to entry point to ``buildtest build`` method. This was to ensure we can check the settings file if specified from command line or use the default one.
- I removed BUILDTEST_BUILD_LOGFILE for reasons specified in commit message
- I removed ``config_opts`` from buildtest/config.py and now this can be retrieved via ``get_default_configuration`` note this is only used in ``buildtest config edit`` for time being and only one reference is made to method ``get_default_configuration``. 
-It seems okay for me to pass the variable ``config_opts`` to other class instead of importing from ``buildtest/config``.

### DIFF
--- a/buildtest/config.py
+++ b/buildtest/config.py
@@ -9,7 +9,7 @@ from buildtest.defaults import (
     BUILDTEST_CONFIG_FILE,
     BUILDTEST_ROOT,
     DEFAULT_CONFIG_FILE,
-    DEFAULT_CONFIG_SCHEMA,
+    DEFAULT_CONFIG_SCHEMA
 )
 from buildtest.buildsystem.schemas.utils import load_schema
 
@@ -53,7 +53,7 @@ def init():
     create_logfile()
 
 
-def check_configuration():
+def check_configuration(config_path=None):
     """Checks all keys in configuration file (settings/default.yml) are valid
        keys and ensure value of each key matches expected type . For some keys
        special logic is taken to ensure values are correct and directory path
@@ -64,12 +64,19 @@ def check_configuration():
        :rtype: exit code 1 if checks failed
     """
 
+    user_schema = load_schema(config_path)
     config_schema = load_schema(DEFAULT_CONFIG_SCHEMA)
-    validate(instance=config_opts, schema=config_schema)
+    validate(instance=user_schema, schema=config_schema)
 
 
 def load_configuration(config_path=None):
-    """Load the default configuration file."""
+    """Load the default configuration file if no argument is specified.
+
+       Parameters:
+
+       :param config_path: Path to buildtest configuration file
+       :type config_path: str, optional
+    """
 
     init()
 
@@ -79,5 +86,7 @@ def load_configuration(config_path=None):
     return load_schema(config_path)
 
 
-# Run on init, so we only load once
-config_opts = load_configuration()
+def get_default_configuration():
+    """Load and return the default buildtest configuration file. """
+
+    return load_configuration()
--- a/buildtest/config.py
+++ b/buildtest/config.py
@@ -5,7 +5,6 @@ from jsonschema import validate
 
 from buildtest.utils.file import create_dir
 from buildtest.defaults import (
-    BUILDTEST_BUILD_LOGFILE,
     BUILDTEST_CONFIG_FILE,
     BUILDTEST_ROOT,
     DEFAULT_CONFIG_FILE,
@@ -19,15 +18,6 @@ def create_config_files():
 
     if not os.path.exists(BUILDTEST_CONFIG_FILE):
         shutil.copy(DEFAULT_CONFIG_FILE, BUILDTEST_CONFIG_FILE)
-
-
-def create_logfile():
-    """Create a logfile to keep track of messages for the user, if doesn't exist."""
-
-    if not os.path.exists(BUILDTEST_BUILD_LOGFILE):
-        build_dict = {"build": {}}
-        with open(BUILDTEST_BUILD_LOGFILE, "w") as outfile:
-            json.dump(build_dict, outfile, indent=2)
 
 
 def init():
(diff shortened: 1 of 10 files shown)


vsoch:
Looks like there is an issue with the config file check?
image

shahzebsiddiqui:
give me a sec @vsoch  i am fixing the regtest.

shahzebsiddiqui:
Let's wait for Travis report i tested locally to increase coverage report for ``buildtest/config.py``  and fix the Travis bug.

shahzebsiddiqui:
@vsoch Travis is green. I see 100% coverage on ``buildtest/config.py`` that is good.  It's not showing in codecov but i can see in coveralls at https://coveralls.io/builds/29803302
image

buildtest/config.py
```diff
@@ -79,5 +74,7 @@ def load_configuration(config_path=None):
     return load_schema(config_path)
 
 
-# Run on init, so we only load once
-config_opts = load_configuration()
```
vsoch, line 83: I'm worried about this line - we are removing the ability to import config_opts from anywhere (you don't think this is important - to load them at init onset and then not need to do it again?) It would mean that there could be a race condition of loading some config opts, having it changed, and then re-loading all during a run.

That's my only concern with this change - let me know your thoughts.

vsoch (commented):
I was worried about removing the ability to import config_opts from anywhere (you don't think this is important - to load them at init onset and then not need to do it again?) That's my only concern with this change - let me know your thoughts.

buildtest/config.py
shahzebsiddiqui: I observed this change when i was trying to implement this feature and noticed that ``config_opts`` was only loading default configuration (``~/.buildtest/settings.yml``) and it was only imported in two different files

https://github.com/HPC-buildtest/buildtest-framework/blob/74735173b772e924eb7504c01a7e70550ca3cb76/buildtest/menu/config.py#L3

https://github.com/HPC-buildtest/buildtest-framework/blob/74735173b772e924eb7504c01a7e70550ca3cb76/buildtest/menu/build.py#L17

So i figured that if we need to retrieve ``config_opts`` instead of loading this value we can import and call method ``get_default_configuration`` https://github.com/shahzebsiddiqui/buildtest-framework/blob/issue/258/buildtest/config.py#L77 which would be equivalent to using config_opts. 

The issue was having to either preserve ``config_opts`` as a variable and have this passed to other methods while supporting a new configuration file which needs to load the configuration and get an updated ``config_opts`` that can be passed to ``BuildExecutor`` class

https://github.com/shahzebsiddiqui/buildtest-framework/blob/0c5a260dc338417281b6cc57e5d7f1df71e7e6a4/buildtest/menu/build.py#L99

I was also thinking that checking configuration should be done at entry point of buildtest build as pose to before calling the parser. This was to support the command line option and not have to do load configuration again. That's why i removed line https://github.com/HPC-buildtest/buildtest-framework/pull/261/files#diff-b01bc590c712803d7252b15aa8dbd8bcL18

vsoch: okay, I'm not sure that I totally am totally comfortable with this change, but I think you've put a good amount of thought into it and I don't think it hurts too badly to load on demand.

vsoch (approved):

buildtest/config.py
shahzebsiddiqui: I have ran into this cyclic import issue a while back about time when i was refactoring code about time when you joined. I did like the ability to load configuration once and pass ``config_opts`` as variable to other modules, but right now it not being used much so didnt make sense to keep it and find way to add this feature. If codebase changes where ``config_opts`` is needed by several python modules, i think then we should seriously consider defining as a global variable and call it once. It's not good to call ``get_default_configuration`` or ``load_configuration`` several times as it reading file every-time. 

Also note i call ``load_configuration`` and ``check_configuration`` https://github.com/HPC-buildtest/buildtest-framework/pull/261/files#diff-dd4fa250655dc41a1197ad6dc7b8a126R82-R85 

Even though check_configuration calls load_configuration, I wanted these to be separate methods. This way check_configuration is done once while load_configuration can be called one or more times as needed to retrieve the configuration as a dict.

PR #264, opened 2020-04-05
Change output behavior of buildtest build, add system log, add .err file per test

### LINKED ISSUE #256: Change output behavior of buildtest build
This output looks ugly and should design a more elegant way to display content of test being build and run. 
```
gitpod /workspace/buildtest-framework $ buildtest build -c examples/script/slurm.yml 
                             start time: 2020-04-01 16:52:32.036899
                     configuration file: /workspace/buildtest-framework/examples/script/slurm.yml
                                testdir: /workspace/buildtest-framework/.buildtest/slurm
                               testpath: /workspace/buildtest-framework/.buildtest/slurm/slurm_down_nodes_reason.sh
                                logpath: None



STAGE                                    VALUE
[WRITING TEST]                           PASSED
[RUNNING TEST]                           FAILED
Writing results to /workspace/buildtest-framework/.buildtest/slurm/run/slurm_down_nodes_reason_04-01-2020-16-52.out
                             start time: 2020-04-01 16:52:32.043177
                     configuration file: /workspace/buildtest-framework/examples/script/slurm.yml
                                testdir: /workspace/buildtest-framework/.buildtest/slurm
                               testpath: /workspace/buildtest-framework/.buildtest/slurm/slurm_not_responding_nodes.sh
                                logpath: None



STAGE                                    VALUE
[WRITING TEST]                           PASSED
[RUNNING TEST]                           FAILED
Writing results to /workspace/buildtest-framework/.buildtest/slurm/run/slurm_not_responding_nodes_04-01-2020-16-52.out
Finished running 2 total tests.
                         Test summary                         
Executed 2 tests
Passed Tests: 0 Percentage: 0.0%
Failed Tests: 2 Percentage: 100.0%
```

### LINKED ISSUE #257: missing error file from test run
Every buildtest command should an error file that captures output from STDERR stream into file ``.err``

Currently we have ``.out`` that captures both STDOUT and STDERR stream which makes it hard to decipher between each test output and error stream. 

https://github.com/HPC-buildtest/buildtest-framework/blob/da713214e17bb98e27e54d346d9745372b9fa28e/buildtest/buildsystem/base.py#L461-L471

### LINKED ISSUE #265: one log central logfile per build not per each sub-test.
Currently, there is one log file per sub-test and not per build. It makes no sense to have a logger per subtest, since we care about log details from buildtest class when parsing and executing the test via executor. The ``.out`` and ``.err`` should capture of the test which should be sufficient.

For example if you see bzip2 example it has 4 log files one, per sub-test

```
gitpod /workspace/buildtest-framework/.buildtest/bzip2/log $ ll
total 24
drwxr-xr-x 2 gitpod gitpod 4096 Apr  5 20:43 ./
drwxr-xr-x 4 gitpod gitpod 4096 Apr  5 20:43 ../
-rw-r--r-- 1 gitpod gitpod  157 Apr  5 20:43 bzip2_help_04-05-2020-20-43.log
-rw-r--r-- 1 gitpod gitpod 1301 Apr  5 20:43 bzip_compression_04-05-2020-20-43.log
-rw-r--r-- 1 gitpod gitpod  535 Apr  5 20:43 bzip_help_04-05-2020-20-43.log
-rw-r--r-- 1 gitpod gitpod  916 Apr  5 20:43 bzip_version_04-05-2020-20-43.log
```

The content of each logfile is a subset of them depending on order the subtest were processed.

For example ``bzip2_help`` is the first subtest
```
gitpod /workspace/buildtest-framework/.buildtest/bzip2/log $ cat bzip2_help_04-05-2020-20-43.log 
2020-04-05 20:43:18,817 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip2_help.sh
```

The second test is ``bzip_help`` and notice its log file has content from previous logfile too

```
gitpod /workspace/buildtest-framework/.buildtest/bzip2/log $ cat bzip_help_04-05-2020-20-43.log 
2020-04-05 20:43:18,808 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip_help.sh
2020-04-05 20:43:18,815 [base.py:105 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
2020-04-05 20:43:18,816 [base.py:105 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]
2020-04-05 20:43:18,817 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip2_help.sh
```

This becomes evident when you look at the last subtest is ``bzip_compression``.  The content is repeated. 

```
gitpod /workspace/buildtest-framework/.buildtest/bzip2/log $ cat bzip_compression_04-05-2020-20-43.log 
2020-04-05 20:43:18,757 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip_compression.sh
2020-04-05 20:43:18,796 [base.py:105 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
2020-04-05 20:43:18,798 [base.py:105 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]
2020-04-05 20:43:18,799 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip_version.sh
2020-04-05 20:43:18,806 [base.py:105 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
2020-04-05 20:43:18,807 [base.py:105 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]
2020-04-05 20:43:18,808 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip_help.sh
2020-04-05 20:43:18,815 [base.py:105 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
2020-04-05 20:43:18,816 [base.py:105 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]
2020-04-05 20:43:18,817 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/bzip2/bzip2_help.sh
```

The logger file should be created by taking the name of the config name https://github.com/shahzebsiddiqui/buildtest-framework/blob/abcfad94b69598f87f15c58e3c0f0762d9c9a99f/buildtest/buildsystem/base.py#L229

It could be declared in the ``init`` method of BuilderBase class. We may need a separate logger to capture logging for buildtest that is independent of the config name. For example, we can't do logging until we know the name of test logger which is happens around https://github.com/shahzebsiddiqui/buildtest-framework/blob/devel/buildtest/buildsystem/base.py#L291 which is way too late

shahzebsiddiqui:
This PR will address the following fix
- Change output behavior #256 
- Add .err file #257 
- Add system logging #265 and remove log per test 
- Add additional logging throughout codebase
- Fix regtest issue where BuildConfig should not be called with any invalid config file, that's because all checks are done in advance in method ``discover_configs``
- Add additional system detection including environment variable, operating system details, and python path.

This first attempt to change output behavior of buildtest build.

The output looks as follows with left align all the fields.

```
gitpod /workspace/buildtest-framework $ buildtest build -c examples/script/
Config Name                    SubTest                        Status                         Config Path                   
slurm                          slurm_down_nodes_reason        PASSED                         /workspace/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     PASSED                         /workspace/buildtest-framework/examples/script/slurm.yml
zlib                           zlib_compression               PASSED                         /workspace/buildtest-framework/examples/script/zlib.yml
zlib                           zlib_decompress                PASSED                         /workspace/buildtest-framework/examples/script/zlib.yml
python-circle                  circle_area                    PASSED                         /workspace/buildtest-framework/examples/script/python-circle.yml
bzip2                          bzip_compression               PASSED                         /workspace/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip_version                   PASSED                         /workspace/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip_help                      PASSED                         /workspace/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip2_help                     PASSED                         /workspace/buildtest-framework/examples/script/bzip2.yml
                         Test summary                         
Executed 9 tests
Passed Tests: 7 Percentage: 77.77777777777777%
Failed Tests: 2 Percentage: 22.22222222222222%

```

There was some output displayed to stdout that was intended for logstream. Looking at the way ``init_logger`` was setup, it was not intended to work as it should so for now i removed a few log statement. I will cover the log issue in a separate issue to describe the issue.

The origin output as per devel head is the following

```
gitpod /workspace/buildtest-framework $ buildtest build -c examples/script/python-circle.yml 
2020-04-05 20:39:41,617 [base.py:105 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
[builder-script-circle_area]
2020-04-05 20:39:41,617 [base.py:105 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]
                             start time: 2020-04-05 20:39:41.617493
                     configuration file: /workspace/buildtest-framework/examples/script/python-circle.yml
                                testdir: /workspace/buildtest-framework/.buildtest/python-circle
                               testpath: /workspace/buildtest-framework/.buildtest/python-circle/circle_area.py
                                logpath: None



STAGE                                    VALUE
2020-04-05 20:39:41,618 [base.py:550 - _write_test() ] - [INFO] Opening Test File for Writing: /workspace/buildtest-framework/.buildtest/python-circle/circle_area.py
[WRITING TEST]                           PASSED
[RUNNING TEST]                           PASSED
Writing results to /workspace/buildtest-framework/.buildtest/python-circle/run/circle_area_04-05-2020-20-39.out
Finished running 1 total tests.
                         Test summary                         
Executed 1 tests
Passed Tests: 1 Percentage: 100.0%
Failed Tests: 0 Percentage: 0.0%
```

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -466,7 +466,7 @@ class BuilderBase:
         else:
             print(
                 "{:<30} {:<30} {:<30} {:<30}".format(
-                    self.config_name, self.name, "PASSED", self.config_file
+                    self.config_name, self.name, "FAILED", self.config_file
                 )
             )
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -376,7 +376,9 @@ class BuilderBase:
             os.path.join(self.testdir, self.name),
             self.get_test_extension(),
         )
-        self.metadata["testdir"] = os.path.dirname(os.path.expandvars(self.metadata["testpath"]))
+        self.metadata["testdir"] = os.path.dirname(
+            os.path.expandvars(self.metadata["testpath"])
+        )
 
         # The start time to print for the user
         self.metadata["start_time"] = datetime.datetime.now()
@@ -432,14 +434,10 @@ class BuilderBase:
         self.metadata["end_time"] = datetime.datetime.now()
 
         # Keep an output file
-        run_output_file = os.path.join(
-            self.metadata.get("rundir"), self.build_id
-        )
+        run_output_file = os.path.join(self.metadata.get("rundir"), self.build_id)
 
         # Keep an error file
-        run_error_file = os.path.join(
-            self.metadata.get("rundir"), self.build_id
-        )
+        run_error_file = os.path.join(self.metadata.get("rundir"), self.build_id)
 
         # write output of test to .out file
         with open(run_output_file + ".out", "w") as fd:
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -376,7 +376,7 @@ class BuilderBase:
             os.path.join(self.testdir, self.name),
             self.get_test_extension(),
         )
-        self.metadata["testdir"] = os.path.dirname(self.metadata["testpath"])
+        self.metadata["testdir"] = os.path.dirname(os.path.expandvars(self.metadata["testpath"]))
 
         # The start time to print for the user
         self.metadata["start_time"] = datetime.datetime.now()
@@ -433,20 +433,20 @@ class BuilderBase:
 
         # Keep an output file
         run_output_file = os.path.join(
-            self.metadata.get("rundir"), "%s.out" % self.build_id
+            self.metadata.get("rundir"), self.build_id
         )
 
         # Keep an error file
         run_error_file = os.path.join(
-            self.metadata.get("rundir"), "%s.err" % self.build_id
+            self.metadata.get("rundir"), self.build_id
         )
 
         # write output of test to .out file
-        with open(run_output_file, "w") as fd:
+        with open(run_output_file + ".out", "w") as fd:
             fd.write("\n".join(out))
 
         # write error from test to .err file
-        with open(run_error_file, "w") as fd:
+        with open(run_error_file + ".err", "w") as fd:
             fd.write("\n".join(err))
 
         result["RETURN_CODE"] = command.returncode
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -65,7 +65,9 @@ class BuildConfig:
         # Read the lookup to get schemas available
         self.lookup = get_schemas_available()
 
-        self.logger.debug(f"buildtest found the available schema: {self.lookup} in schema library")
+        self.logger.debug(
+            f"buildtest found the available schema: {self.lookup} in schema library"
+        )
 
         # Load the configuration file, fails on any error
         self.load(config_file)
@@ -520,7 +522,9 @@ class BuilderBase:
             testpath,
             stat.S_IRWXU | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH,
         )
-        self.logger.debug(f"Applying permission 755 to {testpath} so that test can be executed")
+        self.logger.debug(
+            f"Applying permission 755 to {testpath} so that test can be executed"
+        )
         return testpath
 
     def _get_test_lines(self):
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -252,7 +252,7 @@ class BuilderBase:
             )
 
         # Start with just terminal logger (changes to file on run)
-        self._init_logger(to_file=False)
+        # self._init_logger(to_file=False)
 
     def __str__(self):
         return "[builder-%s-%s]" % (self.type, self.name)
@@ -284,13 +284,13 @@ class BuilderBase:
            so the testpath is defined with the build id.
         """
 
-        if to_file:
-            self.metadata["logfile"] = os.path.join(
-                self.metadata["logdir"], "%s.log" % self.build_id
-            )
-            self.logger = init_logfile(self.metadata["logfile"])
-        else:
-            self.logger = init_log()
+        # if to_file:
+        self.metadata["logfile"] = os.path.join(
+            self.metadata["testdir"], "log", "%s.log" % self.build_id
+        )
+        self.logger = init_logfile(self.metadata["logfile"])
+        # else:
+        #    self.logger = init_log()
 
     def get_test_extension(self):
         """Return the test extension, which depends on the shell used. Based
@@ -444,6 +444,7 @@ class BuilderBase:
 
         # Run the test file using the shell
         cmd = [self.get_shell(), testfile]
+
         command = BuildTestCommand(cmd)
         out, err = command.execute()
 
@@ -455,31 +456,35 @@ class BuilderBase:
             self.metadata.get("rundir"), "%s.out" % self.build_id
         )
 
-        # Run the test file, print output to file
+        # Keep an error file
+        run_error_file = os.path.join(
+            self.metadata.get("rundir"), "%s.err" % self.build_id
+        )
+
+        # write output of test to .out file
         with open(run_output_file, "w") as fd:
+            fd.write("\n".join(out))
 
-            fd.write("Test Name:" + self.build_id + "\n")
-            fd.write("Return Code: %s \n" % command.returncode)
-
-            if out:
-                fd.write("---------- START OF TEST OUTPUT ---------------- \n")
-                fd.write("\n".join(out))
-                fd.write("------------ END OF TEST OUTPUT ---------------- \n")
-            if err:
-                fd.write("---------- START OF TEST ERROR ---------------- \n")
-                fd.write("\n".join(err))
-                fd.write("------------ END OF TEST ERROR ---------------- \n")
+        # write error from test to .err file
+        with open(run_error_file, "w") as fd:
+            fd.write("\n".join(err))
 
         result["RETURN_CODE"] = command.returncode
         result["END_TIME"] = self.get_formatted_time("end_time")
 
         # Print the test result for the user
         if command.returncode == 0:
-            print("{:<40} {}".format("[RUNNING TEST]", "PASSED"))
+            print(
+                "{:<30} {:<30} {:<30} {:<30}".format(
+                    self.config_name, self.name, "PASSED", self.config_file
+                )
+            )
         else:
-            print("{:<40} {}".format("[RUNNING TEST]", "FAILED"))
-
-        print("Writing results to " + run_output_file)
+            print(
+                "{:<30} {:<30} {:<30} {:<30}".format(
+                    self.config_name, self.name, "PASSED", self.config_file
+                )
+            )
 
         # Return to starting directory for next test
         os.chdir(self.pwd)
@@ -501,21 +506,16 @@ class BuilderBase:
 
     def show_prepare(self):
         """Print basic run information to the user, if defined, before the run."""
-
-        start_time = self.get_formatted_time("start_time")
+        # return for now, later remove this method.
+        return
 
         print("{:_<80}".format(""))
-        print("{:>40} {}".format("start time:", self.metadata.get("start_time")))
 
         if self.config_file:
             print("{:>40} {}".format("configuration file:", self.config_file))
         print("{:>40} {}".format("testdir:", self.metadata.get("testdir")))
         print("{:>40} {}".format("testpath:", self.metadata.get("testpath")))
         print("{:>40} {}".format("logpath:", self.metadata.get("logfile")))
-        print("{:_<80}".format(""))
-
-        print("\n\n")
-        print("{:<40} {}".format("STAGE", "VALUE"))
         print("{:_<80}".format(""))
 
     @run_wrapper
@@ -558,7 +558,7 @@ class BuilderBase:
             testpath,
             stat.S_IRWXU | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH,
         )
-        print("{:<40} {}".format("[WRITING TEST]", "PASSED"))
+
         return testpath
 
     def _get_test_lines(self):
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -9,17 +9,14 @@ expose functions to run builds.
 """
 
 import datetime
+import logging
 import os
 import re
 import shutil
 import stat
 import sys
 
-from copy import deepcopy
 from jsonschema import validate
-from jsonschema.exceptions import ValidationError
-
-from buildtest.log import init_logfile, init_log
 from buildtest.defaults import (
     BUILDTEST_SHELL,
     build_sections,
@@ -60,10 +57,15 @@ class BuildConfig:
 
            config_file: the pull path to the configuration file, must exist.
         """
+
+        self.logger = logging.getLogger(__name__)
+
         self.recipe = None
 
         # Read the lookup to get schemas available
         self.lookup = get_schemas_available()
+
+        self.logger.debug(f"buildtest found the available schema: {self.lookup} in schema library")
 
         # Load the configuration file, fails on any error
         self.load(config_file)
@@ -77,14 +79,6 @@ class BuildConfig:
            config_file: the pull path to the configuration file, must exist.
         """
         self.config_file = os.path.abspath(config_file)
-
-        if not os.path.exists(self.config_file):
-            sys.exit("Build configuration %s does not exist." % self.config_file)
-
-        elif os.path.isdir(self.config_file):
-            sys.exit(
-                "Please provide a file path (not a directory path) to a build recipe."
-            )
 
         # all test configs must pass global validation (sets self.recipe)
         self._validate_global()
@@ -143,15 +137,14 @@ class BuildConfig:
         """
 
         config_file = config_file or self.config_file
-        outer_schema = load_schema(os.path.join(here, "outer.schema.json"))
+        outer_schema_file = os.path.join(here, "outer.schema.json")
+
+        outer_schema = load_schema(outer_schema_file)
         self.recipe = load_recipe(config_file)
-        try:
-            validate(instance=self.recipe, schema=outer_schema)
-        except ValidationError:
-            sys.exit(
-                "Test configuration %s does not pass validation of the outer schema."
-                % config_file
-            )
+
+        self.logger.debug(f"Validating {config_file} with schema: {outer_schema_file}")
+        validate(instance=self.recipe, schema=outer_schema)
+        self.logger.debug("Validation was successful")
 
     # Builders
 
@@ -230,6 +223,7 @@ class BuilderBase:
         self.testdir = testdir or os.path.join(
             os.getcwd(), ".buildtest", self.config_name
         )
+        self.logger = logging.getLogger(__name__)
 
         # A builder is required to define the type attribute
         if not hasattr(self, "type"):
@@ -251,9 +245,6 @@ class BuilderBase:
                 % (self.type, self.recipe.get("type"))
             )
 
-        # Start with just terminal logger (changes to file on run)
-        # self._init_logger(to_file=False)
-
     def __str__(self):
         return "[builder-%s-%s]" % (self.type, self.name)
 
@@ -274,23 +265,10 @@ class BuilderBase:
         """
         testdir = self._get_testdir()
         create_dir(testdir)
-        for folder in ["run", "log"]:
+        for folder in ["run"]:
             name = "%sdir" % folder
             self.metadata[name] = os.path.join(testdir, folder)
             create_dir(self.metadata[name])
-
-    def _init_logger(self, to_file=True):
-        """Initialize the logger. This is called at the end of prepare_run,
-           so the testpath is defined with the build id.
-        """
-
-        # if to_file:
-        self.metadata["logfile"] = os.path.join(
-            self.metadata["testdir"], "log", "%s.log" % self.build_id
-        )
-        self.logger = init_logfile(self.metadata["logfile"])
-        # else:
-        #    self.logger = init_log()
 
     def get_test_extension(self):
         """Return the test extension, which depends on the shell used. Based
@@ -347,14 +325,12 @@ class BuilderBase:
     def run_wrapper(func):
         """The run wrapper will execute any prepare_run and finish_run 
            sections around some main run function (run or dry_run).
-           A return the result. The show_prepare function
-           log updates to the terminal for the user. The function sets
-           self.result and also returns it to the calling function.
+           A return the result. The function sets self.result and also
+           returns it to the calling function.
         """
 
         def wrapper(self):
             self.prepare_run()
-            self.show_prepare()
             self.result = func(self)
             self.finish_run()
             return self.result
@@ -422,7 +398,6 @@ class BuilderBase:
 
         # Create test directory and run folder they don't exist
         self._create_test_folders()
-        self._init_logger()
         testfile = self._write_test()
         result = self.run_tests(testfile)
         return result
@@ -441,9 +416,12 @@ class BuilderBase:
 
         # Change to the test directory
         os.chdir(self._get_testdir())
+        self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
         # Run the test file using the shell
         cmd = [self.get_shell(), testfile]
+
+        self.logger.debug(f"Running Test via command: {cmd}")
 
         command = BuildTestCommand(cmd)
         out, err = command.execute()
@@ -504,20 +482,6 @@ class BuilderBase:
             timestamp = timestamp.strftime(fmt)
         return timestamp
 
-    def show_prepare(self):
-        """Print basic run information to the user, if defined, before the run."""
-        # return for now, later remove this method.
-        return
-
-        print("{:_<80}".format(""))
-
-        if self.config_file:
-            print("{:>40} {}".format("configuration file:", self.config_file))
-        print("{:>40} {}".format("testdir:", self.metadata.get("testdir")))
-        print("{:>40} {}".format("testpath:", self.metadata.get("testpath")))
-        print("{:>40} {}".format("logpath:", self.metadata.get("logfile")))
-        print("{:_<80}".format(""))
-
     @run_wrapper
     def dry_run(self):
         """Akin to a build preview, we prepare and finish a run, but only
@@ -525,7 +489,6 @@ class BuilderBase:
         """
 
         # Dry run just prints the testing script
-        self._init_logger(to_file=False)
         lines = self._get_test_lines()
         print("\n".join(lines))
 
@@ -545,7 +508,6 @@ class BuilderBase:
         # '$HOME/.buildtest/testdir/<name>/<name>_<timestamp>.sh'
         # This will put output (latest run) in same directory - do we want this?
         testpath = os.path.expandvars(self.metadata["testpath"])
-        testdir = os.path.dirname(testpath)
 
         self.logger.info(f"Opening Test File for Writing: {testpath}")
 
@@ -558,7 +520,7 @@ class BuilderBase:
             testpath,
             stat.S_IRWXU | stat.S_IRGRP | stat.S_IXGRP | stat.S_IROTH | stat.S_IXOTH,
         )
-
+        self.logger.debug(f"Applying permission 755 to {testpath} so that test can be executed")
         return testpath
 
     def _get_test_lines(self):
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -266,8 +266,7 @@ class BuilderBase:
 
            Returns: full path to testing directory
         """
-        testpath = os.path.expandvars(self.metadata["testpath"])
-        return os.path.dirname(testpath)
+        return os.path.dirname(self.metadata["testpath"])
 
     def _create_test_folders(self):
         """Create all needed test folders on init, and add their paths
@@ -384,9 +383,8 @@ class BuilderBase:
             os.path.join(self.testdir, self.name),
             self.get_test_extension(),
         )
-        self.metadata["testdir"] = os.path.dirname(
-            os.path.expandvars(self.metadata["testpath"])
-        )
+        self.metadata["testpath"] = os.path.expandvars(self.metadata["testpath"])
+        self.metadata["testdir"] = os.path.dirname(self.metadata["testpath"])
 
         # The start time to print for the user
         self.metadata["start_time"] = datetime.datetime.now()
@@ -515,7 +513,7 @@ class BuilderBase:
 
         # '$HOME/.buildtest/testdir/<name>/<name>_<timestamp>.sh'
         # This will put output (latest run) in same directory - do we want this?
-        testpath = os.path.expandvars(self.metadata["testpath"])
+        testpath = self.metadata["testpath"]
 
         self.logger.info(f"Opening Test File for Writing: {testpath}")
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -442,15 +442,12 @@ class BuilderBase:
         # Keep an output file
         run_output_file = os.path.join(self.metadata.get("rundir"), self.build_id)
 
-        # Keep an error file
-        run_error_file = os.path.join(self.metadata.get("rundir"), self.build_id)
-
         # write output of test to .out file
         with open(run_output_file + ".out", "w") as fd:
             fd.write("\n".join(out))
 
         # write error from test to .err file
-        with open(run_error_file + ".err", "w") as fd:
+        with open(run_output_file + ".err", "w") as fd:
             fd.write("\n".join(err))
 
         result["RETURN_CODE"] = command.returncode
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -82,6 +82,14 @@ class BuildConfig:
         """
         self.config_file = os.path.abspath(config_file)
 
+        if not os.path.exists(self.config_file):
+            sys.exit("Build configuration %s does not exist." % self.config_file)
+
+        elif os.path.isdir(self.config_file):
+            sys.exit(
+                "Please provide a file path (not a directory path) to a build recipe."
+            )
+
         # all test configs must pass global validation (sets self.recipe)
         self._validate_global()
 
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -103,7 +103,9 @@ class BuildExecutor:
         # Run each step defined for dry run
         for step in executor.steps:
             if getattr(executor, step, None):
-                executor.builder.logger.debug("Running %s for executor %s" % (step, executor))
+                executor.builder.logger.debug(
+                    "Running %s for executor %s" % (step, executor)
+                )
                 getattr(executor, step)()
         return executor.result
 
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -101,9 +101,7 @@ class BuildExecutor:
         # Run each step defined for dry run
         for step in executor.steps:
             if getattr(executor, step, None):
-                executor.builder.logger.debug(
-                    "Running %s for executor %s" % (step, executor)
-                )
+                # executor.builder.logger.debug("Running %s for executor %s" % (step, executor))
                 getattr(executor, step)()
         return executor.result
 
@@ -170,7 +168,7 @@ class BaseExecutor:
         """Setup the executor, meaning we check that the builder is defined,
            the only step needed for a local (base) executor.
         """
-        print(self.builder)
+
         if not self.builder:
             sys.exit("Builder is not defined for executor.")
 
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -103,7 +103,7 @@ class BuildExecutor:
         # Run each step defined for dry run
         for step in executor.steps:
             if getattr(executor, step, None):
-                # executor.builder.logger.debug("Running %s for executor %s" % (step, executor))
+                executor.builder.logger.debug("Running %s for executor %s" % (step, executor))
                 getattr(executor, step)()
         return executor.result
 
--- a/buildtest/executors/base.py
+++ b/buildtest/executors/base.py
@@ -3,7 +3,7 @@ BuildExecutor: manager for test executors
 Copyright (C) 2020 Vanessa Sochat. 
 """
 
-import os
+import logging
 import sys
 
 
@@ -28,9 +28,11 @@ class BuildExecutor:
            :type default: str
         """
         self.executors = {}
-
+        self.logger = logging.getLogger(__name__)
+        self.logger.debug("Getting Executors from buildtest settings")
         # Load the executors
         for name, executor in config_opts.get("executors", {}).items():
+            self.logger.debug(f"Executor Name: {name}  Executor Value: {executor}")
             if executor["type"] == "local":
                 self.executors[name] = LocalExecutor(name, executor)
             elif executor["type"] == "slurm":
--- a/buildtest/system.py
+++ b/buildtest/system.py
@@ -67,12 +67,12 @@ class BuildTestSystem:
             f"Operating System Build Number: {self.system['os']['build_number']}"
         )
 
-        self.logger.info("Session Environment Variables")
-        self.logger.info("{:_<80}".format(""))
+        self.logger.debug("Session Environment Variables")
+        self.logger.debug("{:_<80}".format(""))
         for k, v in self.system["env"].items():
-            self.logger.info(f"{k}: {v}")
-        self.logger.info("{:_<80}".format(""))
-        self.logger.info(f"Python Path: {self.system['python']}")
+            self.logger.debug(f"{k}: {v}")
+        self.logger.debug("{:_<80}".format(""))
+        self.logger.debug(f"Python Path: {self.system['python']}")
 
     def check_lmod(self):
         """Check if the system has Lmod installed, determine by setting
@@ -82,7 +82,7 @@ class BuildTestSystem:
         self.lmod = "LMOD_DIR" in os.environ and os.path.exists(
             os.environ.get("LMOD_DIR", "")
         )
-        self.logger.info(f"LMOD_DIR: {self.lmod}")
+        self.logger.debug(f"LMOD_DIR: {self.lmod}")
 
     def check_scheduler(self):
         """Check for batch scheduler. Currently checks for LSF or SLURM by running
--- a/buildtest/system.py
+++ b/buildtest/system.py
@@ -53,24 +53,24 @@ class BuildTestSystem:
         self.system["os"]["minor_version"] = distro.major_version(best=True)
         self.system["os"]["build_number"] = distro.build_number(best=True)
         self.system["env"] = dict(os.environ)
-        self.system["python"] = shutil.which("python") or None
+        self.system["python"] = shutil.which("python")
 
-        self.logger.info(f"Operating System: {self.system['os']['name']}")
-        self.logger.info(f"Operating System Version: {self.system['os']['version']}")
-        self.logger.info(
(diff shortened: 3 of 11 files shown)


buildtest/buildsystem/base.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, name, recipe_config, config_file=None, testdir=None):
             )
 
         # Start with just terminal logger (changes to file on run)
-        self._init_logger(to_file=False)
+        # self._init_logger(to_file=False)
```
vsoch: @shahzebsiddiqui this will break other tests because this first logger is _just_ to the terminal, the one set later is to file. The logger needs to be present at this stage in the build. If you change one logger (to file) you just need to change the other logger (to the terminal) to do the same. Commenting this out will break things.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: Yea i kinda realized this and wanted to get rid of the log output from stream shown in my first message in PR. I noticed the logger was passed from BuilderBase to BaseExecutor. There were few issues related to logging in #265 which i wanted to address later.

This PR was meant to address #256

shahzebsiddiqui: I don't think it makes sense to have a log stream on dry run. We are doign it for time being https://github.com/shahzebsiddiqui/buildtest-framework/blob/66e49c9edaba674d06442ebfed94d9fd549c41af/buildtest/buildsystem/base.py#L528

vsoch:
I think before this goes in, the logging needs to be refactored to be outside of the builder (and inherited from a higher level?) Once that is done, then it makes sense to update the format here (which btw is very beautiful! Much improved!)

shahzebsiddiqui:
> I think before this goes in, the logging needs to be refactored to be outside of the builder (and inherited from a higher level?) Once that is done, then it makes sense to update the format here (which btw is very beautiful! Much improved!)

Okay let's discuss in issue #265 i have some ideas and i think the change will require some more modification to get logging working correctly. If you want that to be part of this PR, I can create a feature branch and PR in my repo for review. Once it's approved it will be in this PR and we can merge it. Anyhow i have updated the message right now this PR is addressing two issues #256  #264.

shahzebsiddiqui:
@vsoch okay i made some significant changes without affecting this PR i have issued a PR https://github.com/shahzebsiddiqui/buildtest-framework/pull/7 in my fork but not sure if it's possible for you to review, once merged it should update this PR and address the issues you raised up.

vsoch:
Already done @shahzebsiddiqui , when you ping me the other PR I get an email notification, so now I have two :)

shahzebsiddiqui:
> Already done @shahzebsiddiqui , when you ping me the other PR I get an email notification, so now I have two :)

I figured, the @ mention works but wanted to confirm since it was my fork and couldnt assign you as reviewer.

vsoch:
yes you can assume from here on out that any mention with @, or any repo I'm watching (like this one) I'll get an email.

buildtest/buildsystem/base.py
```diff
@@ -78,14 +82,6 @@ def load(self, config_file):
         """
         self.config_file = os.path.abspath(config_file)
 
-        if not os.path.exists(self.config_file):
```
vsoch: Is this check still done in self._validate_global()?

buildtest/buildsystem/base.py
```diff
@@ -274,24 +267,11 @@ def _create_test_folders(self):
         """
         testdir = self._get_testdir()
         create_dir(testdir)
-        for folder in ["run", "log"]:
+        for folder in ["run"]:
             name = "%sdir" % folder
```
vsoch, line 278: okay so just to clarify - this "log" folder is removed with preference for putting the buildtest.log directly in the test output folder.

buildtest/buildsystem/base.py
```diff
@@ -441,9 +418,13 @@ def run_tests(self, testfile):
 
         # Change to the test directory
         os.chdir(self._get_testdir())
```
vsoch, line 428: Is this does any ping to the file system to check for existence, it should be called once and then used. E.g.,
```python
testdir = self._get_testdir()
```
instead of being called multiple times to return the same folder.

buildtest/buildsystem/base.py
```diff
@@ -455,31 +436,35 @@ def run_tests(self, testfile):
             self.metadata.get("rundir"), "%s.out" % self.build_id
         )
 
-        # Run the test file, print output to file
-        with open(run_output_file, "w") as fd:
+        # Keep an error file
+        run_error_file = os.path.join(
```
vsoch: For this, it would be more elegant to create a run_output file:

```python
run_output_file = os.path.join(	
            self.metadata.get("rundir"), self.build_id
        )
```
And then using the basename for each of .out and .err:

```python
with open(run_output_file + ".err")
   # write error

with open(run_output_file + ".out")
   # write output
```
I don't see from these changes where run_output_file and run_error_file are coming from, but that could be adopted to the above.

buildtest/buildsystem/base.py
```diff
@@ -455,31 +436,35 @@ def run_tests(self, testfile):
...
+            print(
+                "{:<30} {:<30} {:<30} {:<30}".format(
+                    self.config_name, self.name, "PASSED", self.config_file
+                )
+            )
         else:
-            print("{:<40} {}".format("[RUNNING TEST]", "FAILED"))
-
-        print("Writing results to " + run_output_file)
+            print(
```
vsoch, line 464: Should this be FAILED if the return code isn't 0?

buildtest/executors/base.py
```diff
@@ -101,9 +103,7 @@ def run(self, builder):
         # Run each step defined for dry run
         for step in executor.steps:
             if getattr(executor, step, None):
-                executor.builder.logger.debug(
-                    "Running %s for executor %s" % (step, executor)
-                )
+                # executor.builder.logger.debug("Running %s for executor %s" % (step, executor))
```
vsoch: Should this be commented out?

buildtest/system.py
```diff
@@ -23,20 +24,30 @@ def __init__(self):
...
         """Based on the module "distro" set the linux distrubution name and version
         """
-        self.system["OS_NAME"] = distro.linux_distribution(
+        self.system["os_name"] = distro.linux_distribution(
             full_distribution_name=False
         )[0]
-        self.system["OS_VERSION"] = distro.linux_distribution(
-            full_distribution_name=False
-        )[1]
+        self.system["os_ver"] = distro.linux_distribution(full_distribution_name=False)[
```
vsoch: What is the namespace for these keys? E.g., why are we using "os_ver" which is sort of short /. choppy?

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: this is not needed in ``self._validate_global()`` because it is done earlier in the ``discover_configs`` stage in particular at https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/menu/build.py#L45 you will notice that we call discover_configs before calling BuildConfig so no point in doing these checks again

https://github.com/HPC-buildtest/buildtest-framework/blob/abcfad94b69598f87f15c58e3c0f0762d9c9a99f/buildtest/menu/build.py#L87

buildtest/buildsystem/base.py
shahzebsiddiqui: Yes there is no log file generated in the testdir and its now just called ``buildtest.log`` in current  directory. Note for now its called ``buildtest.log`` i did this purposely so user can reference one file, it may be useful to remove the file before writing as for now its appending the content.

buildtest/buildsystem/base.py
shahzebsiddiqui: as of now (``devel`` HEAD) its only being called twice. See https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/base.py 

It could be  useful to just retrieve via ``self.metadata.get("testdir")`` 

My only concern is I dont think we have os.path.expandvars set such that a test directory can be set using variable expansion.  This was done correctly in 

https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/base.py#L268

We should probably have test to validate something like ``buildtest build --testdir $HOME`` and see if it writes test correctly.

buildtest/buildsystem/base.py
shahzebsiddiqui: This makes sense i can refactor this stay tuned, it came to my mind also but didn't do it for first implementation.

buildtest/buildsystem/base.py
vsoch: Okay great! Just wanted to double check.

buildtest/buildsystem/base.py
vsoch: Perfect.

buildtest/buildsystem/base.py
vsoch: I think we should maybe aim to call this once, and store for the duration of the test? These sorts of setup variables we wouldn't want to weirdly change (for some buggy reason) in the middle of a test.

buildtest/buildsystem/base.py
vsoch: Sounds good!

buildtest/buildsystem/base.py
shahzebsiddiqui: good catch that was typo on my end due to copy-paste.

buildtest/system.py
shahzebsiddiqui: good point right now these are only used for logging, there was some additional work in this PR https://github.com/HPC-buildtest/buildtest-framework/pull/236/files#diff-da5c353202a52de72277768aebc926b6R35-R44 which i kept open but won't get merged. Idea was to use BuildTestSystem to detect system detail some of which is auto-detected.

As per ``os_ver`` and ``os_name`` i was trying to do a few things
1. Standardize all keys to be lower case and multi-words separated via underscore _
2. Capture OS details as part of logging, could be used for logging or reporting as part of run results.
3. Before buildtest was doing sanity check of binaries by system packages. Initially it supported rpm so i had a requirement for buildtest to only support Centos/Fedora/RHEL based system. That feature is dropped so there is no real purpose of having os_name and os_ver other than logging. The feature was called ``buildtest build --package <package-name>``. Its been dropped long ago.

buildtest/executors/base.py
shahzebsiddiqui: Ok i fixed this locally and looked at logfile and i am getting the message 

```
2020-04-07 13:19:49,169 [base.py:106 -   run() ] - [DEBUG] Running setup for executor [executor-local-runlocal]
2020-04-07 13:19:49,169 [base.py:106 -   run() ] - [DEBUG] Running run for executor [executor-local-runlocal]

```

buildtest/buildsystem/base.py
shahzebsiddiqui: i agree and if i understand the only reason why we are ``chdir`` before running test is to ensure any generated content from the test that may include compilation, executable, or output files get stored in test directory and not in current directory which would be ugly. Unrelated to this, i noticed that we don't remove test directory before writing test. This is an issue when some test fail upon re-execution since they assume a clean directory. 

This is true for example in ``examples/script/bzip2.yml``. The point is should test directory be clean to avoid rewriting content of generated file.

buildtest/system.py
vsoch: Is there any reason to truncate the name? I like `os_name`, but would it make sense to do `os_version` so it's explicitly clear?

buildtest/executors/base.py
vsoch: Cool! It might be possible to simplify that, but I'm okay with it for now :)

buildtest/buildsystem/base.py
vsoch: I always get worried whenever scripts take charge to delete content - for example what if there was a bug with testdir and you shutil.rmtree something important? I think it makes sense to clean up the test directory on re-run, but is there some marker we can use to be absolutely sure that it's a test directory? Perhaps the output file or the folder structure? We haven't talked much about the structure of a test directory, but whatever that might encompass, this could be a good check to do.

buildtest/system.py
shahzebsiddiqui: ill make the change to ``os_version`` makes sense. Then no truncation for time being. If you see in the other PR https://github.com/HPC-buildtest/buildtest-framework/pull/236/files#diff-da5c353202a52de72277768aebc926b6R39-R41 i can go a bit more elegant such that we can 
reference something like 

```
self.system['os']['name']
self.system['os']['version']
self.system['os']['major']
self.system['os']['major']
self.system['os']['build_number']
```

Again we can go crazy with this and would be fine with making any change that makes sense. Let me know your thoughts or a proposal

vsoch: What you showed above is lovely! No worries about adding it here, as you mentioned it can be a separate PR.

buildtest/buildsystem/base.py
shahzebsiddiqui: The other alternate is stick a build id in the sub-directory path to avoid rewriting to same directory, that's what i did before to avoid this issue and worked nice. Previously, buildtest would purge and write again, then this was avoided by added a unique sub-directory that is incremented. Every build was then written to ``build.json`` and there was a method to retrieve total build IDs so that buildtest knows what the next build ID is. 
Adding this complete feature would require a bit more refactoring and it could be its own PR. Right now everything is working as expected.

Another alternate to an ID is use a random hex (32bit) lets say to stick as build ID, later on buildtest would need to capture build ID in some json file to retrieve test reports. Again that's out of scope for another PR and feature request.

vsoch: I agree it's out of scope - I think it would be cleanest to remove the directory, but have some check that it's a build test directory, and maybe a user flag. I don't see it being useful to run multiple builds and have the folders build up (it was really messy last time) but if this approach is wanted, it would work to do a random hash each time (and make that hash very clear to the user in the screen logs). I think this is what I've seen other similar tools, either for workflows or test runs, do.

buildtest/system.py
shahzebsiddiqui: Okay i made some changes in last commit just now to add a few extra details. Here is a snipet of the log, i think adding the environment section can be useful in troubleshooting.


```
(buildtest-framework) ssi29@ag-mxg-hulk090> head -n 40 buildtest.log
2020-04-07 15:04:54,124 [main.py:19 -  main() ] - [INFO] Starting buildtest log
2020-04-07 15:04:54,124 [system.py:28 - __init__() ] - [DEBUG] Starting System Compatibility Check
2020-04-07 15:04:54,196 [system.py:58 - init_system() ] - [INFO] Operating System: rhel
2020-04-07 15:04:54,197 [system.py:59 - init_system() ] - [INFO] Operating System Version: 7.6
2020-04-07 15:04:54,197 [system.py:60 - init_system() ] - [INFO] Operating System Major Version: 7
2020-04-07 15:04:54,197 [system.py:61 - init_system() ] - [INFO] Operating System Minor Version: 7
2020-04-07 15:04:54,197 [system.py:62 - init_system() ] - [INFO] Operating System Build Number:
2020-04-07 15:04:54,197 [system.py:64 - init_system() ] - [INFO] Session Environment Variables
2020-04-07 15:04:54,197 [system.py:65 - init_system() ] - [INFO] ________________________________________________________________________________
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] REMOTEHOST: lt4-th3-dsa.dsone.3ds.com
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] CABAQUS: /usr/abaqus60/udev/source
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] EBVERSIONZLIB: 1.2.11
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] MANPATH: /mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/man:/mxg-hpc/users/ssi29/easybuild/software/libffi/3.2.1-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/XZ/5.2.4-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/SQLite/3.29.0-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/Tcl/8.6.9-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/Tcl/8.6.9-GCCcore-8.3.0/man:/mxg-hpc/users/ssi29/easybuild/software/libreadline/8.0-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/ncurses/6.1-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/zlib/1.2.11-GCCcore-8.3.0/share/man:/mxg-hpc/users/ssi29/easybuild/software/bzip2/1.0.8-GCCcore-8.3.0/man:/mxg-hpc/users/ssi29/easybuild/software/GCCcore/8.3.0/share/man:/r/cluster_software/pub_lsf/10.1/man:/usr/local/man:/usr/share/man:/usr/man:/usr/X11R6/man:/usr/openwin/man:/usr/share/man/allman:/usr/abaqus60/udev/generated/man:/usr/abaqus60/udev/object/lnx86_64/External/graphviz/man
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] FIGS: /p/marketing/figures
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] EBVERSIONPYTHON: 3.7.4
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] XDG_SESSION_ID: 144609
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] CABAQUS_O: /usr/abaqus60/udev/object/lnx86_64
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] HOSTNAME: ag-mxg-hulk090
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] EBDEVELPYTHON: /mxg-hpc/users/ssi29/easybuild/software/Python/3.7.4-GCCcore-8.3.0/easybuild/Python-3.7.4-GCCcore-8.3.0-easybuild-devel
2020-04-07 15:04:54,197 [system.py:67 - init_system() ] - [INFO] _ModuleTable003_: Zy1ocGMvdXNlcnMvc3NpMjkvZWFzeWJ1aWxkL21vZHVsZXMvYWxsL1NRTGl0ZS8zLjI5LjAtR0NDY29yZS04LjMuMC5sdWEiLFsiZnVsbE5hbWUiXT0iU1FMaXRlLzMuMjkuMC1HQ0Njb3JlLTguMy4wIixbImxvYWRPcmRlciJdPTcscHJvcFQ9e30sWyJzdGFja0RlcHRoIl09MSxbInN0YXR1cyJdPSJhY3RpdmUiLFsidXNlck5hbWUiXT0iU1FMaXRlLzMuMjkuMC1HQ0Njb3JlLTguMy4wIix9LFRjbD17WyJmbiJdPSIvbXhnLWhwYy91c2Vycy9zc2kyOS9lYXN5YnVpbGQvbW9kdWxlcy9hbGwvVGNsLzguNi45LUdDQ2NvcmUtOC4zLjAubHVhIixbImZ1bGxOYW1lIl09IlRjbC84LjYuOS1HQ0Njb3JlLTguMy4wIixbImxvYWRPcmRlciJdPTYscHJvcFQ9e30sWyJzdGFja0RlcHRo

```

I think in future it would help capture paths to launchers (srun, sbatch, mpirun, mpiexec) I had some ideas in buildtest setting to allow user to specify path to launcher paths let's say in case of mpirun you or other mpi specific launcher you dont want to rely what's in $PATH as it can change due to shell or changes to module so to ensure consistency it would be useful to let user specify a full path to launcher. Again out of scope for this task.  

For now i think it's safe to say this issue is resolved. I changed ``[os_ver]`` to ``['os']['version']``

vsoch: Great! One final tweak - all of those variables should be for level DEBUG and not INFO.

buildtest/buildsystem/base.py
shahzebsiddiqui: Okay so there is an issue with Travis see https://travis-ci.com/github/HPC-buildtest/buildtest-framework/jobs/316447547 and it is related to how regtest was written since ``discover_config`` is not used in test before invoking BuildConfig

https://github.com/shahzebsiddiqui/buildtest-framework/blob/change_output_format/tests/buildsystem/test_base.py#L22

I guess either we refactor regtest or i undo my change. What your thoughts?

buildtest/buildsystem/base.py
shahzebsiddiqui: Just FYI. No log file in test directory now

```
(buildtest-framework) ssi29@ag-mxg-hulk090> tree .buildtest
.buildtest
├── python-circle
│   ├── circle_area.py
│   └── run
│       ├── circle_area_04-07-2020-13-19.err
│       ├── circle_area_04-07-2020-13-19.out
│       ├── circle_area_04-07-2020-13-26.err
│       ├── circle_area_04-07-2020-13-26.out
│       ├── circle_area_04-07-2020-13-27.err
│       └── circle_area_04-07-2020-13-27.out
└── slurm
    ├── run
    │   ├── slurm_down_nodes_reason_04-07-2020-13-17.err
    │   ├── slurm_down_nodes_reason_04-07-2020-13-17.out
    │   ├── slurm_down_nodes_reason_04-07-2020-14-59.err
    │   ├── slurm_down_nodes_reason_04-07-2020-14-59.out
    │   ├── slurm_down_nodes_reason_04-07-2020-15-02.err
    │   ├── slurm_down_nodes_reason_04-07-2020-15-02.out
    │   ├── slurm_down_nodes_reason_04-07-2020-15-04.err
    │   ├── slurm_down_nodes_reason_04-07-2020-15-04.out
    │   ├── slurm_not_responding_nodes_04-07-2020-13-17.err
    │   ├── slurm_not_responding_nodes_04-07-2020-13-17.out
    │   ├── slurm_not_responding_nodes_04-07-2020-14-59.err
    │   ├── slurm_not_responding_nodes_04-07-2020-14-59.out
    │   ├── slurm_not_responding_nodes_04-07-2020-15-02.err
    │   ├── slurm_not_responding_nodes_04-07-2020-15-02.out
    │   ├── slurm_not_responding_nodes_04-07-2020-15-04.err
    │   └── slurm_not_responding_nodes_04-07-2020-15-04.out
    ├── slurm_down_nodes_reason.sh
    └── slurm_not_responding_nodes.sh

4 directories, 25 files

```

buildtest/buildsystem/base.py
vsoch: The discover_config, in that it finds and validates a config file, must of course happen before you try to load it.

buildtest/buildsystem/base.py
shahzebsiddiqui: I'll try to open a separate issue as this was related to Stampede test i was running. Some tests couldn't rerun after first time because directory existed. I think the value of reruns would be run test for reproducible results .

One good use-case is to run acceptance test once a week, and maybe performance test once a month. Right now all the tests for stampede2 are in one directory so i can just do ``buildtest build -c tests`` and it will run everything. Or i can go by sub-directory. But eventually i'd like to have this as repeated operation so buildtest can support multiple reruns cleanly so reports can be generated based on a build ID with timestamp on when it was run. 

I think the framework should support a way to gather data from test reports which users can understand.  Running ad-hoc commands once would not be useful unless all the data is results are captured somewhere that buildtest can query and report.

buildtest/system.py
shahzebsiddiqui: makes sense i will make that change, should i make all of ``self.system`` be DEBUG to be consistent?

shahzebsiddiqui: ill make it DEBUG for all them to be consistent.

vsoch: yes.

buildtest/buildsystem/base.py
shahzebsiddiqui: Okay so i found that culprit. If i remove the lines https://github.com/shahzebsiddiqui/buildtest-framework/blob/change_output_format/tests/buildsystem/test_base.py#L21-22 then test pass. 

I see this test would be there when no config is passed and I think what you had in mind was if user runs ``buildtest build`` it should try to search from ``~/.buildtest/site`` i.e TESTROOT, but thing is we have a check at end if no config files found that it will exit which is here https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/menu/build.py#L60

So i think it's not possible to have ``BuildConfig("")`` as a call since discover_config will catch all errors. Hope that makes sense. I ran test locally when commenting out line 21,22 and test was working. I was able to reproduce same error in Travis when the lines were uncommented.

So i'd recommend we should remove these lines.  I think there should be test for method ``discover_config`` which should catch all issues related to configuration, that way checks are done in advance before task is delegated to other classes. 
What do you think?

I think

shahzebsiddiqui: Also looking at codecov https://codecov.io/gh/HPC-buildtest/buildtest-framework/src/devel/buildtest/menu/build.py we dont have any test for ``discover_config`` and i think some regtest should be written to catch all types of input for configuration (file, directory, or path relative to ``.site``). That way the test are modular per method as pose to one or two monolothic test such as one in ``buildsystem/test_base.py``

shahzebsiddiqui: Okay travis is now  working with latest fix i proposed. I think this is the last remaining issue. I updated scope of PR as this will address three different issues. Let me know if this fix is acceptable and we can come to resolution.

vsoch: > Okay so i found that culprit. If i remove the lines https://github.com/shahzebsiddiqui/buildtest-framework/blob/change_output_format/tests/buildsystem/test_base.py#L21-22 then test pass.

This isn't a culprit - if the user doesn't provide a valid path the test should fail with systemExit, which was working before and is exactly what that test is doing. What seems to be the issue is that discover_config is not SystemExit-ing as it should. I think the lines should be kept, and the discover_config should system exit as it was originally done.

shahzebsiddiqui: Okay so would you want me to undo the change on line 81  and revert my regtest so that ``BuildTestConfig("")`` can be invoked which should raise SystemExit. 

```
if not os.path.exists(self.config_file):
```

vsoch: yes, removing the check that the config file exists isn't logical. The lines that I added for both a file, and then a directory, should not be removed.

shahzebsiddiqui: @vsoch i undid my changes and added test to validate ``discover_config`` See latest commit. This increases coverage of ``buildtest/menu/build.py`` which was previously at 20% now at 49%. See https://codecov.io/gh/HPC-buildtest/buildtest-framework/tree/devel/buildtest/menu

buildtest/buildsystem/base.py
```diff
@@ -545,7 +516,6 @@ def _write_test(self, lines=None):
         # '$HOME/.buildtest/testdir/<name>/<name>_<timestamp>.sh'
         # This will put output (latest run) in same directory - do we want this?
         testpath = os.path.expandvars(self.metadata["testpath"])
```
vsoch: Should we save this with the builder so it doesn't need to be run (with expandvars) more than once? It should be done at the beginning of run/dry_run or something shared between them before that.

buildtest/buildsystem/base.py
shahzebsiddiqui: I think you can see the coverage in coveralls here https://coveralls.io/builds/29955036

buildtest/system.py
```diff
@@ -23,20 +24,55 @@ def __init__(self):
...
         )[0]
-        self.system["OS_VERSION"] = distro.linux_distribution(
+        self.system["os"]["version"] = distro.linux_distribution(
             full_distribution_name=False
         )[1]
+        self.system["os"]["major_version"] = distro.major_version(best=True)
+        self.system["os"]["minor_version"] = distro.major_version(best=True)
+        self.system["os"]["build_number"] = distro.build_number(best=True)
+        self.system["env"] = dict(os.environ)
+        self.system["python"] = shutil.which("python") or None
```
vsoch: I think if shutil.which() can't find something it already returns None.

```python
import shutil                                                           
res = shutil.which('nope')                                              
res == None                                                             
True
```
so the "or None" is not necessary.

buildtest/buildsystem/base.py
shahzebsiddiqui: hmm, that does make sense shall i overwrite the variable ``self.metadata["testpath"]``

buildtest/system.py
```diff
@@ -23,20 +24,55 @@ def __init__(self):
...
+
+        self.logger.info(f"Operating System: {self.system['os']['name']}")
+        self.logger.info(f"Operating System Version: {self.system['os']['version']}")
+        self.logger.info(
+            f"Operating System Major Version: {self.system['os']['major_version']}"
```
vsoch: Is there any reason to not assemble a complete version string and have a single, clean message for OS version over printing a line for each?

buildtest/system.py
```diff
@@ -23,20 +24,55 @@ def __init__(self):
...
+        self.logger.info(
+            f"Operating System Minor Version: {self.system['os']['minor_version']}"
+        )
+        self.logger.info(
+            f"Operating System Build Number: {self.system['os']['build_number']}"
+        )
+
+        self.logger.debug("Session Environment Variables")
+        self.logger.debug("{:_<80}".format(""))
+        for k, v in self.system["env"].items():
```
vsoch: This is actually a security risk - we should not be printing environment variables to a log file. If absolutely necessary, just print the keys.

vsoch (changes requested):

buildtest/buildsystem/base.py
vsoch: That would be easy enough to do - the reason to have the expansion is to honor some custom environment variable definition, but once the build has started the user can consider that "frozen" so I think updating it (and using that full path for the rest of tests) makes most sense.

buildtest/system.py
vsoch: Also now we need to be very careful for what is done with:
```python
self.system["env"] = dict(os.environ)
```
because what goes in the environment? Secrets, of course.

buildtest/system.py
vsoch: I would suggest namespacing here - for buildtest envars, we prefix with BUILDTEST and then we at least know the ones that are safe to skip / print, and we can filter os.environ down to anything that starts with `BUILDTEST` or another white listed set.

buildtest/system.py
shahzebsiddiqui: i fixed the shutil issue. 

Oh yea i know if i figure capturing environment shell can be useful for debugging for both end-users and maintainers

buildtest/system.py
shahzebsiddiqui: Yea i used to have buildtest variables prefixed with BUILDTEST that's not the case, if you want i can just print the keys without the value.

vsoch: Ok let’s open an issue for the environment namespace, other than filtering we don’t want a conflict of envars with some other tool that happens to use a variable with the same name.

buildtest/system.py
shahzebsiddiqui: no particular reason, just providing extra detail. I mean this is sufficient also at the moment it not providing any extra value other that additional detail. 

```
 distro.linux_distribution()
('Red Hat Enterprise Linux Server', '7.6', 'Maipo')
```

buildtest/buildsystem/base.py
shahzebsiddiqui: okay just to confirm with latest change it should be working. A test for variable expansion would be something like 

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/script/slurm.yml --testdir $PWD/dummy
Config Name                    SubTest                        Status                         Config Path
slurm                          slurm_down_nodes_reason        PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
                         Test summary
Executed 2 tests
Passed Tests: 0 Percentage: 0.0%
Failed Tests: 2 Percentage: 100.0%
(buildtest-framework) ssi29@ag-mxg-hulk090> tree dummy/
dummy/
├── run
│   ├── slurm_down_nodes_reason_04-08-2020-21-25.err
│   ├── slurm_down_nodes_reason_04-08-2020-21-25.out
│   ├── slurm_not_responding_nodes_04-08-2020-21-25.err
│   └── slurm_not_responding_nodes_04-08-2020-21-25.out
├── slurm_down_nodes_reason.sh
└── slurm_not_responding_nodes.sh

1 directory, 6 files

```

buildtest/system.py
shahzebsiddiqui: okay for now i print the keys and its fixed in the logs.

Here is a snippet

```
2020-04-08 21:25:52,375 [system.py:70 - init_system() ] - [DEBUG] Session Environment Variables
2020-04-08 21:25:52,375 [system.py:71 - init_system() ] - [DEBUG] ________________________________________________________________________________
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] REMOTEHOST
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] CABAQUS
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] EBVERSIONZLIB
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] MANPATH
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] FIGS
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] EBVERSIONPYTHON
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] XDG_SESSION_ID
2020-04-08 21:25:52,375 [system.py:73 - init_system() ] - [DEBUG] CABAQUS_O

```

buildtest/buildsystem/base.py
vsoch: Why does it say there are 2 percentage failed tests (then 100%) but report them above as PASSED?

vsoch: For the first part I understand it’s a count followed by “Percentage 100%” so it’s just a confusing read for the user. Can we do:

2/2: 100% 

Or something simple like that!

buildtest/system.py
vsoch: Yes let’s do that, it’s more human readable as a single line. If you want to join with a space to avoid the extra layer of data structure that works too.

buildtest/buildsystem/base.py
shahzebsiddiqui: Let me look into this, good catch i didn't pay attention hold on

shahzebsiddiqui: okay this is fixed now i had typo again for some reason i was reporting test as PASSED

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/script/slurm.yml
Config Name                    SubTest                        Status                         Config Path
slurm                          slurm_down_nodes_reason        FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
                         Test summary
Executed 2 tests
Passed Tests: 0/2 Percentage: 0.0%
Failed Tests: 2/2 Percentage: 100.0%

(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/script/python-circle.yml
Config Name                    SubTest                        Status                         Config Path
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
                         Test summary
Executed 1 tests
Passed Tests: 1/1 Percentage: 100.0%
Failed Tests: 0/1 Percentage: 0.0%

```

buildtest/system.py
shahzebsiddiqui: okay this is fixed now it looks like this now

```
(buildtest-framework) ssi29@ag-mxg-hulk090> cat buildtest.log | grep Operating
2020-04-08 22:00:23,242 [system.py:49 - init_system() ] - [DEBUG] Operating System: Red Hat Enterprise Linux Server 7.6 Maipo

```

buildtest/buildsystem/base.py
vsoch: Looks good!

buildtest/system.py
vsoch: Much improved.

buildtest/buildsystem/base.py
```diff
@@ -441,45 +426,49 @@ def run_tests(self, testfile):
...
 
         # Keep an output file
-        run_output_file = os.path.join(
-            self.metadata.get("rundir"), "%s.out" % self.build_id
-        )
+        run_output_file = os.path.join(self.metadata.get("rundir"), self.build_id)
 
-        # Run the test file, print output to file
-        with open(run_output_file, "w") as fd:
+        # Keep an error file
```
vsoch: Having run_output_file and run_error_file is redundant, you can remove the second and then add an extension for err or out to the own variable.

tests/buildsystem/schemas/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+@pytest.mark.xfail(
+    reason="Invalid File Extension for loading schema", raises=SystemExit
+)
+def test_load_schema_invalid_ext():
+
+    # invalid file extension should fail
+    load_schema(os.path.join(root, "README.rst"))
```
vsoch, line 16: One thing we will need to think about for another PR is how to handle skipping over yaml files that are discovers that aren’t intended for buildtest.

buildtest/buildsystem/base.py
vsoch: Was this added back, so dry run has stdout logging? Just want to verify before resolving.

buildtest/buildsystem/base.py
shahzebsiddiqui: thats quite odd, i recall making this change but for some reason this was not made something must have happened. Let me fix this quick

tests/buildsystem/schemas/test_utils.py
shahzebsiddiqui: indeed good discussion for another PR :)

buildtest/buildsystem/base.py
shahzebsiddiqui: This is fixed now

vsoch (approved):
Great work @shahzebsiddiqui ! I think this is ready for merge :)

shahzebsiddiqui:
wohoo! thanks @vsoch. Thanks for being a good reviewer! Off to bed now

PR #268, opened 2020-04-09
exclude config option

### LINKED ISSUE #267: exclude configs from being processed in a directory.
Okay so i know this is going to be needed when i run stampede2 test i can just do ``buildtest build -c tests`` which will run all these tests

```
tests/
├── apps
│   ├── openmp
│   │   ├── hello.yml
│   │   └── src
│   │       └── omp_hello.c
│   └── serial
│       ├── hello.yml
│       └── src
│           ├── hello.c
│           ├── hello.cpp
│           └── hello.f90
├── sched
│   ├── failed_jobs.yml
│   ├── jobs.yml
│   └── partitions.yml
├── system
│   ├── account_alias.yml
│   ├── bzip2_compression.yml
│   ├── check_login.yml
│   ├── dns.yml
│   ├── env.yml
│   ├── idev.yml
│   ├── kernel.yml
│   ├── module.yml
│   ├── mount.yml
│   ├── selinux.yml
│   ├── shared_fs.yml
│   ├── systemd.yml
│   └── timezone.yml
└── utils
    └── tacc_help.yml

```

There are some test that would get scheduled to scheduler that may take a long time, there is no timeout on test so either we need to have some option in command line to exclude a test. Something like: ``buildtest build -c tests/ -x tests/sched/partitions.yml`` The -x for exclude could be used for specifying multiple configs maybe separated by a comma. The other option is have a key maybe called ``skip: True`` as part of a sub-test or part of the global config which skips all tests.

Something like this if its per subtest

```
version: 0.0.1
test1:
  skip: True
  type: script
  description: dummy test
  run: echo "dummy"
```

Or part of global configs such that it skips all tests something like this

```
version: 0.0.1
skip: True
test1:

  type: script
  description: dummy test
  run: echo "dummy"
```

Or we can go more crazy by skip by name of subtest so that user can do something like this which will skip test1 and test2

```
version: 0.0.1
skip: [ "test1", "test2" ]
test1:
  type: script
  description: dummy test
  run: echo "dummy"
test2:
  type: script
  description: dummy test
  run: echo "dummy"
test3:
  type: script
  description: dummy test
  run: echo "dummy"
```

shahzebsiddiqui:
This PR will address #267 
- The option is ``-x`` and long option ``--exclude``
- The exclude can be passed multiple times and it can be a file or directory or both and works with shell expansion
- If -x is not specified then nothing to exclude
- Exclusion happens after configs are discovered
- If no configs are to be processed after exclusion then buildtest will stop since nothing is to be done
- Exclusion happens after ``discover_configs``
- There is a regtest to cover exclude
- Also updated a few example configs so that all test run to completion whether they PASS/FAIL 

Here are few examples to illustrate how it works

Undo detection example
```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples -x examples/
There are no config files to process.
```

Example removing two configs  file from exclusion
```

(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/script/ -x examples/script/bzip2.yml -x examples/script/slurm.yml
Config Name                    SubTest                        Status                         Config Path              
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
zlib                           zlib_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
zlib                           zlib_decompress                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
                         Test summary
Executed 3 tests
Passed Tests: 3/3 Percentage: 100.0%
Failed Tests: 0/3 Percentage: 0.0%
```

Exclude by directory in this case everything is skipped under ``examples/script`` 

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/ -x examples/script/
Config Name                    SubTest                        Status                         Config Path              
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 2 tests
Passed Tests: 2/2 Percentage: 100.0%
Failed Tests: 0/2 Percentage: 0.0%

```

Here is an example where one directory and file is passed. Notice in this case we exclude the whole benchmark sub-directory  in this case only contains stream test and we exclude bzip2 as well.

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/ -x examples/benchmarks/ -x examples/script/bzip2.yml
Config Name                    SubTest                        Status                         Config Path              
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
zlib                           zlib_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
zlib                           zlib_decompress                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
slurm                          slurm_down_nodes_reason        FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
                         Test summary
Executed 5 tests
Passed Tests: 3/5 Percentage: 60.0%
Failed Tests: 2/5 Percentage: 40.0%

```

### DIFF
--- a/buildtest/menu/build.py
+++ b/buildtest/menu/build.py
@@ -15,6 +15,8 @@ from buildtest.config import load_configuration, check_configuration
 from buildtest.executors.base import BuildExecutor
 from buildtest.utils.file import walk_tree, resolve_path
 
+logger = logging.getLogger(__name__)
+
 
 def discover_configs(config_file):
     """Given a config file specified by the user with buildtest build -c,
@@ -34,7 +36,7 @@ def discover_configs(config_file):
        # relative directory path in build test root (returns multiple)
        buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/
     """
-    logger = logging.getLogger(__name__)
+
     config_files = []
 
     # If no config file provided, assume discovering across buildtest/site
@@ -96,8 +98,6 @@ def include_file(file_path, white_list_patterns):
        :rtype: bool or str
     """
 
-    logger = logging.getLogger(__name__)
-
     if not white_list_patterns:
         logger.debug("No white list patterns provided, returning True")
         return True
@@ -127,9 +127,6 @@ def func_build_subcmd(args):
        :rtype: None
     """
 
-    # buildtest logger
-    logger = logging.getLogger(__name__)
-
     # if buildtest settings specified on CLI, it would be in args.settings otherwise set
     # to default configuration (BUILDTEST_CONFIG_FILE)
     config_file = args.settings or BUILDTEST_CONFIG_FILE
@@ -152,16 +149,11 @@ def func_build_subcmd(args):
     # if no files discovered let's stop now
     if not config_files:
         msg = "There are no config files to process."
-        logger.error(msg)
         sys.exit(msg)
 
-    print("\n")
-    print("{:^45}".format("Discovered Files"))
-    print("\n")
-    for config in config_files:
-        print(config)
-    print("\n")
-    print("\n")
+    print("\n {:^45} \n".format("Discovered Files"))
+    [print(config) for config in config_files]
+    print("\n\n")
 
     logger.debug(
         f"Based on input argument: -c {args.config} buildtest discovered the following configuration {config_files}"
@@ -169,14 +161,10 @@ def func_build_subcmd(args):
 
     if args.exclude:
         logger.debug(f"The exclude config pattern are the following: -e {args.exclude}")
-
-    config_files = [
-        config for config in config_files if include_file(config, args.exclude)
-    ]
-
-    logger.debug(
-        f"Normalized configuration list that buildtest will process are the following: {config_files}"
-    )
+        config_files = [
+            config for config in config_files if include_file(config, args.exclude)
+        ]
+        logger.debug(f"Configuration List after applying exclusion: {config_files}")
 
     # Keep track of total metrics
     total_tests = 0
@@ -214,9 +202,7 @@ def func_build_subcmd(args):
                 result = executor.dry_run(builder)
 
     if not args.dry:
-        print("\n")
-        print("\n")
-        print("{:=<60}".format(""))
+        print("\n\n{:=<60}".format(""))
         print("{:^60}".format("Test summary"))
         print("{:=<60}".format(""))
         print(f"Executed {total_tests} tests")
--- a/buildtest/menu/build.py
+++ b/buildtest/menu/build.py
@@ -86,14 +86,14 @@ def include_file(file_path, white_list_patterns):
        provided method will return ``True``. Otherwise it will return the list of files that don't
        match the regular expression
 
-    Parameters:
+       Parameters:
 
-    :param file_path: file path to run regular expression upon
-    :type file_path: str, required
-    :param white_list_patterns: the exclude list provided on command line option
-    :type white_list_patterns: list, required
-    :return: Returns True or a list of files that don't match regular expression
-    :rtype: bool or str
+       :param file_path: file path to run regular expression upon
+       :type file_path: str, required
+       :param white_list_patterns: the exclude list provided on command line option
+       :type white_list_patterns: list, required
+       :return: Returns True or a list of files that don't match regular expression
+       :rtype: bool or str
     """
 
     logger = logging.getLogger(__name__)
@@ -116,13 +116,15 @@ def include_file(file_path, white_list_patterns):
 
 def func_build_subcmd(args):
     """Entry point for ``buildtest build`` sub-command. Depending on the command
-    arguments, buildtest will set values in dictionary ``config_opts`` that is used
-    to trigger the appropriate build action.
+       arguments, buildtest will set values in dictionary ``config_opts`` that is used
+       to trigger the appropriate build action.
 
-    :param args: arguments passed from command line
-    :type args: dict, required
+       Parameters:
 
-    :rtype: None
+       :param args: arguments passed from command line
+       :type args: dict, required
+
+       :rtype: None
     """
 
     # buildtest logger
@@ -147,6 +149,20 @@ def func_build_subcmd(args):
     # Discover list of one or more config files based on path provided
     config_files = discover_configs(args.config)
 
+    # if no files discovered let's stop now
+    if not config_files:
+        msg = "There are no config files to process."
+        logger.error(msg)
+        sys.exit(msg)
+
+    print ("\n")
+    print ("{:^45}".format("Discovered Files"))
+    print ("\n")
+    for config in config_files:
+        print(config)
+    print ("\n")
+    print ("\n")
+
     logger.debug(
         f"Based on input argument: -c {args.config} buildtest discovered the following configuration {config_files}"
     )
@@ -162,12 +178,7 @@ def func_build_subcmd(args):
         f"Normalized configuration list that buildtest will process are the following: {config_files}"
     )
 
-    # config_files = exclude_configs(config_files, args.exclude)
 
-    if not config_files:
-        msg = "There are no config files to process."
-        logger.error(msg)
-        sys.exit(msg)
 
     # Keep track of total metrics
     total_tests = 0
@@ -205,16 +216,18 @@ def func_build_subcmd(args):
                 result = executor.dry_run(builder)
 
     if not args.dry:
-        print
-        print
-        print("==============================================================")
-        print("                         Test summary                         ")
+        print ("\n")
+        print ("\n")
+        print("{:=<60}".format(""))
+        print("{:^60}".format("Test summary"))
+        print("{:=<60}".format(""))
         print(f"Executed {total_tests} tests")
-        print(
-            f"Passed Tests: {passed_tests}/{total_tests} Percentage: {passed_tests*100/total_tests}%"
-        )
-        print(
-            f"Failed Tests: {failed_tests}/{total_tests} Percentage: {failed_tests*100/total_tests}%"
-        )
+
+        pass_rate = passed_tests * 100 / total_tests
+        fail_rate = failed_tests * 100 / total_tests
+
+        print(f"Passed Tests: {passed_tests}/{total_tests} Percentage: {pass_rate:.3f}%")
+
+        print(f"Failed Tests: {failed_tests}/{total_tests} Percentage: {fail_rate:.3f}%")
         print
         print
--- a/buildtest/menu/build.py
+++ b/buildtest/menu/build.py
@@ -74,58 +74,44 @@ def discover_configs(config_file):
         logger.error(msg)
         sys.exit(msg)
 
-    # return all configs with absolute path
-    tmp = [os.path.abspath(config) for config in config_files]
-    config_files = tmp
+    # return all configs by resolving path, this gets the real canonical path and address shell expansion and user expansion
+    config_files = [resolve_path(config) for config in config_files]
 
     logger.info(f"Found the following config files: {config_files}")
     return config_files
 
 
-def exclude_configs(config_files, exclude_list):
-    """This method will exclude configs from being processed after discovery and
-       this implements option ``buildtest build -x``. User can pass multiple
-       configs to be exclude which could be a file or a directory. This method
-       will return a modified list with configs that are excluded. If an invalid
-       file or directory is passed to exclude_list it will be ignored.
+def include_file(file_path, white_list_patterns):
+    """Check if file is included based on OR regular expression. If no white_list_patterns
+       provided method will return ``True``. Otherwise it will return the list of files that don't
+       match the regular expression
 
-       Parameters:
+    Parameters:
 
-       :param config_files: List of discovered configurations
-       :type config_files: list
-       :param exclude_list: List of exclude configurations
-       :type exclude_list: list
-       :return: a modified list after discovery with excluded configurations
-       :rtype: list
+    :param file_path: file path to run regular expression upon
+    :type file_path: str, required
+    :param white_list_patterns: the exclude list provided on command line option
+    :type white_list_patterns: list, required
+    :return: Returns True or a list of files that don't match regular expression
+    :rtype: bool or str
     """
-    # if there is nothing to exclude return original list
-    if not exclude_list:
-        return config_files
 
-    # check all configs exist and return a list of configs with absolute path with shell expansion.
-    tmp = [resolve_path(config) for config in exclude_list]
-    exclude_list = tmp
+    logger = logging.getLogger(__name__)
 
-    # empty list to store all exclude config files
-    exclude_config_files = []
-    for config in exclude_list:
-        # if its a directory traverse directory and find all config files with .yml extension
-        if os.path.isdir(config):
-            tmp = walk_tree(config, ".yml")
-            exclude_config_files += tmp
-        elif os.path.isfile(config):
-            exclude_config_files.append(config)
+    if not white_list_patterns:
+        logger.debug("No white list patterns provided, returning True")
+        return True
 
-    # if exclude list is empty after attempting to find all configs then return original list since nothing to return
-    if not exclude_config_files:
-        return config_files
+    logger.debug(f"white list pattern before resolving paths: {white_list_patterns}")
 
-    # for every exclude config see if it exist in discovered list of configs and try to remove it
-    for config in exclude_config_files:
-        if config in config_files:
-            config_files.remove(config)
+    white_list_patterns = [resolve_path(path) for path in white_list_patterns]
 
-    return config_files
+    logger.debug(f"white list pattern after resolving paths: {white_list_patterns}")
+
+    regexp = "(%s)" % "|".join(white_list_patterns)
+    logger.debug(f"Applying Regular Expression Search: {regexp} to file: {file_path}")
+
+    return not re.search(regexp, file_path)
 
 
 def func_build_subcmd(args):
@@ -161,10 +147,26 @@ def func_build_subcmd(args):
     # Discover list of one or more config files based on path provided
     config_files = discover_configs(args.config)
 
-    config_files = exclude_configs(config_files, args.exclude)
+    logger.debug(
+        f"Based on input argument: -c {args.config} buildtest discovered the following configuration {config_files}"
+    )
+
+    if args.exclude:
+        logger.debug(f"The exclude config pattern are the following: -e {args.exclude}")
+
+    config_files = [
+        config for config in config_files if include_file(config, args.exclude)
+    ]
+
+    logger.debug(
+        f"Normalized configuration list that buildtest will process are the following: {config_files}"
+    )
+
+    # config_files = exclude_configs(config_files, args.exclude)
 
     if not config_files:
         msg = "There are no config files to process."
+        logger.error(msg)
         sys.exit(msg)
 
     # Keep track of total metrics
--- a/buildtest/menu/build.py
+++ b/buildtest/menu/build.py
@@ -13,7 +13,7 @@ from buildtest.defaults import TESTCONFIG_ROOT, BUILDTEST_CONFIG_FILE
 from buildtest.buildsystem.base import BuildConfig
 from buildtest.config import load_configuration, check_configuration
 from buildtest.executors.base import BuildExecutor
-from buildtest.utils.file import walk_tree
+from buildtest.utils.file import walk_tree, resolve_path
 
 
 def discover_configs(config_file):
@@ -103,13 +103,8 @@ def exclude_configs(config_files, exclude_list):
         return config_files
 
     # check all configs exist and return a list of configs with absolute path with shell expansion.
-    tmp = [
-        os.path.expandvars(os.path.abspath(config))
-        for config in exclude_list
-        if os.path.exists(os.path.expandvars(os.path.abspath(config)))
-    ]
+    tmp = [resolve_path(config) for config in exclude_list]
     exclude_list = tmp
-
 
     # empty list to store all exclude config files
     exclude_config_files = []
--- a/buildtest/menu/build.py
+++ b/buildtest/menu/build.py
@@ -74,9 +74,57 @@ def discover_configs(config_file):
         logger.error(msg)
         sys.exit(msg)
 
+    # return all configs with absolute path
+    tmp = [ os.path.abspath(config) for config in config_files ]
+    config_files = tmp
+
     logger.info(f"Found the following config files: {config_files}")
     return config_files
 
+def exclude_configs(config_files, exclude_list):
+    """This method will exclude configs from being processed after discovery and
+       this implements option ``buildtest build -x``. User can pass multiple
+       configs to be exclude which could be a file or a directory. This method
+       will return a modified list with configs that are excluded. If an invalid
+       file or directory is passed to exclude_list it will be ignored.
+
+       Parameters:
+
+       :param config_files: List of discovered configurations
+       :type config_files: list
+       :param exclude_list: List of exclude configurations
+       :type exclude_list: list
+       :return: a modified list after discovery with excluded configurations
+       :rtype: list
+    """
+    # if there is nothing to exclude return original list
+    if not exclude_list:
+        return config_files
+
+    # check all configs exist and return a list of configs with absolute path with shell expansion.
+    tmp = [ os.path.expandvars(os.path.abspath(config)) for config in exclude_list if os.path.exists(os.path.expandvars(os.path.abspath(config))) ]
+    exclude_list = tmp
+
+    # empty list to store all exclude config files
+    exclude_config_files = []
+    for config in exclude_list:
+        # if its a directory traverse directory and find all config files with .yml extension
+        if os.path.isdir(config):
+            tmp = walk_tree(config, ".yml")
+            exclude_config_files += tmp
+        elif os.path.isfile(config):
+            exclude_config_files.append(config)
+
+    # if exclude list is empty after attempting to find all configs then return original list since nothing to return
+    if not exclude_config_files:
+        return config_files
+
+    # for every exclude config see if it exist in discovered list of configs and try to remove it
+    for config in exclude_config_files:
+        if config in config_files:
+            config_files.remove(config)
+
+    return config_files
 
 def func_build_subcmd(args):
     """Entry point for ``buildtest build`` sub-command. Depending on the command
@@ -110,6 +158,13 @@ def func_build_subcmd(args):
 
     # Discover list of one or more config files based on path provided
     config_files = discover_configs(args.config)
+
+    config_files = exclude_configs(config_files, args.exclude)
+
+    if not config_files:
+        msg = "There are no config files to process."
+        sys.exit(msg)
+
 
     # Keep track of total metrics
     total_tests = 0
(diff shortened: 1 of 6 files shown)


shahzebsiddiqui:
There is one issue which i am not really sure how to address and i think its a really corner case and it has to do with symlinks

My working directory is the following

```
(buildtest-framework) ssi29@ag-mxg-hulk090> pwd
/u/users/ssi29/gpfs/buildtest-framework
```

But in reality it is not the case since i have a symlink to the gpfs workspace 

```
(buildtest-framework) ssi29@ag-mxg-hulk090> ll -l /u/users/ssi29/ | grep gpfs
lrwxrwxrwx  1 ssi29 hks      21 Oct  9  2019 gpfs -> /mxg-hpc/users/ssi29/
```

So what happens is the shell expansion such as $PWD is right but in fact the abspath to file is in /mxg-hpc

What happens is exclude doesnt work in this case

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/benchmarks/ -x $PWD/examples/benchmarks/
exclude: ['/u/users/ssi29/gpfs/buildtest-framework/examples/benchmarks/stream.yml']
include: ['/mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml']
Config Name                    SubTest                        Status                         Config Path              
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 2 tests
Passed Tests: 2/2 Percentage: 100.0%
Failed Tests: 0/2 Percentage: 0.0%

```

But if i go to actual directory path without cd from symlink then it would work as expected. Don't mind the extra prints were for debuging and not going to be committed.

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/benchmarks/ -x $PWD/examples/benchmarks/
exclude: ['/mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml']
include: ['/mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml']
There are no config files to process.
(buildtest-framework) ssi29@ag-mxg-hulk090> pwd
/mxg-hpc/users/ssi29/buildtest-framework

```

vsoch:
I think you could use `os.path.islink()`. If you then used `os.readlink` that would only go one level, so instead you can use Pathlib. See https://stackoverflow.com/questions/33232417/how-to-get-symlink-target-in-python?rq=1. I would add a utils function that we can give any path to and it returns the resolved link.

buildtest/menu/build.py
```diff
@@ -74,10 +74,65 @@ def discover_configs(config_file):
         logger.error(msg)
         sys.exit(msg)
 
+    # return all configs with absolute path
+    tmp = [os.path.abspath(config) for config in config_files]
```
vsoch: This is where you'd want to put a more elegant util function to handle returning absolute, symlink resolved paths.

vsoch: Also we should try to avoid a variable name like `tmp` if possible, there is usually a better way to do it.

buildtest/menu/build.py
```diff
@@ -74,10 +74,65 @@ def discover_configs(config_file):
...
+    config_files = tmp
+
     logger.info(f"Found the following config files: {config_files}")
     return config_files
 
 
+def exclude_configs(config_files, exclude_list):
```
vsoch: This is way overly complicated a function for such a simple action! We basically need to get the resolved paths for all discovered configs, create a single regular expression that combines the exclude patterns, and then (in one loop) go over them and keep ones not excluded. Here is an example of a function I wrote that does the same thing in just a few lines: https://github.com/urlstechie/urlchecker-python/blob/04c84629eabaa70a131a670308eb54c741d62cf2/urlchecker/core/fileproc.py#L37

And then is used in a single iterator https://github.com/urlstechie/urlchecker-python/blob/04c84629eabaa70a131a670308eb54c741d62cf2/urlchecker/core/fileproc.py#L82

examples/script/bzip2.yml
```diff
@@ -11,12 +11,8 @@ bzip_compression:
     dd if=/dev/zero of="$FILE" bs=1M count=1
     tar cvjf "$TARFILE" "$FILE"
     ls -l "$TARFILE"
-
-bzip_version:
```
vsoch, line 15: Why did you remove this test?

examples/script/zlib.yml
```diff
@@ -2,18 +2,28 @@ version: 0.0.1
...
+  env:
+    FILE: 1mb.txt
+    TARFILE: 1mb.tar.gz
   run: |
-    dd if=/dev/zero of=1mb.txt bs=1M count=1
-    tar cvzf 1mb.tar.gz 1mb.txt
-    ls -lh 1mb.tar.gz
-    gzip -l 1mb.tar.gz
+    dd if=/dev/zero of=$FILE bs=1M count=1
+    tar cvzf $TARFILE $FILE
```
vsoch: Is there a reason you don't quote bash variables or put them in `{}`? I thought that was considered best practice.

vsoch (changes requested):

examples/script/bzip2.yml
shahzebsiddiqui: it was getting a non-zero exit and was failing

examples/script/zlib.yml
shahzebsiddiqui: that's a typo on my end i will use ``{}`` i can fix this

buildtest/menu/build.py
shahzebsiddiqui: When i was implementing ``--exclude`` i didn't intend for regular expression. Regex would be nice but i foresee that to be a significant change or maybe even a new option. Right now i mimic ``-c`` with ``-x`` which supports file or directory. Both of them don't support regex. 

I think advanced users that know regex would prefer a pattern search for both ``-c`` and ``-x``, but most novice users would prefer just passing a file or directory.

Would it make sense to stand up a new option for regular expression, or we can support regex into both ``-c`` and ``-x`` to be consistent. One thing that i realized was ``-c`` can't be used twice which i can change by adding action ``append`` so something like

```
buildtest build -c dir1 -c dir2
```

vsoch: The --exclude doesn't take a regexp, it turns into a list, and that would be parsed into a regular expression. E.g.,:

```python
exclude = ["dir1", "dir2"]
```
turns into

```python
"(dir1|dir2)"
```
The command line interface you have does not change.

buildtest/menu/build.py
shahzebsiddiqui: Okay i was able to resolve this and the bug i posted about symlinks by using os.realpath https://docs.python.org/2/library/os.path.html#os.path.realpath which resolves symlinks.

Now this test is working as expected. 
 ```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/benchmarks/  -x $PWD/examples/benchmarks
There are no config files to process.
```

My symlink gpfs is now resolving any any subdirectory in symlink works too
```
>>> os.path.realpath('/u/users/ssi29/gpfs')
'/mxg-hpc/users/ssi29'
>>> os.path.realpath('/u/users/ssi29/gpfs/buildtest-framework')
'/mxg-hpc/users/ssi29/buildtest-framework'
```

What won't work is  shell expansion is required then realpath won't expand automatically

```
>>> os.path.realpath('$HOME/gpfs/buildtest-framework')
'/mxg-hpc/users/ssi29/buildtest-framework/$HOME/gpfs/buildtest-framework'
```

But doing shell expansion first then doing realpath would work

```
>>> os.path.realpath(os.path.expandvars('$HOME/gpfs/buildtest-framework'))
'/mxg-hpc/users/ssi29/buildtest-framework'
```

Then it comes to the other odd case of ``~`` if its used when referring to file, so not sure if ``resolve_path`` method should take care of all of these cases in the following order
- Run os.path.expandvars address ``$HOME``
- Run os.path.expandusers address ``~``
- Run os.path.realpath to resolve symlinks
- Finally run ``os.path.exists`` to be absolutely sure its a path.

File: buildtest/utils/file.py [hunk 5]
```diff
@@ -119,3 +119,22 @@ def create_dir(dirname):
...
+       :param path: file path to resolve
+       :type path: str, required
+       :return: return realpath to file if found otherwise return None
+       :rtype: str or None
+    """
+    shell_expansion = os.path.expandvars(path)
+    real_path = os.path.realpath(shell_expansion)
+    if os.path.exists(real_path):
+        return real_path
+    else:
```
vsoch: A function will by default return None if nothing is returned, so you don't need this else -> return None.

shahzebsiddiqui: that was a dumb mistake good catch

shahzebsiddiqui: This is fixed now

buildtest/menu/build.py
vsoch: You don't need the tmp here, you can have an iterator that returns to a variable with the same name (not like a dictionary that you can't update in place).
```python
config_files = [os.path.abspath(config) for config in config_files]
```

buildtest/menu/build.py
shahzebsiddiqui: I see the regex examples i am not entirely if regex will do trick or my lack of understanding. For a directory i need to retrieve all yml files and add to exclude list before checking if they are already discovered so i can remove them.

vsoch: Maybe this will help:

 1. You start with a list of files, config_files
 2. You run them through an iterate to filter that determines if they should be included
 3. The filter takes the entire list of excludes and generates _one_ regular expression to check the _one_ config file path that is being passed over
 4. The function returns True or False to determine to include or not. E.g.,

```
exclude_list = ["dir1", "dir2"]
config_files = [config for config in config_files if include_file(config, exclude_list)]
```
And the include file function is similar to what I showed you.

shahzebsiddiqui: Okay i fixed this now and i was able to fix resolve path to support both shell and user expansion and applied resolve path to config files too. So this two tests are now discovered properly

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c $PWD/examples
Config Name                    SubTest                        Status                         Config Path              
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
zlib                           zlib_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
zlib                           zlib_decompress                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
bzip2                          bzip_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip_help                      PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip2_help                     PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
slurm                          slurm_down_nodes_reason        FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 10 tests
Passed Tests: 8/10 Percentage: 80.0%
Failed Tests: 2/10 Percentage: 20.0%
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c ~/gpfs/buildtest-framework/examples
Config Name                    SubTest                        Status                         Config Path              
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
zlib                           zlib_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
zlib                           zlib_decompress                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
bzip2                          bzip_compression               PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip_help                      PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
bzip2                          bzip2_help                     PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml
slurm                          slurm_down_nodes_reason        FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 10 tests
Passed Tests: 8/10 Percentage: 80.0%
Failed Tests: 2/10 Percentage: 20.0%

```

shahzebsiddiqui: Thanks @vsoch  the solution does work i had to make a minor adjustment to apply ``resolve_path`` before running regex since paths needed to be resolved for regex to work, i had applied some logging to show some details.

Here is an example with user expansion
 
```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c ~/gpfs/buildtest-framework/examples -x ~/gpfs/buildtest-framework/examples/
There are no config files to process.
```

The logfile will have the following
```
2020-04-10 09:38:39,875 [build.py:155 - func_build_subcmd() ] - [DEBUG] The exclude config pattern are the following: -e ['/u/users/ssi29/gpfs/buildtest-framework/examples/']
2020-04-10 09:38:39,876 [build.py:105 - include_file() ] - [DEBUG] white list pattern before resolving paths: ['/u/users/ssi29/gpfs/buildtest-framework/examples/']
2020-04-10 09:38:39,876 [build.py:109 - include_file() ] - [DEBUG] white list pattern after resolving paths: ['/mxg-hpc/users/ssi29/buildtest-framework/examples']
2020-04-10 09:38:39,876 [build.py:112 - include_file() ] - [DEBUG] Applying Regular Expression Search: (/mxg-hpc/users/ssi29/buildtest-framework/examples) to file: /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
```

Here is an example with shell expansion where only ``examples/benchmark`` is supposed to be processed

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/ -x $HOME/gpfs/buildtest-framework/examples/script
Config Name                    SubTest                        Status                         Config Path              
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 2 tests
Passed Tests: 2/2 Percentage: 100.0%
Failed Tests: 0/2 Percentage: 0.0%
```

This example just shows we pass multiple exclusion via ``-x``  which works by filtering output as expected

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/ -x examples/script/bzip2.yml -x examples/script/zlib.yml
Config Name                    SubTest                        Status                         Config Path              
python-circle                  circle_area                    PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/python-circle.yml
slurm                          slurm_down_nodes_reason        FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
slurm                          slurm_not_responding_nodes     FAILED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/script/slurm.yml
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
                         Test summary
Executed 5 tests
Passed Tests: 3/5 Percentage: 60.0%
Failed Tests: 2/5 Percentage: 40.0%
```

And the logfile will have the following regex

```
2020-04-10 09:45:06,706 [build.py:112 - include_file() ] - [DEBUG] Applying Regular Expression Search: (/mxg-hpc/users/ssi29/buildtest-framework/examples/script/bzip2.yml|/mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml) to file: /mxg-hpc/users/ssi29/buildtest-framework/examples/script/zlib.yml
```

shahzebsiddiqui:
@vsoch i changed the output a bit 

- Add list of discovered files on top
- Change output of floating point to 3 digits. Previously it was printing a whole bunch of digits after decimal which was not needed

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -c examples/benchmarks/


              Discovered Files


/mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml




Config Name                    SubTest                        Status                         Config Path              
stream                         stream_uniprocess_c            PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml
stream                         stream_openmp_c                PASSED                         /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks/stream.yml




                        Test summary
Executed 2 tests
Passed Tests: 2/2 Percentage: 100.000%
Failed Tests: 0/2 Percentage: 0.000%
```

buildtest/menu/build.py
```diff
@@ -74,19 +74,57 @@ def discover_configs(config_file):
...
+    # return all configs by resolving path, this gets the real canonical path and address shell expansion and user expansion
+    config_files = [resolve_path(config) for config in config_files]
+
     logger.info(f"Found the following config files: {config_files}")
     return config_files
 
 
+def include_file(file_path, white_list_patterns):
+    """Check if file is included based on OR regular expression. If no white_list_patterns
+       provided method will return ``True``. Otherwise it will return the list of files that don't
```
vsoch, line 88: so much cleaner! <3

buildtest/menu/build.py
```diff
@@ -74,19 +74,57 @@ def discover_configs(config_file):
...
+
+       :param file_path: file path to run regular expression upon
+       :type file_path: str, required
+       :param white_list_patterns: the exclude list provided on command line option
+       :type white_list_patterns: list, required
+       :return: Returns True or a list of files that don't match regular expression
+       :rtype: bool or str
+    """
+
+    logger = logging.getLogger(__name__)
```
vsoch: Is there a reason that the logger is instantiated here, instead of for the module?

buildtest/menu/build.py
```diff
@@ -74,19 +74,57 @@ def discover_configs(config_file):
...
+        return True
+
+    logger.debug(f"white list pattern before resolving paths: {white_list_patterns}")
+
+    white_list_patterns = [resolve_path(path) for path in white_list_patterns]
+
+    logger.debug(f"white list pattern after resolving paths: {white_list_patterns}")
+
+    regexp = "(%s)" % "|".join(white_list_patterns)
+    logger.debug(f"Applying Regular Expression Search: {regexp} to file: {file_path}")
```
vsoch, line 112: This is probably one too many lines for the logger - is there a lower level than debug (e.g., verbose) for at least the last line?

buildtest/menu/build.py
```diff
@@ -111,6 +149,35 @@ def func_build_subcmd(args):
     # Discover list of one or more config files based on path provided
     config_files = discover_configs(args.config)
 
+    # if no files discovered let's stop now
+    if not config_files:
+        msg = "There are no config files to process."
+        logger.error(msg)
```
vsoch: What is the reason for wanting the message to go to the logger.error _and_ system.exit?

buildtest/menu/build.py
```diff
@@ -111,6 +149,35 @@ def func_build_subcmd(args):
...
+        sys.exit(msg)
+
+    print("\n")
+    print("{:^45}".format("Discovered Files"))
+    print("\n")
```
vsoch: Can we put the newlines with other statements so this single block doesn't need to take up 7 lines?

buildtest/menu/build.py
```diff
@@ -111,6 +149,35 @@ def func_build_subcmd(args):
...
+
+    logger.debug(
+        f"Based on input argument: -c {args.config} buildtest discovered the following configuration {config_files}"
+    )
+
+    if args.exclude:
+        logger.debug(f"The exclude config pattern are the following: -e {args.exclude}")
+
+    config_files = [
```
vsoch: Is this needed to run if args.exclude isn't set? Shouldn't it be under the if args.exclude?

buildtest/menu/build.py
```diff
@@ -111,6 +149,35 @@ def func_build_subcmd(args):
...
+        config for config in config_files if include_file(config, args.exclude)
+    ]
+
+    logger.debug(
```
vsoch: What do you mean normalized?

buildtest/menu/build.py
```diff
@@ -147,16 +214,22 @@ def func_build_subcmd(args):
                 result = executor.dry_run(builder)
 
     if not args.dry:
-        print
-        print
-        print("==============================================================")
-        print("                         Test summary                         ")
+        print("\n")
+        print("\n")
+        print("{:=<60}".format(""))
```
vsoch: Let's also combine some of these statements, you can add the \n to have more than one per print().

buildtest/utils/file.py
```diff
@@ -119,3 +119,24 @@ def create_dir(dirname):
...
+       when a symlink is provided in the file. This method assumes file already exists.
+
+       Parameter:
+
+       :param path: file path to resolve
+       :type path: str, required
+       :return: return realpath to file if found otherwise return None
+       :rtype: str or None
+    """
+    # apply shell expansion  when file includes something like $HOME/example
```
vsoch, line 135: Great job on documentation here!

buildtest/utils/file.py
```diff
@@ -119,3 +119,24 @@ def create_dir(dirname):
         except OSError as err:
             print(err)
             raise
+
+
+def resolve_path(path):
+    """This method will resolve a file path with accounts for shell expansion and resolve paths in
```
vsoch: This method will resolve a file path _to account for_ shell expansion...

examples/script/bzip2.yml
```diff
@@ -8,15 +8,11 @@ bzip_compression:
...
-  type: script
-  description: "bzip2 version"
-  shell: sh
-  run: "bzip2 --version"
+    dd if=/dev/zero of=${FILE} bs=1M count=1
+    tar cvjf ${TARFILE} ${FILE}
+    ls -l ${TARFILE}
+    rm $TARFILE
```
vsoch: Nit: be consistent in using the {} for the lsat line as well.

buildtest/menu/build.py
shahzebsiddiqui: ahh you mean just instantiate once per file as global variable and not call inside each method. I think that makes sense. If that's what you mean i think it makes sense and i agree with your logic.

vsoch: Exactly - that's why we use __name__ for the logger, it refers to the active module.

buildtest/menu/build.py
shahzebsiddiqui: DEBUG is the lowest level i believe according to https://docs.python.org/3/howto/logging.html We could either remove the line if it's too much detail or we can set default debug setting to INFO. The default is ``WARNING``.

I figured it could be useful for users to troubleshoot how exclusion is happening if they have information on the regex command.

vsoch: ah okay, then okay to leave it. I think this is why I implemented a custom logger for most of my libraries that has also logger.verbose().

buildtest/menu/build.py
shahzebsiddiqui: Good point, i don't have any particular reasoning. I guess it doesn't make sense to have same error message if printed to screen. I guess it would be only useful if an exception message is captured before exiting for detailed error message. That's what i am accustomed to, just ``tail``  the last few lines to get the error message.

buildtest/menu/build.py
shahzebsiddiqui: Makes sense i can make this adjustment

buildtest/menu/build.py
vsoch: I was thinking it would be cleaner to do something like:

```python
logger.error("There was an error")  # <--- error message goes to log
sys.exit(1)                                       # <--- exit with error code
```

shahzebsiddiqui: I will remove this line and fixed locally.

buildtest/menu/build.py
shahzebsiddiqui: that would be most appropriate i will make this fix.

buildtest/menu/build.py
shahzebsiddiqui: Bad naming, what about *filtered*, need some way to denote that configs discovered after exclusion are the ones that buildtest will process.

buildtest/menu/build.py
shahzebsiddiqui: makes sense. Fixed now

buildtest/utils/file.py
shahzebsiddiqui: fixed locally

examples/script/bzip2.yml
shahzebsiddiqui: fixed locally good catch.

buildtest/menu/build.py
shahzebsiddiqui: I fixed this now

buildtest/menu/build.py
```diff
@@ -152,31 +149,22 @@ def func_build_subcmd(args):
...
-    print("\n")
-    print("{:^45}".format("Discovered Files"))
-    print("\n")
-    for config in config_files:
-        print(config)
-    print("\n")
-    print("\n")
+    print("\n {:^45} \n".format("Discovered Files"))
+    [print(config) for config in config_files]
+    print("\n\n")
```
vsoch, line 156: Much more succinct.

vsoch (approved):

PR #269, opened 2020-04-10
Redefinition of buildtest terms 

### LINKED ISSUE #266: define a set of buildtest terminology for docs and refactor code accordingly
Currently, there are few terms used in buildtest codebase that is confusing to developers and end-users & docs. Futhermore it prevents me from writing descriptive log message until we have the terminology defined for each buildtest object. 

Once these objects are defined we should update the terminology section in docs right now at https://buildtest.readthedocs.io/en/devel/#terminology.

I see few buildtest objects
1. buildtest configuration file (settings.yml) aka site configuration
2. buildtest test configuration (YAML file) for writing test compatible with buildtest schema library
3. buildtest testscript - generated test from buildtest
4. subtest - A single test configuration can define one or more sub-test, there is a one-one relation between subtest and testscript generated by buildtest
5. outer schema - global keys shared by all schemas
6. Executors - an executor object used to launch the test
7. ExecutorType - a type of Executor (LocalExecutor, SlurmExecutor, etc...) 
8. Executor defined in site-configuration  vs one defined in test-configuration
9. Schema library - a library of buildtest schema for writing test configuration. Currently we support ``script`` schema.

Points of confusion
1. buildtest configuration file (``settings.yml``) is called **settings** and **configuration** for example we use methods like ``load_configuration`` and ``check_configuration`` to validate configuration file that is meant to refer to buildtest settings. We should pick a name that represents buildtest configuration such that it avoids conflict with test configuration ``buildtest build --configuration``
2. The default configuration or settings file is named ``default.yml`` https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/settings/default.yml but we copy it and name it settings.yml in **$HOME/.buildtest/settings.yml**. Furthermore this adds confusion with argument ``buildtest build --settings``. 
3. The test configuration passed to buildtest (``buildtest build -c ``) is referred to as **configuration** in the codebase such as **discover_configs**,  **BuildConfig**
4. The outer schema name is bit confusing for users. They may think what is difference between outer and inner schema. The outer schema https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/schemas/outer.schema.json is being used in _validate_global method https://github.com/HPC-buildtest/buildtest-framework/blob/abcfad94b69598f87f15c58e3c0f0762d9c9a99f/buildtest/buildsystem/base.py#L138 
5. The term schema  is not specific, it could refer to schema library such as ``script`` and ``outer.schema.json`` or the buildtest  configuration schema. 

The codebase should be refactored with names meaningful to their object representation in buildtest and log messages would need to be rewritten to make correct use of the buildtest object.

First step would be to define the terminology and possibly update the docs upon agreement and the code-base can be refactored according to defined terminology.

shahzebsiddiqui:
This PR address some term definition that buildtest can adopt when referencing the docs and most importantly refactor the code with the appropriate names. Since there is a lot of confusion between using the right term, we should get this approved first before refactor the code. 

Once the terms are approved, this would imply some of the following changes 
- rename method names as needed
- rewrite log message
- rename options 
- rewrite docs 

All of these would be scoped for separate PR.

his how the page looks like built locally

image

### DIFF
--- a/docs/terminology.rst
+++ b/docs/terminology.rst
@@ -1,0 +1,22 @@
+.. _terminology:
+
+Terminology
+===========
+
+
+.. csv-table:: Core Concepts
+    :header: "name", "Description"
+    :widths: 30, 150
+
+    **SpecFile**,"A ``SpecFile`` is a YAML file that buildtest interprets when generating the test. The SpecFile may contain one or more specs."
+    **Spec**,"A Spec is the basic building block for writing a test specification that is compatible with one of the Spec Schema.",
+    **Spec Schema**,"A Spec Schema is a JSON file defining the schema with valid key/value pairs that are acceptable for the schema."
+    **Schema Library**,"A Schema Library is a collection of one or more ``Spec Schema`` provided by buildtest."
+    **TestScript**,"A TestScript is a generated shell script by buildtest as a result of processing one of the SpecFile."
+    **Settings**,"Settings is a buildtest configuration file that can be in YAML/JSON that configures buildtest at your site. The Settings file must be compatible with the Settings Schema."
+    **Settings Schema**,"A schema definition that dictates how to validate a ``Settings`` file."
+    **Executor**,"An ``Executor`` defines how a **TestScript** is executed. An executor can be of several types such as ``local``, ``slurm`` which defines if test is run locally of via scheduler.  The executors are defined in the ``Settings`` file."
+
+
+
+
--- a/docs/terminology.rst
+++ b/docs/terminology.rst
@@ -8,15 +8,15 @@ Terminology
     :header: "Name", "Description"
     :widths: 30, 60
 
-    **Buildspec**,"``Buildspec`` is a YAML file that buildtest interprets when generating the test. A Buildspec may
+    **Buildspec**," is a YAML file that buildtest interprets when generating the test. A Buildspec may
     contain one or more test blocks that is validated by a ``Buildspec Schema``."
     **Buildspec Schema**,"is a JSON file defining  valid/invalid key value pairs and formatting for the file"
-    **Schema Library**,"A Schema Library is a collection of one or more ``Buildspec Schema`` provided by buildtest."
-    **Test Script**,"A Test Script is a generated shell script by buildtest as a result of processing one of the Buildspec."
-    **Settings**,"Settings is a buildtest configuration file that can be in YAML/JSON that configures buildtest at your
+    **Schema Library**," is a collection of one or more ``Buildspec Schema`` provided by buildtest."
+    **Test Script**," is a generated shell script by buildtest as a result of processing one of the Buildspec."
+    **Settings**," is a buildtest configuration file that can be in YAML/JSON that configures buildtest at your
     site. The Settings file must be compatible with the Settings Schema."
-    **Settings Schema**,"A schema definition that dictates how to validate a ``Settings`` file."
-    **Executor**,"An ``Executor`` defines how a **TestScript** is executed. An executor can be of several types such as
+    **Settings Schema**," is a schema definition that dictates how to validate a ``Settings`` file."
+    **Executor**," defines how a **TestScript** is to be executed. An executor can be of several types such as
     ``local``, ``slurm`` which defines if test is run locally of via scheduler. The executors are defined in the
     ``Settings`` file."
 
--- a/docs/terminology.rst
+++ b/docs/terminology.rst
@@ -8,11 +8,11 @@ Terminology
     :header: "Name", "Description"
     :widths: 30, 60
 
-    **BuildSpec**,"``BuildSpec`` is a YAML file that buildtest interprets when generating the test. A BuildSpec may
-    contain one or more test blocks that is validated by a ``BuildSpec Schema``."
-    **BuildSpec Schema**,"is a JSON file defining  valid/invalid key value pairs and formatting for the file"
-    **Schema Library**,"A Schema Library is a collection of one or more ``BuildSpec Schema`` provided by buildtest."
-    **Test Script**,"A Test Script is a generated shell script by buildtest as a result of processing one of the BuildSpec."
+    **Buildspec**,"``Buildspec`` is a YAML file that buildtest interprets when generating the test. A Buildspec may
+    contain one or more test blocks that is validated by a ``Buildspec Schema``."
+    **Buildspec Schema**,"is a JSON file defining  valid/invalid key value pairs and formatting for the file"
+    **Schema Library**,"A Schema Library is a collection of one or more ``Buildspec Schema`` provided by buildtest."
+    **Test Script**,"A Test Script is a generated shell script by buildtest as a result of processing one of the Buildspec."
     **Settings**,"Settings is a buildtest configuration file that can be in YAML/JSON that configures buildtest at your
     site. The Settings file must be compatible with the Settings Schema."
     **Settings Schema**,"A schema definition that dictates how to validate a ``Settings`` file."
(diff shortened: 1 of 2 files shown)


docs/terminology.rst
```diff
@@ -0,0 +1,28 @@
...
+
+Terminology
+===========
+
+
+.. csv-table:: Core Concepts
+    :header: "Name", "Description"
+    :widths: 30, 60
+
+    **SpecFile**,"A ``SpecFile`` is a YAML file that buildtest interprets when generating the test. The SpecFile may
```
vsoch: Hmm, this is confusing for several reasons. First, calling it a "SpecFile" is just really general, how is anyone supposed to know it's relevant to build test? To follow convention of Dockerfile, Makefile, Singularity recipe, I think the file should be called something specific to buildtest? A Buildspec? And then "Spec" is confusing, because it's really referring to a block (a test) in the file. Why not just refer to this as a Build Test? Basiclly:

 - a Buildspec is a file that buildtest interprets
 - A build test defined as a block in the file that is validated by a Buildspec Schema
 - a Buildspec Schema is a JSON file defining valid / invalid key value pairs and formatting for the file
 - And then you can still use Schema Library and the rest those make sense.

But overall I think the first three are a bit confusing and we need to rethink!

vsoch (changes requested):

docs/terminology.rst
shahzebsiddiqui: I like ``BuildSpec`` that is a unique name so if we adopt this name then ``--config`` could be ``--buildspec``. I think its important the names be meaningful and unambiguous for end-users and developers.

shahzebsiddiqui: **A build test defined as a block in the file that is validated by a Buildspec Schema** is this a definition part of ``BuildSpec``. Should i remove the keyword ``Spec``.

vsoch: You shouldn't use my words verbatim, i was just giving an example.

shahzebsiddiqui: I updated can you check 2nd revision. I removed ``Spec`` that was addressing one of more test block. That is now just part of ``BuildSpec`` to avoid confusion.

docs/terminology.rst
```diff
@@ -8,14 +8,11 @@ Terminology
...
 
-    **SpecFile**,"A ``SpecFile`` is a YAML file that buildtest interprets when generating the test. The SpecFile may
-    contain one or more specs."
-    **Spec**,"A Spec is the basic building block for writing a test specification that is compatible with one of the
-    Spec Schema."
-    **Spec Schema**,"A Spec Schema is a JSON file defining the schema with valid key/value pairs that are acceptable
-    for the schema."
-    **Schema Library**,"A Schema Library is a collection of one or more ``Spec Schema`` provided by buildtest."
-    **Test Script**,"A Test Script is a generated shell script by buildtest as a result of processing one of the SpecFile."
+    **BuildSpec**,"``BuildSpec`` is a YAML file that buildtest interprets when generating the test. A BuildSpec may
```
vsoch: Do you want that capitalization? You know, it's generally `Dockerfile` and not `DockerFile`. I'm thinking we want either `buildspec` or `Buildspec` but not `BuildSpec`

shahzebsiddiqui: Okay let's do ``Buildspec`` in that case it can be used at start of sentence when it get's confusing if first letter is to be caps.
Ill fix it

shahzebsiddiqui: Okay i made this change now

docs/terminology.rst
```diff
@@ -8,11 +8,11 @@ Terminology
     :header: "Name", "Description"
     :widths: 30, 60
 
-    **BuildSpec**,"``BuildSpec`` is a YAML file that buildtest interprets when generating the test. A BuildSpec may
-    contain one or more test blocks that is validated by a ``BuildSpec Schema``."
-    **BuildSpec Schema**,"is a JSON file defining  valid/invalid key value pairs and formatting for the file"
-    **Schema Library**,"A Schema Library is a collection of one or more ``BuildSpec Schema`` provided by buildtest."
-    **Test Script**,"A Test Script is a generated shell script by buildtest as a result of processing one of the BuildSpec."
+    **Buildspec**,"``Buildspec`` is a YAML file that buildtest interprets when generating the test. A Buildspec may
```
vsoch: nit: we don't need to say it twice, just do like "Buildspec: is a yaml file..." to follow convention of the others.

shahzebsiddiqui:
Do we need a term to define the global or outer schema https://github.com/HPC-buildtest/schemas/tree/master/global 

It's bit confusing since we have method like ``_validate_global`` but file is called outer.schema and it's not clear to users. Or is it not worth documenting this. 

I guess global makes sense, the point is to have a set of keys that are required for any Buildspec. Like Kubernetes they have some keys that are required for every yml file like ``apiVersion``, ``kind``, ``metadata``. 

At some point buildtest docs should have a glossary of all key/value pairs something like Kubernetes https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/

docs/terminology.rst
shahzebsiddiqui: got it. I will fix this and all the others as needed.

vsoch:
> Do we need a term to define the global or outer schema https://github.com/HPC-buildtest/schemas/tree/master/global

I don't think so, it's hugely unlikely for this to change, very unlikely for users to interact with it (or need to know about it) and even still unlikely for developers to need to know. Adding another term for it would probably confuse more than help.

> It's bit confusing since we have method like _validate_global but file is called outer.schema and it's not clear to users. Or is it not worth documenting this.

Ah, that's a good point, we could better rename the function to match the schema, e.g., `_validate_outer_schema` might be more obvious.

> I guess global makes sense, the point is to have a set of keys that are required for any Buildspec. Like Kubernetes they have some keys that are required for every yml file like apiVersion, kind, metadata.

Or you could also call it the global schema instead of outer (both approaches seem good to me). I originally called the function _validate_global because my thinking was that there are more global validation things we might want to do, not just the schema.

> At some point buildtest docs should have a glossary of all key/value pairs something like Kubernetes https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/

Definitely!

docs/terminology.rst
shahzebsiddiqui: Okay this is now fixed in latest commit.

shahzebsiddiqui:
> > Do we need a term to define the global or outer schema https://github.com/HPC-buildtest/schemas/tree/master/global
> 
> I don't think so, it's hugely unlikely for this to change, very unlikely for users to interact with it (or need to know about it) and even still unlikely for developers to need to know. Adding another term for it would probably confuse more than help.
> 
> > It's bit confusing since we have method like _validate_global but file is called outer.schema and it's not clear to users. Or is it not worth documenting this.
> 
> Ah, that's a good point, we could better rename the function to match the schema, e.g., `_validate_outer_schema` might be more obvious.
> 
> > I guess global makes sense, the point is to have a set of keys that are required for any Buildspec. Like Kubernetes they have some keys that are required for every yml file like apiVersion, kind, metadata.
> 
> Or you could also call it the global schema instead of outer (both approaches seem good to me). I originally called the function _validate_global because my thinking was that there are more global validation things we might want to do, not just the schema.
> 
> > At some point buildtest docs should have a glossary of all key/value pairs something like Kubernetes https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/
> 
> Definitely!

I will do a complete refactor of all code-base after this PR is done I am leaning towards keeping ``global`` as the name and keep ``_validate_global``  but just change the name of json file outer.schema.json to global.schema.json if that makes sense.

vsoch (approved):

PR #270, opened 2020-04-10
refactor codebase to use name buildspec 

shahzebsiddiqui:
This PR will clean up a few things that were bit of a mess.
- First attempt to refactor code base to use the word ``Buildspec``. For variables we refer to as ``buildspec``. 
- rename option``buildtest build --config`` to ``buildtest build --buildspec`` 
- Rename schema ``outer.schema.json`` --> ``global.schema.json``
- Rename file ``default.yml`` --> ``settings.yml``
- we remove ``init_log`` not needed since it was used once. 
- Add regtest for get.py for clone feature.
- change behavior of is_dir and is_file to return a boolean and these methods run ``resolve_path`` first
- Remove ``create_file`` since its not used in codebase we can add this in if it's needed .
- ``create_dir`` won't use ``resolve_path`` since input directory is non-existent so we run the resolve manually.
- Reuse ``is_file`` and ``is_dir`` in regtest in places of os.path.exists when applicable to test buildtest modules. 
- refactor a few assert statements to be condensed such as ``assert True is statement`` to ``assert statement``


shahzebsiddiqui:
@vsoch  codecov report looks promising, this PR is meant to increase coverage and cleanup some of the codebase.

buildtest/menu/build.py
```diff
@@ -166,6 +162,15 @@ def func_build_subcmd(args):
         ]
         logger.debug(f"Configuration List after applying exclusion: {config_files}")
 
+        # if no files remain after exclusion let's stop now.
+        if not config_files:
```
vsoch: Just to check here - you don't want any message printed to the log, but you are passing to sys.exit?

buildtest/utils/file.py
```diff
@@ -11,60 +11,62 @@
...
     """This method will check if file exist and if not found throws an exception.
 
     :param file: file path
-    :type file: string, required
+    :type file: str, required
 
-    :raises BuildTestError: sub-class of Exception
     :return: returns True if file exists otherwise terminates with an exception
-    :rtype: True if successful, otherwise throws an exception
+    :rtype: bool
```
vsoch, line 24: Should this be rtype (does it mean return type or is it a typo?)

buildtest/utils/file.py
```diff
@@ -125,18 +106,19 @@ def resolve_path(path):
...
+
     # apply shell expansion  when file includes something like $HOME/example
     path = os.path.expandvars(path)
     # apply user expansion when file includes something like  ~/example
     path = os.path.expanduser(path)
 
     real_path = os.path.realpath(path)
-    if os.path.exists(real_path):
-        return real_path
+    # if os.path.exists(real_path):
```
vsoch: Why is this commented out now?

tests/menu/test_get.py
```diff
@@ -0,0 +1,27 @@
+import os
+import pytest
+import shutil
+from buildtest.defaults import TESTCONFIG_ROOT
+from buildtest.menu.get import clone
+from buildtest.utils.file import is_dir, create_dir
+
+
+def test_clone():
```
vsoch: This should use pytest tmp_path.

tests/utils/test_file.py
```diff
@@ -1,84 +1,53 @@
...
-    is_dir("$HOME")
-    is_dir("~")
-
-    assert True is os.path.isdir(os.path.expanduser("~"))
-    assert True is os.path.isdir(os.path.expandvars("$HOME"))
+    file_name = str(uuid.uuid4())
+    assert not is_file(file_name)
 
+    file1 = "~/.profile"
+    file2 = "$HOME/.profile"
```
vsoch, line 24: Can we be sure these are universally present?

tests/utils/test_file.py
```diff
@@ -1,84 +1,53 @@
...
-    # checking ~ expansion
-    assert True is os.path.isfile(os.path.expanduser("~/b.txt"))
-
-    os.remove("/tmp/a.txt")
-    os.remove(os.path.expandvars("$HOME/a.txt"))
-    os.remove(os.path.expanduser("~/b.txt"))
 
+def test_directory_expansion():
+    dir1 = "$HOME"
+    dir2 = "~"
```
vsoch, line 32: Same question as above.

tests/utils/test_file.py
```diff
@@ -1,84 +1,53 @@
...
 
 
 def test_create_dir():
-    create_dir("$HOME/a/b/c")
-    create_dir("~/x/y/z")
-    assert True is os.path.isdir(os.path.expandvars("$HOME/a/b/c"))
-    assert True is os.path.isdir(os.path.expanduser("~/x/y/z"))
+    dir1 = "$HOME/a/b/c"
+    dir2 = "~/x/y/z"
+    create_dir(dir1)
```
vsoch: Same question as above. It's generally good practice to use os.path.sep instead of hard coding like that.

buildtest/menu/build.py
shahzebsiddiqui: yes that's right, buildtest is there to inform user after discovery or after exclusion step if there are no configs. I know it's repeated twice after each stage. Atleast user should know what buildtest is trying to do.

buildtest/utils/file.py
shahzebsiddiqui: Yea this is return type, there are built-in types atleast documented in https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html i dont think it's an official standard from docstring but what happens is this formatting helps document all variables and return type that makes API docs look much better. 

Some IDEs have built-in support to auto-fill some sections atleast PyCharm can do this. https://www.jetbrains.com/help/pycharm/using-docstrings-to-specify-types.html

buildtest/utils/file.py
shahzebsiddiqui: I will revert this back, it was addressing an issue in ``is_file`` and ``is_dir`` but i have a better solution while preserving this. Stay tuned.

buildtest/buildsystem/base.py
```diff
@@ -69,28 +69,28 @@ def __init__(self, config_file):
...
-           config_file: the pull path to the configuration file, must exist.
+           buildspec_file: the pull path to the configuration file, must exist.
         """
-        self.config_file = os.path.abspath(config_file)
+        self.buildspec_file = os.path.abspath(buildspec_file)
 
-        if not os.path.exists(self.config_file):
-            sys.exit("Build configuration %s does not exist." % self.config_file)
+        if not os.path.exists(self.buildspec_file):
+            sys.exit("BuildSpec File: %s does not exist." % self.buildspec_file)
```
vsoch: Remember the capitalization stuffs we discussed!

buildtest/buildsystem/base.py
```diff
@@ -69,28 +69,28 @@ def __init__(self, config_file):
...
 
-        elif os.path.isdir(self.config_file):
+        elif os.path.isdir(self.buildspec_file):
             sys.exit(
```
vsoch, line 89: For a lot of these saying “Buildspec file” is redundant, a buildspec is a file.

buildtest/buildsystem/base.py
```diff
@@ -105,24 +105,24 @@ def __repr__(self):
     # Validation
 
     def _validate(self):
-        """Given a loaded recipe, validate that the type is known in the lookup
+        """Given a loaded Buildspec recipe, validate that the type is known in the lookup
            to buildtest. If a version is provided, honor it. If not, use latest.
            We also don't allow repeated keys in the same file.
         """
```
vsoch: The only variable I think we should keep constant and not change is recipe - it is implied that recipe for a build config that loads a buildspec file Is a build spec recipe. Making the variable this long isn’t as clean and nice looking :-)

buildtest/buildsystem/base.py
```diff
@@ -457,13 +463,13 @@ def run_tests(self, testfile):
         if command.returncode == 0:
             print(
                 "{:<30} {:<30} {:<30} {:<30}".format(
```
vsoch, line 462: Name is also something that I’d rather not rename like this

vsoch (changes requested):
Good start, but let’s please limit using buildspec for only when it’s related to the file - not the loaded recipe or name for the instance.

shahzebsiddiqui:
@vsoch i have refactored part of the code-base where applicable to use the word **Buildspec**. Hope it makes sense, i tried to do as much as i could for first attempt.

buildtest/utils/file.py
shahzebsiddiqui: I have reverted this behavior now in commit https://github.com/HPC-buildtest/buildtest-framework/pull/270/commits/2a8d6749580f2abad0c3e4c65a85b5a3d9e580e3 .

tests/menu/test_get.py
shahzebsiddiqui: Okay this is now fixed in https://github.com/HPC-buildtest/buildtest-framework/pull/270/commits/542e243641b7a88a091dac4fbaf5869fe99ccfa9

buildtest/buildsystem/base.py
shahzebsiddiqui: oh that is a typo in line ``sys.exit("BuildSpec File: %s does not exist`` This should be ``Buildspec`` the S was caps by accident. 

Yea for variables it was assumed to be all lower case. I can fix this

buildtest/buildsystem/base.py
shahzebsiddiqui: Yea i agree, i like the word ``recipe`` its the loaded Buildspec i can revert this.

buildtest/buildsystem/base.py
shahzebsiddiqui: I will fix this and change to ``self.buildspec``

buildtest/buildsystem/base.py
shahzebsiddiqui: i fixed this in commit https://github.com/HPC-buildtest/buildtest-framework/pull/270/commits/e052d3d83da9ca2200bbe16ab4c6c43208b4d27c

tests/utils/test_file.py
shahzebsiddiqui: i used ``os.path.join`` i guess the purpose of these two tests was to check shell and user expansion
with recursive directory creation. Change be seen in https://github.com/HPC-buildtest/buildtest-framework/pull/270/commits/2a8d6749580f2abad0c3e4c65a85b5a3d9e580e3#diff-a42582ff8b56b5783fdd220f2eb5ffab

tests/utils/test_file.py
shahzebsiddiqui: I can't say definitely, but knowing as sysadmin and the Xenial image provides a basic setup but it's not documented about startup configs travis https://docs.travis-ci.com/user/reference/xenial/. I'd say travis user is probably going to be bash, I have not tested this but could make some changes to ``.travis.yml`` for confirmation. 

Since test were passing it would mean these files exist.

vsoch: Did you finish this comment?

shahzebsiddiqui: That was typo in my last comment. Anyhow i didn't touch this regtest. The test will pass on most linux distros if it's a concern we can refactor test if needed. I needed something to validate with user and shell expansion. For user expansion it has to be a file that must exist and startup files are usually found as part of user creation (``useradd``)

shahzebsiddiqui:
another thought i had was to change name of ``BuildConfig`` the class is mostly validating the recipe and passing it to the ScriptBuilder. Since we are now calling it a ``Buildspec`` and not a ``config`` it should not have ``Config`` in the class name. I think a more appropriate name for the class could be to use the verb ``validate`` since that what it's doing. It has two private methods ``_validate`` and ``_validate_global``. Another option would be to use ``Parse``.

Some options could be 

- ValidateBuildspec
- ValidateSpec
- ParseBuildspec
- Parsespec

Any suggestion on the name of class?

vsoch:
I don't think this is correct. The BuildSpec exists to hold a build spec, and validation happens to be one of the functions that it does. If you want to make the name more specific, maybe something like `BuildspecParser?`

shahzebsiddiqui:
> I don't think this is correct. The BuildSpec exists to hold a build spec, and validation happens to be one of the functions that it does. If you want to make the name more specific, maybe something like `BuildspecParser?`

I am fine with BuildspecParser, would you want me to refactor this in this PR, there would be some more changes that would be made, i dont mind.

shahzebsiddiqui:
> > I don't think this is correct. The BuildSpec exists to hold a build spec, and validation happens to be one of the functions that it does. If you want to make the name more specific, maybe something like `BuildspecParser?`
> 
> I am fine with BuildspecParser, would you want me to refactor this in this PR, there would be some more changes that would be made, i dont mind.

@vsoch  okay i renamed the method to ``BuildspecParser`` in commit https://github.com/HPC-buildtest/buildtest-framework/pull/270/commits/b5afd448b1fd3309734353442f1a1c94d9e33b6e 

I feel like this looks good to me now.

shahzebsiddiqui:
@vsoch do you think this PR is good for merge?

vsoch:
I have not reviewed it finally yet, so no.

buildtest/buildsystem/base.py
```diff
@@ -1,8 +1,8 @@
 """
```
vsoch, line 1: One quick thought - why not just do BuildParser?

buildtest/buildsystem/base.py
```diff
@@ -208,31 +207,29 @@ class BuilderBase:
        any kind of builder.
     """
 
-    def __init__(self, name, recipe_config, config_file=None, testdir=None):
-        """initiate a builder base. A recipe configuration (loaded) is required.
-           this can be handled easily with the BuildConfig class:
+    def __init__(self, name, recipe, buildspec=None, testdir=None):
+        """Initiate a builder base. A recipe configuration (loaded) is required.
+           this can be handled easily with the BuildspecParser class:
```
vsoch, line 213: bc originally referred to BuildConfig so if this is now BuildParser it should be something like bp

vsoch, line 213: bc originally referred to BuildConfig so if this is now BuildParser it should be something like bp

vsoch, line 213: bc originally referred to BuildConfig so if this is now BuildParser it should be something like bp

buildtest/buildsystem/base.py
```diff
@@ -208,31 +207,29 @@ class BuilderBase:
...
         self.result = {}
         self.build_id = None
         self.metadata = {}
-        self.config_file = config_file
-        self.config_name = re.sub("[.](yml|yaml)", "", os.path.basename(config_file))
-        self.testdir = testdir or os.path.join(
-            os.getcwd(), ".buildtest", self.config_name
-        )
+        self.buildspec = buildspec
+        self.name = re.sub("[.](yml|yaml)", "", os.path.basename(buildspec))
```
vsoch: This is not correct - self.name is already set above. This is supposed to be the name of the file, not the name of the key/step.

buildtest/buildsystem/base.py
```diff
@@ -208,31 +207,29 @@ class BuilderBase:
...
+        self.testdir = testdir or os.path.join(os.getcwd(), ".buildtest", self.name)
```
vsoch: This also needs to be changed back to reference the config name.

buildtest/buildsystem/base.py
```diff
@@ -368,14 +369,14 @@ class data structures that are used by both run and dry_run
...
-        # These should all be validated for type, format, by the schema validator
+        # Metadata includes known sections in a Buildspec (pre/post_run, env)
+        # These should all be validated for type, format, by the Buildspec schema
         self.metadata = {}
         for known_section in known_sections:
             if known_section in self.recipe:
                 self.metadata[known_section] = self.recipe.get(known_section)
 
-        # Every build recipe has a shell (defaults to BUILDTEST_SHELL_DEFAULT
+        # Every Buildspec recipe has a shell (defaults to BUILDTEST_SHELL_DEFAULT)
```
vsoch, line 381: Is this true?

buildtest/buildsystem/base.py
```diff
@@ -457,13 +458,13 @@ def run_tests(self, testfile):
         if command.returncode == 0:
             print(
                 "{:<30} {:<30} {:<30} {:<30}".format(
-                    self.config_name, self.name, "PASSED", self.config_file
+                    self.name, self.name, "PASSED", self.buildspec
```
vsoch: This needs to be changed back to be the config name, then self.name, which is the key in the file.

vsoch (changes requested):

buildtest/buildsystem/base.py
```diff
@@ -457,13 +458,13 @@ def run_tests(self, testfile):
...
                 )
             )
         else:
             print(
                 "{:<30} {:<30} {:<30} {:<30}".format(
-                    self.config_name, self.name, "FAILED", self.config_file
+                    self.name, self.name, "FAILED", self.buildspec
```
vsoch: Same here....

buildtest/menu/__init__.py
```diff
@@ -88,10 +88,9 @@ def build_menu(self):
         ##################### buildtest build     ###########################
 
         parser_build.add_argument(
-            "-c",
-            "--config",
-            help="Specify test configuration file.",
-            metavar="TEST CONFIGURATION",
+            "-bs",
```
vsoch: Are you sure we want to have a “bs” flag?

buildtest/buildsystem/base.py
shahzebsiddiqui: I think ``BuildspecParser`` is more specific to say it's parsing and validating a ``Buildspec`` as pose to ``BuildParser``. If there are other objects other than ``Buildspec`` that buildtest may support in future then ``BuildParser`` would be too vague. It won't tell what it is parsing. Hope that makes sense

buildtest/buildsystem/base.py
shahzebsiddiqui: i fixed this locally.

buildtest/buildsystem/base.py
vsoch: Yes, I think that's right, I'm okay with leaving the name as is, even though it's a little long :) But what other kind of Parser do you anticipate?

buildtest/menu/__init__.py
shahzebsiddiqui: what about ``-b`` and ``--buildspec``

vsoch: I like it!

buildtest/buildsystem/base.py
shahzebsiddiqui: hmm i guess according to schema https://github.com/HPC-buildtest/schemas/blob/master/script/script-v0.0.1.schema.json#L44 it defaults to ``bash`` there is no enum on ``shell`` so user can specify anything. 

I think the comment can be a bit misleading, every Buildspec recipe doesn't need to declare a ``shell``. It defaults to ``bash`` which is ``BUILDTEST_SHELL_DEFAULT``.

buildtest/menu/__init__.py
shahzebsiddiqui: This is resolved now in latest commit.

buildtest/buildsystem/base.py
shahzebsiddiqui: This is resolved now

buildtest/buildsystem/base.py
shahzebsiddiqui: i resolved this now, changed **bc** --> **bp**

buildtest/buildsystem/base.py
shahzebsiddiqui: What i had in mind, though i dont have this thought out correctly is the following...

At the moment we are talking about a ``Buildspec`` that is a YAML file that is being parsed into a test script. At the moment buildtest does not have a concept of objects similar to ``kind`` in Kubernetes. In testing realm there could be objects such as ``Makefiles``, ``Dockerfile``, ``Singularityfile``.  A Buildspec that defines such an object would be parsed in a different way, or even delegated to a subclass of ``BuildspecParser`` such that it can validate the buildspec along with generated content such as Dockerfile, Singularityfile. Again these are bit out of scope and I dont have a clear picture how this will work out but certainly i think it adds value when user can specify what type of test object to generate. buildtest would need to parse correctly hence ``BuildParser`` would be bit too vague. 

There would be classes for each type of parser to ensure each artifact generated by buildtest is valid before executing.  I think it's bit hard to validate a test script or a job script without executing it. 
I think some basic validation for a test script would be something like

- check if shebang is present ``#!/bin/bash``
- check if test has read and execute permission (minimum ``555``)
- check if test has ``.sh``, ``.bash`` extension 

For jobscript it can be as simple as checking if there is atleast one scheduler directives in ``#SBATCH`` in the  file.

Again all of this is out of scope but something to think about

vsoch: Oh that would make sense - if there is some other config file that warrants a special parsing. +1.

buildtest/buildsystem/base.py
shahzebsiddiqui: I changed line 231 to use the name ``self.config_name`` i realized that was a mistake on my end, where ``self.name=name``  where name was passed as argument to ``BuilderBase``

shahzebsiddiqui:
@vsoch anything else pending you want me to look at, just went over most of your comments, i resolved ones that were fixed already.

buildtest/menu/build.py
```diff
@@ -10,58 +10,58 @@
...
-       buildtest build -c github.com/HPC-buildtest/tutorials/hello-world/
+       buildtest build --buildspec github.com/HPC-buildtest/tutorials/hello-world/
     """
 
-    config_files = []
+    buildspec_files = []
 
     # If no config file provided, assume discovering across buildtest/site
-    if not config_file:
+    if not buildspec:
```
vsoch, line 43: I think this config_file needs to be changed?

buildtest/menu/build.py
```diff
@@ -143,28 +146,33 @@ def func_build_subcmd(args):
...
     logger.debug(
-        f"Based on input argument: -c {args.config} buildtest discovered the following configuration {config_files}"
+        f"Based on input argument: --buildspec {args.buildspec} buildtest discovered the following Buildspecs: {buildspecs}"
     )
 
     if args.exclude:
         logger.debug(f"The exclude config pattern are the following: -e {args.exclude}")
-        config_files = [
-            config for config in config_files if include_file(config, args.exclude)
+        buildspecs = [
```
vsoch, line 163: If you use buildspecs here, I think we should be consistent and use the same variable in the previous function instead of buildspec_files.

buildtest/menu/build.py
```diff
@@ -175,15 +183,15 @@ def func_build_subcmd(args):
...
     )
     print("{:_<120}".format(""))
     # Each configuration file can have multiple tests
-    for config_file in config_files:
+    for buildspec in buildspecs:
 
-        # Read in all config files here, loading each will validate the entire file
-        bc = BuildConfig(config_file)
+        # Read in buildspec file here, loading each will validate the buildspec file
+        bc = BuildspecParser(buildspec)
```
vsoch: This should be bp not bc still.

buildtest/settings/settings.schema.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://raw.githubusercontent.com/HPC-buildtest/buildtest-framework/devel/buildtest/default.schema.json",
+  "$id": "https://raw.githubusercontent.com/HPC-buildtest/buildtest-framework/devel/buildtest/settings.schema.json",
```
vsoch, line 2: Was this changed in the schemas repo?

tests/buildsystem/test_base.py
```diff
@@ -19,12 +19,12 @@ def test_load_configs():
...
     # An empty path evaluated to be a directory should exit
     with pytest.raises(SystemExit) as e_info:
-        BuildConfig("")
+        BuildspecParser("")
 
     # Test loading config files
     for config_file in os.listdir(examples_dir):
         config_file = os.path.join(examples_dir, config_file)
-        bc = BuildConfig(config_file)
+        bc = BuildspecParser(config_file)
```
vsoch: bp

buildtest/menu/build.py
shahzebsiddiqui: fixed locally. will commit soon.

tests/utils/test_file.py
```diff
@@ -1,84 +1,53 @@
...
-    assert True is os.path.isdir(os.path.expanduser("~/x/y/z"))
-    shutil.rmtree(os.path.expandvars("$HOME/a/"))
-    shutil.rmtree(os.path.expanduser("~/x/"))
+    dir1 = os.path.join("$HOME", "a", "b", "c")
+    dir2 = os.path.join("~", "x", "y", "z")
+    create_dir(dir1)
+    create_dir(dir2)
+    assert is_dir(dir1)
+    assert is_dir(dir2)
+    shutil.rmtree(os.path.expandvars(os.path.join("$HOME", "a")))
```
vsoch: This shutil rmtree to HOME makes me VERY uncomfortable. A test should not be trying to modify anything outside of tmp_path.

buildtest/settings/settings.schema.json
shahzebsiddiqui: right now this schema is in buildtest-framework as posed to schema repo. And yes this file has been updated, you can see in my branch the file is called ``settings.schema.json`` once its merged it will be in the devel location. 
https://github.com/shahzebsiddiqui/buildtest-framework/blob/regtest_fixups/buildtest/settings/settings.schema.json

tests/buildsystem/test_base.py
shahzebsiddiqui: fixed locally will commit soon

tests/utils/test_file.py
shahzebsiddiqui: what would you like me to do in order to test shell and user expansion. Should i remove this test and just refactor this as part of ``resolve_path`` which just a simple assert without checking if ``create_dir`` can resolve path. The thing is ``create_dir`` doesn't use ``resolve_path`` because of the reason i mentioned in commit message https://github.com/shahzebsiddiqui/buildtest-framework/commit/2a8d6749580f2abad0c3e4c65a85b5a3d9e580e3 so i wanted to test this feature works.

buildtest/menu/build.py
shahzebsiddiqui: okay i this fix this locally and call them ``buildspecs`` instead of ``buildspec_files``  this makes sense.

buildtest/menu/build.py
shahzebsiddiqui: yep agreed good catch. I renamed this as ``settings_file`` since that's what we defined this based on the Terminology https://buildtest.readthedocs.io/en/devel/terminology.html.

tests/utils/test_file.py
shahzebsiddiqui: Just to clarify these lines are needed instead of calling ``resolve_path`` since non-existent file will return ``None`` from ``resolve_path``. Since input to ``create_dir`` expects a non-existent file we can't use ``resolve_path``. We can expect ``create_dir`` to be called in this case with using ``buildtest get`` or even ``buildtest build --testdir`` where directory is auto-created. 

https://github.com/shahzebsiddiqui/buildtest-framework/blob/f8e34452871d53df2ca22b8cfb75eca1977af92b/buildtest/utils/file.py#L109-L111

i did shutil.rmtree to clean up after test creation. It may not be the best way to do it and am open to a better solution so long as we test shell and user expansion.

Without this test someone using ``buildtest build --testdir $HOME`` will result in a bug.

vsoch: I think it would make sense to create an environment variable relative to the tmp_dir created by pytest, and then expand that. We should absolutely not be venturing into an actual user home and touching anything.

buildtest/menu/build.py
vsoch: Wait, so buildspec == settings_file ? I'm confused. If this is the settings_file, then why is it called buildspec?

buildtest/settings/settings.schema.json
vsoch: Can we update the schema repository before committing changes here?

shahzebsiddiqui: Recall the ``default.schema.json`` was originally part of buildtest-framework, all i did was just rename the file to comply with the correct term. Question is does it make sense to put this in schema given that only buildtest-framework will use it to validate configuration. If we put in schema repo which i don't really have any strong opinion where would we place it?

shahzebsiddiqui: Also recall this schema won't be using any versions like ``script`` and other schema types. This just for validating the buildtest setting that each site will need to do to get buildtest configured.

shahzebsiddiqui:
> Sorry for the nits - but if this namespace is now "Buildspec files" it doesn't make sense that we have a TESTCONFIG_ROOT. I can't look at those two things and determine that buildspecs can be found in TESTCONFIG_ROOT. What do you think?

Good call i think some of this needs to be renamed let me work on this. I know all of these are bit confusing but hopefully this exercise will make things more clear going forward. The one biggest confusion was the interchanging the word ``config`` to refer to the YAML and the buildtest settings file. I will fix this.

tests/utils/test_file.py
shahzebsiddiqui: Let me see if i can use ``tmp_path`` and create a sub-directory i can only do shell expansion. User expansion is out of question since ``~`` refers to home directory. I dont have a way to test that other than what i have already.

buildtest/settings/settings.schema.json
vsoch: Ah okay, then we don't need to do that.

shahzebsiddiqui:
@vsoch i renamed a few more variables in my commit https://github.com/shahzebsiddiqui/buildtest-framework/commit/a273b2f9a4d53a5e7d5f4526e6dfe61fa9552303 thanks to PyCharm Find/Replace and I ran regtest locally to ensure things were still working fine after the update.

There are still a few more changes that need to be done going through all the comments, docstring to replace the words. That's a bit extensive work and i figured i do later as we work on code.

shahzebsiddiqui:
if you there are any suggestion for the new name changes let me know the variable name and desired name you would like. I can make those changes. Hopefully now that **config**  is out of the way the code should start to make more sense.

buildtest/config.py
```diff
@@ -33,16 +33,16 @@ def init():
...
-    # Create config files, module files, and log file
-    create_config_file()
+    # Create settings files
+    create_settings_file()
 
 
-def check_configuration(config_path=None):
-    """Checks all keys in configuration file (settings/default.yml) are valid
+def check_settings(config_path=None):
+    """Checks all keys in configuration file (settings/settings.yml) are valid
```
vsoch, line 45: config_path maybe changed here too?

buildtest/config.py
```diff
@@ -54,34 +54,34 @@ def check_configuration(config_path=None):
...
 
     logger.debug(f"Validating user schema: {user_schema} with schema: {config_schema}")
     validate(instance=user_schema, schema=config_schema)
     logger.debug("Validation was successful")
 
 
-def load_configuration(config_path=None):
-    """Load the default configuration file if no argument is specified.
+def load_settings(config_path=None):
+    """Load the default settings file if no argument is specified.
```
vsoch, line 68: And here as well.

buildtest/config.py
shahzebsiddiqui: yes that makes sense i will fix this in next commit

buildtest/config.py
shahzebsiddiqui: i replaced all ``config_paths`` with ``settings_path`` in ``buildtest/config.py``. Yes there are more instance i see. Good catch

tests/utils/test_file.py
shahzebsiddiqui: okay for now i am not testing directory expansion within ``tmp_path`` I am not really sure how to go about that. 
To keep it simple i just test directory creation with a sub-directory in ``tmp_path``. See https://github.com/shahzebsiddiqui/buildtest-framework/commit/e955909d73ab4391bc2ad5864f32425e5bbf30f9

vsoch (approved):
Okay, I don't want to nit forever, this looks good enough for a first go. In the future could you please try to go about small PRs? If you need to make a list of things in the description, it's probably too big. This was incredibly challenging to review - I had to look over a very large set of changes every time, and I'm not confident that I didn't miss something (e.g., a variable that was forgotten to be renamed outside of the changes that are shown).

shahzebsiddiqui:
> Okay, I don't want to nit forever, this looks good enough for a first go. In the future could you please try to go about small PRs? If you need to make a list of things in the description, it's probably too big. This was incredibly challenging to review - I had to look over a very large set of changes every time, and I'm not confident that I didn't miss something (e.g., a variable that was forgotten to be renamed outside of the changes that are shown).

I totally understand your situation, i prefer short PRs too. Anyhow, it's done thanks for your feedback overall i like these changes. I intend to have smaller PRs going forward. :)

PR #271, opened 2020-04-12
fixing saving of output for BuildTestCommand

### LINKED ISSUE #263: Issue with BuildTestCommand poll, output and error not returning
There is an issue with logic on how BuildTestCommand is implemented. In particular I have seen that the condition shown below is reached asap and ouptut and error stream is not captured since returncode is not None hence it breaks from while loop.

https://github.com/HPC-buildtest/buildtest-framework/blob/abcfad94b69598f87f15c58e3c0f0762d9c9a99f/buildtest/utils/command.py#L63

For example when running ``buildtest build -c examples/script/python-circle.yml`` i see the return code is 0 when troubleshooting and the output and error stream is not captured.

the output file is empty as shown below

```
gitpod /workspace/buildtest-framework $ ll -l /workspace/buildtest-framework/.buildtest/python-circle/run/circle_area_04-05-2020-19-22.out
-rw-r--r-- 1 gitpod gitpod 0 Apr  5 19:22 /workspace/buildtest-framework/.buildtest/python-circle/run/circle_area_04-05-2020-19-22.out
```

The expected output of this test should be the following

```
gitpod /workspace/buildtest-framework $ python .buildtest/python-circle/circle_area.py 
Circle Radius  2
Area of circle  12.566370614359172
Block
```

This is not captured in the test output file.

vsoch:
This will fix the saving out output (out) and error (err) by the BuildTestCommand. Specifically, let's say we have a script.sh file that looks like this:

```bash
#!/bin/bash

for i in 1 2 3 4 5 6; do
    echo "LOOP ${i}"
    sleep 2
done
```
We can then create a BuildTestCommand to run it:

```python
from buildtest.utils.command import BuildTestCommand
cmd = BuildTestCommand(['/bin/bash', 'script.sh'])
```
And then execute the command:

```python
cmd.execute()
```
The out and err is now saved with the command, which was previously missing (and returned with the execute function above as a tuple)

```python
cmd.execute()                                                           
(['LOOP 1\n', 'LOOP 2\n', 'LOOP 3\n', 'LOOP 4\n', 'LOOP 5\n', 'LOOP 6\n'], [])

cmd.out                                                                 
['LOOP 1\n', 'LOOP 2\n', 'LOOP 3\n', 'LOOP 4\n', 'LOOP 5\n', 'LOOP 6\n']

cmd.err                                                                 
[]
```

This will fix #263 

### DIFF
--- a/buildtest/utils/command.py
+++ b/buildtest/utils/command.py
@@ -52,23 +52,26 @@ class BuildTestCommand:
 
         # open the process for writing
         process = subprocess.Popen(cmd, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        returncode = process.poll()
 
         # Iterate through the output
-        while True:
+        while returncode is None:
+
             out = self.decode(process.stdout.readline())
             err = self.decode(process.stderr.readline())
-
-            # If we have a return value, break
-            returncode = process.poll()
-            if returncode is not None:
-                self.returncode = returncode
-                break
 
             # Append output and error
             if out:
                 self.out.append(out)
             if err:
                 self.err.append(out)
+            returncode = process.poll()
+
+        # Get the remainder of lines, add return code
+        out, err = process.communicate()
+        self.out += ["%s\n" % x for x in self.decode(out).split("\n") if x]
+        self.err += ["%s\n" % x for x in self.decode(err).split("\n") if x]
+        self.returncode = returncode
 
         return (self.out, self.err)
 
--- a/buildtest/utils/command.py
+++ b/buildtest/utils/command.py
@@ -38,11 +38,16 @@ class BuildTestCommand:
         :return: Output and Error from shell command
         :rtype: two str objects
         """
-        # The executable must be found.
+        # Reset the output and error records
+        self.out = []
+        self.err = []
+
+        # The executable must be found, return code 1 if not
         executable = shutil.which(self.cmd[0])
         if not executable:
-            err = ["%s not found." % self.cmd[0]]
-            return (self.out, err)
+            self.err = ["%s not found." % self.cmd[0]]
+            self.returncode = 1
+            return (self.out, self.err)
 
         # remove the original executable
         args = self.cmd[1:]


vsoch:
Just to give a little explanation for the implementation - the process.poll() will be None until there is a returncode, so that's proper to place in the loop as the while condition. The stdout and stderr.readline() will return only one line, and if we break from the loop before we get all the lines we need to add the rest with process.communicate()

This is currently run in serial, but it would be fairly easy to have a BuildTestCommand that can run in parallel, and then have the self.out and self.err write (as they generate) to some output file. But let's keep it simple for now until we have a good testing example for this more advanced use case.

shahzebsiddiqui:
Okay i was just trying out a few tests not sure if this is expected to happen

Running command multiple times not the intent but output is appended each time.
```
>>> a = BuildTestCommand(["hostname"])
>>> a.execute()
(['ag-mxg-hulk090\n'], [])
>>> a.execute()
(['ag-mxg-hulk090\n', 'ag-mxg-hulk090\n'], [])
>>> a.execute()
(['ag-mxg-hulk090\n', 'ag-mxg-hulk090\n', 'ag-mxg-hulk090\n'], [])
```

Testing the error stream, i get no value from the method or class variable.

```
>>> b = BuildTestCommand(["xyz"])
>>> b.execute()
([], ['xyz not found.'])
>>> b.get_error()
[]
>>> b.err
[]
```

Also we should be able to get the return code through a method like a method ``returnCode`` that fetches ``self.returncode``

There is an issue when running invalid command return code is ``None`` so nothing get's printed

Here is an example
```
>>> b = BuildTestCommand(["xyz"])
>>> b.execute()
([], ['xyz not found.'])
>>> b.returncode
>>>

```

Return code is fetched if command is successful, so the behavior is not consistent when invalid command is passed.

```
>>> a = BuildTestCommand(["hostname"])
>>> a.execute()
(['ag-mxg-hulk090\n'], [])
>>> a.returncode
0
```

Another observation, i noticed that we can't pass multiple words in a single list element because shlex is splitting this into a list. For example this is not valid but, if i put ``-a`` as a second element then it works. Just wanted to confirm that this was your intention 

```
>>> a = BuildTestCommand(["lsb_release -a"])
>>> a.execute()
([], ['lsb_release -a not found.'])

>>> a = BuildTestCommand(["lsb_release","-a"])
>>> a.execute()
(['LSB Version:\t:core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch\n', 'Distributor ID:\tRedHatEnterpriseServer\n', 'Description:\tRed Hat Enterprise Linux Server release 7.6 (Maipo)\n', 'Release:\t7.6\n', 'Codename:\tMaipo\n'], [])

```

vsoch:
These are all bugs (that I know how to fix!) will do so tomorrow!

vsoch:
> a = BuildTestCommand(["lsb_release -a"])

This is generally incorrect usage - it either goes in as a full string (`lsb_release -a`) or a list of single elements `["lsb_release", "-a"]` but not a list with a single string.

shahzebsiddiqui:
> > a = BuildTestCommand(["lsb_release -a"])
> 
> This is generally incorrect usage - it either goes in as a full string (`lsb_release -a`) or a list of single elements `["lsb_release", "-a"]` but not a list with a single string.

Yea i figured, i just wanted to be sure that was your intent. Makes sense to me.

vsoch:
just kidding about tomorrow, the changes were quick so I wanted to get them in :) This should be ready for another look over!

shahzebsiddiqui:
@vsoch  i tested this locally and it looks much better. Thanks! 

If i understand correctly we are trying to avoid using ``shell=True`` because of security something like ``subprocess.Popen(cmd, shell=True, stderr=subprocess.PIPE, stdout=subprocess.PIPE)``

Also ``BuildTestCommand`` can be compatible to run any shell command or run a script as well ``/bin/bash script.bash``.

shahzebsiddiqui (approved):

PR #277, opened 2020-04-15
Making some general fixs

shahzebsiddiqui:
- Update docstring where parameters are missing
- Reorder imports such that they are listed in alphabetical order
- Add some more logging during configuration check, and build process.
- Make use of is_file, is_dir when applicable for checking file path as
  pose to 'os' library
- In walk_tree if directory doesn't exist, we return with empty list
- Remove sub-directory "root"  in BUILDTEST_ROOT since it's not being
  used for anything, and refactor test
- Refactor test for walk_tree, add some print statement for
  troubleshooting, and test when invalid path is set. Also walk_tree is
  not going to raise an error for invalid directory it just returns an
  empty list

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -29,7 +29,7 @@ from buildtest.defaults import (
     variable_sections,
 )
 from buildtest.utils.command import BuildTestCommand
-from buildtest.utils.file import create_dir, is_file, is_dir
+from buildtest.utils.file import create_dir, is_file, is_dir, resolve_path
 
 known_sections = variable_sections + build_sections
 
@@ -82,15 +82,14 @@ class BuildspecParser:
            :param buildspec: the pull path to the configuration file, must exist.
            :type buildspec: str, required
         """
-        self.buildspec = os.path.abspath(buildspec)
 
-        if not is_file(self.buildspec):
-            sys.exit("Buildspec File: %s does not exist." % self.buildspec)
+        self.buildspec = resolve_path(buildspec)
 
-        elif is_dir(self.buildspec):
-            sys.exit(
-                "Please provide a file path (not a directory path) to a Buildspec file."
-            )
+        if not self.buildspec:
+            sys.exit("Can't process input: %s " % buildspec)
+
+        if is_dir(self.buildspec):
+            sys.exit(f"Detected {self.buildspec} is a directory, please provide a file path (not a directory path) to BuildspecParser.")
 
         # Buildspec must pass global validation (sets self.recipe)
         self._validate_global()
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -89,7 +89,9 @@ class BuildspecParser:
             sys.exit("Can't process input: %s " % buildspec)
 
         if is_dir(self.buildspec):
-            sys.exit(f"Detected {self.buildspec} is a directory, please provide a file path (not a directory path) to BuildspecParser.")
+            sys.exit(
+                f"Detected {self.buildspec} is a directory, please provide a file path (not a directory path) to BuildspecParser."
+            )
 
         # Buildspec must pass global validation (sets self.recipe)
         self._validate_global()
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -82,6 +82,9 @@ class BuildspecParser:
            :param buildspec: the pull path to the configuration file, must exist.
            :type buildspec: str, required
         """
+        # if invalid input for buildspec
+        if not buildspec:
+            sys.exit("Invalid input type for Buildspec, must be of type 'string'.")
 
         self.buildspec = resolve_path(buildspec)
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -2,8 +2,8 @@
 BuildspecParser: loader and manager for Buildspec, and schema validation
 Copyright (C) 2020 Vanessa Sochat. 
 
-BuildspecParser is intended to read in a configuration file with one or
-more buildtest setups defined, and then generate builders based on the type 
+BuildspecParser is intended to read in a Buildspec file with one or
+more test blocks, and then generate builders based on the type
 of each. The BuilderBase is the base class for all builders that 
 expose functions to run builds.
 """
@@ -17,19 +17,19 @@ import stat
 import sys
 
 from jsonschema import validate
-from buildtest.defaults import (
-    BUILDTEST_SHELL,
-    build_sections,
-    variable_sections,
-)
-from buildtest.utils.file import create_dir
-from buildtest.utils.command import BuildTestCommand
 from buildtest.buildsystem.schemas.utils import (
     load_schema,
     load_recipe,
     get_schemas_available,
     here,
 )
+from buildtest.defaults import (
+    BUILDTEST_SHELL,
+    build_sections,
+    variable_sections,
+)
+from buildtest.utils.command import BuildTestCommand
+from buildtest.utils.file import create_dir, is_file, is_dir
 
 known_sections = variable_sections + build_sections
 
@@ -55,7 +55,8 @@ class BuildspecParser:
 
            Parameters:
 
-           buildspec: the pull path to the Buildspec file, must exist.
+           :param buildspec: the pull path to the Buildspec file, must exist.
+           :type buildspec: str, required
         """
 
         self.logger = logging.getLogger(__name__)
@@ -78,14 +79,15 @@ class BuildspecParser:
 
            Parameters:
 
-           buildspec: the pull path to the configuration file, must exist.
+           :param buildspec: the pull path to the configuration file, must exist.
+           :type buildspec: str, required
         """
         self.buildspec = os.path.abspath(buildspec)
 
-        if not os.path.exists(self.buildspec):
+        if not is_file(self.buildspec):
             sys.exit("Buildspec File: %s does not exist." % self.buildspec)
 
-        elif os.path.isdir(self.buildspec):
+        elif is_dir(self.buildspec):
             sys.exit(
                 "Please provide a file path (not a directory path) to a Buildspec file."
             )
@@ -144,8 +146,8 @@ class BuildspecParser:
     def _validate_global(self, buildspec=None):
         """The global validation ensures that the overall structure of the
            file is sound for further parsing. We load in the global.schema.json
-           for this purpose. The function also allows a custom config to be 
-           to extend the usage of the class.
+           for this purpose. The function also allows a custom Buildspec to
+           extend the usage of the class.
         """
 
         buildspec = buildspec or self.buildspec
@@ -164,6 +166,11 @@ class BuildspecParser:
         """Based on a loaded Buildspec file, return the correct builder
            for each based on the type. Each type is associated with a known 
            Builder class.
+
+           Parameters:
+
+           :param testdir: Test Destination directory, specified by --testdir
+           :type testdir: str, optional
         """
 
         builders = []
@@ -217,9 +224,14 @@ class BuilderBase:
 
            Parameters:
 
-           name: a name for the Buildspec recipe (required)
-           recipe: the loaded section from the buildspec for the user.
-           buildspec: the pull path to the Buildspec file, must exist.
+           :param name: a name for the Buildspec recipe
+           :type name: str, required
+           :param recipe: the loaded section from the buildspec for the user.
+           :type recipe: dict, required
+           :param buildspec: the pull path to the Buildspec file, must exist.
+           :type buildspec: str, optional
+           :param testdir: Test Destination directory where to write test
+           :type testdir: str, optional
         """
 
         self.name = name
@@ -378,7 +390,7 @@ class BuilderBase:
             if known_section in self.recipe:
                 self.metadata[known_section] = self.recipe.get(known_section)
 
-        # Every Buildspec recipe has a shell (defaults to BUILDTEST_SHELL_DEFAULT)
+        # Get the shell (sh, bash) for writing testscript. A Buildspec could specify this via ``shell`` key
         self.metadata["shell"] = self.get_shell()
 
         # Derive the path to the test script
@@ -419,6 +431,11 @@ class BuilderBase:
         """The shared _run function will run a test file, which must be
            provided. This is called by run() after generation of the
            test file, and it return a result object (dict).
+
+           Parameters:
+
+           :param testfile: Generated testfile by buildtest as a result of a Buildspec. This file is now executed based on the executor type
+           :type testfile: str, required
         """
 
         # Keep a result object
@@ -434,7 +451,7 @@ class BuilderBase:
         # Run the test file using the shell
         cmd = [self.get_shell(), testfile]
 
-        self.logger.debug(f"Running Test via command: {cmd}")
+        self.logger.debug(f"Running Test via command: {' '.join(cmd)}")
 
         command = BuildTestCommand(cmd)
         out, err = command.execute()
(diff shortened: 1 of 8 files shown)


vsoch:
@shahzebsiddiqui I seem to remember that Apache airflow has a test that imports are in the right order - if this is important to you, do you want to add that integration? Without it, this is 100% likely to get messed up :)

tests/test_config.py
```diff
@@ -39,5 +39,4 @@ def test_init_creation_of_buildtest_dir():
 
     shutil.rmtree(BUILDTEST_ROOT)
     init()
```
vsoch: Since we know this is in a user's home, this test needs to run to be based in tmp_path. We can specify the BUILDTEST_ROOT to be there.

tests/utils/test_file.py
```diff
@@ -57,12 +57,24 @@ def test_fail_create_dir():
...
 
-@pytest.mark.xfail(
-    reason="This test is expected to fail since we passed invalid path for directory traversal",
-    raises=BuildTestError,
-)
-def test_walk_tree_invalid_dir():
-    walk_tree("/xyz", ".py")
+def test_walk_tree_no_files(tmp_path):
+    list_of_files = walk_tree(tmp_path, ".py")
+    print(f"Detected {len(list_of_files)} .py files found in directory: {here}")
```
vsoch: Is this really testing walking a tree, if there will never be a tree in this location?

vsoch (changes requested):

shahzebsiddiqui:
> @shahzebsiddiqui I seem to remember that Apache airflow has a test that imports are in the right order - if this is important to you, do you want to add that integration? Without it, this is 100% likely to get messed up :)

I think it would be a good best practice on imports but i dont have strong opinion on having an integration check for imports.

Atleast user should know to do imports in the following order
1. ``import os``
2. ``from shutil import rmtree``
3. ``from buildtest.menu import menu``

Within each of these sections, atleast the imports should be alphabetically ordered. Atleast that is what i am trying to do. Is this even worth documenting?

What's your thoughts, do other open source projects have developer docs on the nitty aspects of code best practice. I am just worried too much docs can turn folks away and no one will bother reading or updating docs.

tests/utils/test_file.py
shahzebsiddiqui: Ok there is a typo in the print statement, i fixed locally but yes it is doing a directory traversal in tmp_path expecting to find no .py files. This is different than the other test ``test_walk_tree_invalid_dir`` which is testing if invalid directory is specified. 
```
(buildtest-framework) ssi29@ag-mxg-hulk090> pytest -vrP tests/utils/test_file.py
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.7.4, pytest-5.4.1, py-1.8.1, pluggy-0.13.1 -- /u/users/ssi29/.local/share/virtualenvs/buildtest-framework-LDt8wyRf/bin/python3.7
cachedir: .pytest_cache
rootdir: /mxg-hpc/users/ssi29/buildtest-framework
collected 8 items

tests/utils/test_file.py::test_checking_directory PASSED                                                                                                                                                [ 12%]
tests/utils/test_file.py::test_checking_file XPASS                                                                                                                                                      [ 25%]
tests/utils/test_file.py::test_directory_expansion PASSED                                                                                                                                               [ 37%]
tests/utils/test_file.py::test_create_dir PASSED                                                                                                                                                        [ 50%]
tests/utils/test_file.py::test_fail_create_dir XFAIL                                                                                                                                                    [ 62%]
tests/utils/test_file.py::test_walk_tree PASSED                                                                                                                                                         [ 75%]
tests/utils/test_file.py::test_walk_tree_no_files PASSED                                                                                                                                                [ 87%]
tests/utils/test_file.py::test_walk_tree_invalid_dir PASSED                                                                                                                                             [100%]

=================================================================================================== PASSES ====================================================================================================
_______________________________________________________________________________________________ test_walk_tree ________________________________________________________________________________________________
-------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------
Detected 3 .py files found in directory: /mxg-hpc/users/ssi29/buildtest-framework/tests/utils
___________________________________________________________________________________________ test_walk_tree_no_files ___________________________________________________________________________________________
-------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------
Detected 0 .py files found in directory: /scratch/pytest-of-ssi29/pytest-1/test_walk_tree_no_files0
_________________________________________________________________________________________ test_walk_tree_invalid_dir __________________________________________________________________________________________
-------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------
Removing directory: /scratch/pytest-of-ssi29/pytest-1/test_walk_tree_invalid_dir0 first before doing directory traversal on invalid directory
Returned following files: [] with .py extension for path: /scratch/pytest-of-ssi29/pytest-1/test_walk_tree_invalid_dir0
=================================================================================== 6 passed, 1 xfailed, 1 xpassed in 0.93s ===================================================================================
```

tests/test_config.py
shahzebsiddiqui: since ``init`` takes no argument the only way i can do this is set ``BUILDTEST_ROOT`` to ``tmp_path`` by updating variable in ``buildtest.defaults`` module so that init is using tmp_path. Do you think that what you want?

vsoch:
To be completely honest, I think it's annoying to require that. If someone has strong feelings about a particular order, they can write that on the review. But otherwise, it seems like it would just make simple contributions harder, and get in the way.

shahzebsiddiqui:
I agree with you 100%, more freedom to developers and something code reviewer need to evaluate on per PR basis. This may require some user education to new maintainers since every project operates differently and I would be in favor of being more lenient in beginning for new maintainers to encourage contribution as pose to being strict to comply with contribution standards.

tests/test_config.py
shahzebsiddiqui: ok i was able to tweak the way we do ``init`` and creation process so that regression test can work with ``tmp_path`` see https://github.com/HPC-buildtest/buildtest-framework/pull/277/commits/e1f9598e3d8e49cb12ad2606dc6ba80561165b36 More importantly we can now pass any directory path where to store configuration file. Since 'site' is a sub-directory in BUILDTEST_ROOT, we could even alter where ``buildtest get``   will clone the repo. Generally user may not want stuff cloned in $HOME due to quota limitation.

shahzebsiddiqui:
There are few other changes i'd like to make before i make i wanted to discuss. I'll do them in a subsequent PR to keep this PR simple.

I'd like to remove Buildspec file sanity check in BuildspecParser https://github.com/HPC-buildtest/buildtest-framework/blob/devel/buildtest/buildsystem/base.py#L85 since this should be handled earlier right after ``discover_buildspecs``. I don't think BuildspecParser should ever be called without any argument something like ``BuildspecParser("")`` which is done here https://github.com/HPC-buildtest/buildtest-framework/blob/devel/tests/buildsystem/test_base.py#L22

One other thought i had was should we change the option ``buildtest configs`` to ``buildtest settings`` so that it matches the term settings.

vsoch:
I disagree, in that the BuildspecParser loss the file, we have to do sanity checks there. You can’t just assume that a future user of the class will be perfect in loading the paths.

shahzebsiddiqui:
The code will not reach BuildspecParser unless there is a valid file. It will terminate early with a message and raise a System Error. It helps if these checks are done early so that BuildspecParser doesn't have to handle these checks and just focus on validating the schema.

This removes any doubt later down the code on file check since BuildspecParser will invoke a subclass like ScriptBuilder and all of these changes would get propagated. This leads to redundant checks which are not required.

vsoch:
Yes but it's a class provided by Buildtest, there is no reason that a user can't implement one on the fly. Why are you so focused on removing _error checking_ ??

shahzebsiddiqui:
> Yes but it's a class provided by Buildtest, there is no reason that a user can't implement one on the fly. Why are you so focused on removing _error checking_ ??

I am not following, what is the problem we are trying to solve with having the check in ``BuildspecParser``. Or put it this way what input parameter to buildtest will cause the code to reach file check in BuildspecParser.

shahzebsiddiqui:
Anyhow, do you have any feedback on this PR in it's current state.  I didn't want to add many changes in one PR.

vsoch:
Let's say I'm a user. I import a BuildspecParser to use it outside of the buildtest client. This can't be allowed:

```
BuildspecParser("oops-I_dont_exist.yml")
```
This is _basic_ error checking. You don't remove it.

buildtest/buildsystem/base.py
```diff
@@ -78,14 +79,15 @@ def load(self, buildspec):
...
+           :type buildspec: str, required
         """
         self.buildspec = os.path.abspath(buildspec)
 
-        if not os.path.exists(self.buildspec):
+        if not is_file(self.buildspec):
             sys.exit("Buildspec File: %s does not exist." % self.buildspec)
 
-        elif os.path.isdir(self.buildspec):
+        elif is_dir(self.buildspec):
```
vsoch: This change doesn't make sense - if you just checked with is_file this statement will never be reached. The original check was intended to first check if the path exists, and then parse if it's a directory or not (hitting the second statement). Now this is unreachable, unless there is some niche case of a symlink not being identified as a file and then hitting the next one.

buildtest/config.py
```diff
@@ -3,42 +3,48 @@
...
     # check if $HOME/.buildtest exists, if not create directory
-    if not os.path.exists(BUILDTEST_ROOT):
-        print(
-            f"Creating buildtest configuration directory: \
-                 {BUILDTEST_ROOT}"
-        )
-        os.mkdir(BUILDTEST_ROOT)
+    if not is_dir(dirname):
+        create_dir(dirname)
+        msg = f"Creating buildtest settings directory: {dirname}"
```
vsoch, line 39: This seems redundant - if you want to print a message to the user and logger, use logger.info.

shahzebsiddiqui: okay i can fix this

buildtest/buildsystem/base.py
shahzebsiddiqui: Yea I see your point, the elif wont be reached i was testing this locally just for my sanity and hope we can adopt these methods as much as we can. 

is_file works with shell expansion and works by running ``os.path.isfile`` underneath
```
>>> is_file("$HOME/.buildtest/settings.yml")
True
>>> is_file("$HOME/.buildtest/")
False
```

If a directory is specified ``is_file`` will return ``False`` and the ``if not is_file`` will cause statment to be ``True`` and never reach the elif.

The thing is os.path.exists wouldn't work for some edge cases such as 

```
>>> os.path.exists("$HOME/.buildtest/settings.yml")
False
>>> is_file("$HOME/.buildtest/settings.yml")
True
>>> os.path.exists("$HOME/.buildtest/")
False
>>> is_dir("$HOME/.buildtest/")
True
```
We also account for tilde ``~`` if passed as input but os.path.exists won't catch this

```
>>> is_file("~/.buildtest/settings.yml")
True
>>> os.path.exists("~/.buildtest/settings.yml")
False
```
Symlinks work with os.path.exists so that's not an issue. It's only for discovery and exclusion we get the real path in-order for exclusion to work correctly.
The reason why i changed os.path.exists to ``is_file`` or ``is_dir`` depending on context of file is to account for tilde or shell expansion
```
>>> os.path.islink("/u/users/ssi29/gpfs")
True
>>> resolve_path("/u/users/ssi29/gpfs")
'/mxg-hpc/users/ssi29'
```

Note that ``is_file`` and ``is_dir`` are calling ``resolve_path`` before the check.

shahzebsiddiqui:
> Let's say I'm a user. I import a BuildspecParser to use it outside of the buildtest client. This can't be allowed:
> 
> ```
> BuildspecParser("oops-I_dont_exist.yml")
> ```
> 
> This is _basic_ error checking. You don't remove it.

I see your point, i did not anticipate this, i guess it would be helpful to call ``resolve_path`` on the input file or directory in the init constructor and check for file or directory existence before doing anything else.

buildtest/buildsystem/base.py
vsoch: So under what conditions would the second elif trigger?

shahzebsiddiqui: Yea i fixed this now in latest commit and let me demo this 

If input is a directory

```
>>> spec = BuildspecParser("$PWD/examples/benchmarks/")
Detected /mxg-hpc/users/ssi29/buildtest-framework/examples/benchmarks is a directory, please provide a file path (not a directory path) to BuildspecParser.
```

If its invalid file

```
>>> spec = BuildspecParser("xyz!@#")
Can't process input: xyz!@#
```

If its valid file we proceed as normal

```
>>> spec = BuildspecParser("$PWD/examples/benchmarks/stream.yml")
>>> spec.recipe
{'version': '0.0.1', 'stream_uniprocess_c': {'type': 'script', 'description': 'STREAM Microbenchmark C test on uniprocessor', 'run': 'wget https://www.cs.virginia.edu/stream/FTP/Code/stream.c\ngcc -O2 stream.c -o stream\n./stream\n'}, 'stream_openmp_c': {'type': 'script', 'description': 'STREAM Microbenchmark C Test with OpenMP', 'run': 'wget https://www.cs.virginia.edu/stream/FTP/Code/stream.c\ngcc -fopenmp -D_OPENMP -O2 stream.c -o stream\n\nexport OMP_NUM_THREAD=2\n./stream\n\nexport OMP_NUM_THREAD=4\n./stream\n\nexport OMP_NUM_THREAD=8\n./stream\n\nexport OMP_NUM_THREAD=16\n./stream\n'}}

```

buildtest/buildsystem/base.py
```diff
@@ -82,15 +82,14 @@ def load(self, buildspec):
            :param buildspec: the pull path to the configuration file, must exist.
            :type buildspec: str, required
         """
-        self.buildspec = os.path.abspath(buildspec)
 
-        if not is_file(self.buildspec):
-            sys.exit("Buildspec File: %s does not exist." % self.buildspec)
+        self.buildspec = resolve_path(buildspec)
```
vsoch, line 89: The check if buildspec is None needs to go _before_ this line - if buildspec here is None, it's going to spit out an error when you do expandvars in resolve_path.

vsoch (changes requested):

buildtest/buildsystem/base.py
vsoch: E.g.,

```python
import os                                                                                                 
# This is the first line of resolve_path
path = os.path.expandvars(None)                                                                           
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-4154076a32c3> in <module>
----> 1 path = os.path.expandvars(None)

~/anaconda3/lib/python3.7/posixpath.py in expandvars(path)
    286     """Expand shell variables of form $var and ${var}.  Unknown variables
    287     are left unchanged."""
--> 288     path = os.fspath(path)
    289     global _varprog, _varprogb
    290     if isinstance(path, bytes):

TypeError: expected str, bytes or os.PathLike object, not NoneType
```

shahzebsiddiqui: Okay i made the change and tested this locally to ensure it hits all three conditionals. 

```
>>> spec = BuildspecParser(None)
Invalid input type for Buildspec, must be of type 'string'.
>>> spec = BuildspecParser("$PWD/invalid/file")
Can't process input: $PWD/invalid/file
>>> spec = BuildspecParser("$HOME/.buildtest")
Detected /u/users/ssi29/.buildtest is a directory, please provide a file path (not a directory path) to BuildspecParser.

```

shahzebsiddiqui: @vsoch  i also added two more checks in ``test_base.py`` in commit https://github.com/shahzebsiddiqui/buildtest-framework/commit/b37a93a98f38f6e7728dca154a834bd5b7b7d93f to make sure we hit those conditionals in coverage report.

tests/utils/test_file.py
```diff
@@ -57,12 +57,24 @@ def test_fail_create_dir():
...
-@pytest.mark.xfail(
-    reason="This test is expected to fail since we passed invalid path for directory traversal",
-    raises=BuildTestError,
-)
-def test_walk_tree_invalid_dir():
-    walk_tree("/xyz", ".py")
+def test_walk_tree_no_files(tmp_path):
+    list_of_files = walk_tree(tmp_path, ".py")
+    print(f"Detected {len(list_of_files)} .py files found in directory: {tmp_path}")
+    assert 0 == len(list_of_files)
```
vsoch, line 67: This test could be improved to actually test that it's walking a directory (e.g., os.listdir()) would have worked here, there is no tree walking. But this is a nit, and otherwise the PR looks good.

vsoch (approved):

PR #278, opened 2020-04-17
Update settings schema field $id and title

shahzebsiddiqui:
This PR must be merged after merging https://github.com/HPC-buildtest/schemas/pull/26

### DIFF
--- a/buildtest/settings/settings.schema.json
+++ b/buildtest/settings/settings.schema.json
@@ -1,7 +1,7 @@
 {
-  "$id": "https://raw.githubusercontent.com/HPC-buildtest/buildtest-framework/devel/buildtest/settings.schema.json",
+  "$id": "https://HPC-buildtest.github.io/schemas/settings/settings.schema.json",
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "Buildtest Configuration Global Schema",
+  "title": "Buildtest Settings Schema",
   "type": "object",
   "required": ["editor", "executors"],
   "additionalProperties": false,


vsoch:
An unrelated question while we are waiting on the other PR - any thought / updates about the name for the organization? I think this should be decided before some newer release in the next few months.

vsoch (approved):

shahzebsiddiqui:
> An unrelated question while we are waiting on the other PR - any thought / updates about the name for the organization? I think this should be decided before some newer release in the next few months.

I was hoping to get namespace **buildtest** https://github.com/buildtest but there has been no response from the group, emailed all three of the folks. 

Unrelated to the org, i was thinking about name for the ``schema``  repo could we name it **buildtest-api** or **buildspec-api** or **api**. Since the test are called Buildspec, i figured it would make sense to name it either **Buildspec-API**  but it contains schema for buildtest as well.

vsoch:
I think it contains schemas, so we should call it schemas :) It's under the buildtest (err what should be buildtest) org so the association is implied.

vsoch:
To give feedback on "buildspec-api" - an api refers to an "application programming interface" which usually comes down to a server that is _serving endpoints that might deliver json_. This is an entirely different concept than a schema, which might be the rules for defining how that json is structured. It would be very misleading to call a repo of schemas an api.

shahzebsiddiqui:
fair enough, according to Github they can't do much i asked them twice and this was their response. We can't wait indefinitely i dont think these three users will ever reply if it's a miracle. We don't have a trademark so Github can't help. 

Even **buildtests** is taken https://github.com/buildtests 👎 

image

**buildtesters** is available. Unfortunately namespace are not case sensitive otherwise we could have done **Buildtest** but that is not possible either.

vsoch:
Any other ideas?

vsoch:
What about `building` ?

shahzebsiddiqui:
So org name doesn't have to imply change to command ``buildtest`` or should it? 
What about ``buildtester`` or ``buildertest`` I am kinda bad at names. I liked the name ``buildtest`` since it's meant to automate building of tests and is very generic. Since it's meant for HPC at that time i couldn't think of anything so i named org ``HPC-buildtest``. But now this also supports HPC Software and System Testing so maybe the org name could be anything. Generally i see the org name is name of the opensource tool but that's not always the case. I think the name should be unique, short and easily searched.  I think the name of org should also be done in sync with the logo so there is a relevant logo to the name of the org.


This may imply changes to docs, but prefer the url

vsoch:
See issue #280 I put a lot of thought there! I like buildspec :)

PR #283, opened 2020-04-19
Refactor builder base

shahzebsiddiqui:


vsoch:
I don't understand how this PR relates to uuid work? What is being undone here?

shahzebsiddiqui:
> I don't understand how this PR relates to uuid work? What is being undone here?

I renamed the branch locally and pushed as new branch it has the previous commits found in  https://github.com/buildntest/buildtest/pull/281

With squash it won't really matter.

shahzebsiddiqui:
I didn't want to start from ``devel`` and apply the other changes so the commit history has some work done with uuid so it can be ignored as this doesn't have anything to do with uuid. 

The PR refactors Builderbase  in regards to how we use testdir.

shahzebsiddiqui:
@vsoch when you get time, can you please review and provide feedback. Kindly let me know at your earliest convenience.

vsoch:
Please address https://github.com/buildntest/buildtest/pull/283#issuecomment-616174661 and describe what exactly this PR is changing, and why.

shahzebsiddiqui:
> Please address [#283 (comment)](https://github.com/buildntest/buildtest/pull/283#issuecomment-616174661) and describe what exactly this PR is changing, and why.

Yea so what happen is i did a ``git checkout -b`` from HEAD of PR https://github.com/buildntest/buildtest/pull/281 to carry over the changes to create new PR since i didn't want to lose my changes but undid my change for uuid in commit https://github.com/buildntest/buildtest/pull/283/commits/2e185891c242a5cd5c2ceba6cf17448ba1d32233 because it was apparent this required a bit more work  mentioned in https://github.com/buildntest/buildtest/pull/281#issuecomment-616002676 hence i closed the PR.

This PR does some generic fix to BuilderBase which i figured we fix it.

vsoch:
That's not what I meant, I meant to describe the changes _to the code_ that you are making here and why.

shahzebsiddiqui:
> That's not what I meant, I meant to describe the changes _to the code_ that you are making here and why.

Ahh, sorry i misunderstood. So this code does the following
1. We don't need method ``_get_testdir`` since we can retrieve this via ``self.metadata["testpath"]``
2. We don't need declare a temporary variable when we can access ``self.metadata["testpath"]`` or ``self.testdir``. 
3. There were some locations where we had a local variable like ``testfile`` passed to ``run_tests`` and return value from ``_write.test`` both of these are not needed since we can reference ``self.metadata["testpath"]``. The behavior is still the same just that we didn't need unnecessary variables and passing them to methods where we can use class variable.

buildtest/buildsystem/base.py
```diff
@@ -427,19 +420,14 @@ def run(self):
...
         # Create test directory and run folder they don't exist
         self._create_test_folders()
-        testfile = self._write_test()
-        result = self.run_tests(testfile)
+        self._write_test()
+        result = self.run_tests()
         return result
 
-    def run_tests(self, testfile):
+    def run_tests(self):
```
vsoch: So the issue that I have here is that the client (before this change) is flexible to be updated with a different test file to run. That's actually the entire point - you can instantiate it, and then update the file at runtime without needing to start over. I would prefer to keep this ability because I can see it being useful for some custom implementation / use case in the future.

buildtest/buildsystem/base.py
```diff
@@ -449,11 +437,12 @@ def run_tests(self, testfile):
...
         # Change to the test directory
-        os.chdir(self._get_testdir())
-        self.logger.debug(f"Changing to directory {self._get_testdir()}")
+        os.chdir(self.testdir)
+
+        self.logger.debug(f"Changing to test directory {self.testdir}")
 
         # Run the test file using the shell
-        cmd = [self.get_shell(), testfile]
+        cmd = [self.get_shell(), self.metadata["testpath"]]
```
vsoch: If you are going to continually reference self.metadata["testpath"], and removing self._get_testpath, a much more elegant solution is to create a property that essentially returns the same thing, e.g.,

```python
@property
def testpath(self, testpath=None):
        """Based on the testfile path, return the testing directory.
           Returns: full path to testing directory
        """
       return self.metadata.get("testpath")
```
You could also combine this sort of thing with being able to parse the original testpath. Anyway, the way you address the issue is up to you, this is a separate suggestion to make reference of the variable much cleaner than self.metadata every time.

vsoch (changes requested):

vsoch:
Also please make sure the urlchecker is passing as well for review!

buildtest/buildsystem/base.py
shahzebsiddiqui: I see your point, i didn't realize you intended to import the BuilderBase class directly. It makes sense, ill fix this.

shahzebsiddiqui:
> Also please make sure the urlchecker is passing as well for review!

I found the issue, seems like CodeFactor changed its link due to org name its now https://www.codefactor.io/repository/github/buildntest/buildtest i guess we can update the org name to ``buildtesters`` now and i can fix the link these urls.

buildtest/buildsystem/base.py
```diff
@@ -539,6 +540,7 @@ def _write_test(self, lines=None):
         self.logger.debug(
             f"Applying permission 755 to {self.metadata['testpath']} so that test can be executed"
         )
+        return self.metadata['testpath']
```
vsoch: What is this return added for?

buildtest/buildsystem/base.py
```diff
@@ -274,23 +275,15 @@ def __str__(self):
...
-        return os.path.dirname(self.metadata["testpath"])
-
     def _create_test_folders(self):
         """Create all needed test folders on init, and add their paths
            to self.metadata.
         """
 
-        testdir = self._get_testdir()
-        create_dir(testdir)
+        create_dir(self.testdir)
```
vsoch, line 283: Didn't you remove this? This needs to be a function (property)

buildtest/buildsystem/base.py
shahzebsiddiqui: this was me undoing my change it's already part of ``devel`` https://github.com/buildntest/buildtest/blob/devel/buildtest/buildsystem/base.py#L554

buildtest/buildsystem/base.py
shahzebsiddiqui: Are you referring to the suggestion you mentioned in https://github.com/buildntest/buildtest/pull/283#discussion_r412489796 

Yes i removed ``self._get_testdir()`` it didn't make sense to me initially, If i understand you would like to retain this feature again?

vsoch: Instead of referencing self.metadata["testdir"] I suggested making a @property function so the user can reference `self.testdir`. I think I even wrote it out for you.

shahzebsiddiqui: Yes i got that, i think we have 2 identical ways to retrieve testdir and it's bit confusing why this was the case and i don't like it either. 

1. Class variable self.testdir on line 249 https://github.com/shahzebsiddiqui/buildtest-framework/blob/192fa17b61b96875ce8a491b903d60580ab01bae/buildtest/buildsystem/base.py#L249


2. self.metadata["testpath"], self.metadata["testdir"] on line 398 https://github.com/shahzebsiddiqui/buildtest-framework/blob/refactor_BuilderBase/buildtest/buildsystem/base.py#L398-399

It's bit redundant having self.testdir, self.metadata["testdir"] and i dont see a purpose of having both.

User can still pass a testdir to BuilderBase class and retrieve the test directory via class variable ``testdir``. So is adding a method ``testpath`` going to solve anything. I am fine btw, of referencing ``self.metadata['testpath']`` to keep things simple.

```
a = BuilderBase(name, recipe, testdir=/path/to/testdir)
a.testdir
```

I am not so familiar with @property, so i'd like to understand what is the end-goal with introducing this method.

vsoch: I prefer the first. When was the self.metadata ["testdir"] added? That doesn't make sense.

vsoch: https://www.freecodecamp.org/news/python-property-decorator/

PR #286, opened 2020-04-22
Status check for test and utility methods read_file and write_file

shahzebsiddiqui:
This PR successfully implements the status key implementation in the framework related to schema change addressed in https://github.com/buildtesters/schemas/pull/24 

The commit message along with comments should explain the logic on how i implemented the test_state (PASS/FAIL) which is used to count the number of pass and failed test.

I have ran three test, one with status check using returncode, one with regex and one using both. 

The three tests were committed to tutorials repo see commit https://github.com/buildtesters/tutorials/commit/5406b84e048557fb2d81b4de0febe847ca862164 

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -b system/systemd.yml -b system/selinux.yml  -b apps/openacc/vecadd.yml

            Discovered Buildspecs

/mxg-hpc/users/ssi29/tutorials/apps/openacc/vecadd.yml
/mxg-hpc/users/ssi29/tutorials/system/systemd.yml
/mxg-hpc/users/ssi29/tutorials/system/selinux.yml



vecadd                         vecadd_gnu                     PASS                           /mxg-hpc/users/ssi29/tutorials/apps/openacc/vecadd.yml
systemd                        systemd_default_target         PASS                           /mxg-hpc/users/ssi29/tutorials/system/systemd.yml
selinux                        selinux_disable                PASS                           /mxg-hpc/users/ssi29/tutorials/system/selinux.yml


                        Test summary
Executed 3 tests
Passed Tests: 3/3 Percentage: 100.000%
Failed Tests: 0/3 Percentage: 0.000%
```

Here is a snippet from the logs to give idea whats going on

**selinux.yml** - This one performs both regex and return code check

```
2020-04-22 11:44:32,122 [base.py:535 - run_tests() ] - [DEBUG] Return code: 0 for test: /mxg-hpc/users/ssi29/tutorials/.buildtest/selinux/selinux_disable.sh
2020-04-22 11:44:32,122 [base.py:555 - run_tests() ] - [DEBUG] Conducting Return Code check
2020-04-22 11:44:32,122 [base.py:558 - run_tests() ] - [DEBUG] Status Return Code: 0   Result Return Code: 0
2020-04-22 11:44:32,122 [base.py:567 - run_tests() ] - [DEBUG] Conducting Regular Expression check
2020-04-22 11:44:32,122 [base.py:471 - check_regex() ] - [DEBUG] Detected regex stream 'stdout' so reading output file: /mxg-hpc/users/ssi29/tutorials/.buildtest/selinux/run/selinux_disable_04-22-2020-11-44.out
2020-04-22 11:44:32,123 [base.py:482 - check_regex() ] - [DEBUG] Applying re.search with exp: ^SELINUX=disabled$
2020-04-22 11:44:32,123 [base.py:572 - run_tests() ] - [INFO] ReturnCode Match: True Regex Match: True

```

**systemd.yml** - this performs return code check only

```
2020-04-22 11:44:32,077 [base.py:513 - run_tests() ] - [DEBUG] Running Test via command: /bin/bash /mxg-hpc/users/ssi29/tutorials/.buildtest/systemd/systemd_default_target.sh
2020-04-22 11:44:32,094 [base.py:530 - run_tests() ] - [DEBUG] Writing run output to file: /mxg-hpc/users/ssi29/tutorials/.buildtest/systemd/run/systemd_default_target_04-22-2020-11-44.out
2020-04-22 11:44:32,095 [base.py:534 - run_tests() ] - [DEBUG] Writing run error to file: /mxg-hpc/users/ssi29/tutorials/.buildtest/systemd/run/systemd_default_target_04-22-2020-11-44.err
2020-04-22 11:44:32,095 [base.py:535 - run_tests() ] - [DEBUG] Return code: 0 for test: /mxg-hpc/users/ssi29/tutorials/.buildtest/systemd/systemd_default_target.sh
2020-04-22 11:44:32,095 [base.py:555 - run_tests() ] - [DEBUG] Conducting Return Code check
2020-04-22 11:44:32,095 [base.py:558 - run_tests() ] - [DEBUG] Status Return Code: 0   Result Return Code: 0
2020-04-22 11:44:32,095 [base.py:572 - run_tests() ] - [INFO] ReturnCode Match: True Regex Match: True
2020-04-22 11:44:32,095 [base.py:70 - __init__() ] - [DEBUG] buildtest found the available schema: 
```

**vecadd.yml** - this performs regex check only

```
2020-04-22 11:44:31,692 [base.py:513 - run_tests() ] - [DEBUG] Running Test via command: /bin/bash /mxg-hpc/users/ssi29/tutorials/.buildtest/vecadd/vecadd_gnu.sh
2020-04-22 11:44:32,068 [base.py:530 - run_tests() ] - [DEBUG] Writing run output to file: /mxg-hpc/users/ssi29/tutorials/.buildtest/vecadd/run/vecadd_gnu_04-22-2020-11-44.out
2020-04-22 11:44:32,069 [base.py:534 - run_tests() ] - [DEBUG] Writing run error to file: /mxg-hpc/users/ssi29/tutorials/.buildtest/vecadd/run/vecadd_gnu_04-22-2020-11-44.err
2020-04-22 11:44:32,069 [base.py:535 - run_tests() ] - [DEBUG] Return code: 0 for test: /mxg-hpc/users/ssi29/tutorials/.buildtest/vecadd/vecadd_gnu.sh
2020-04-22 11:44:32,069 [base.py:567 - run_tests() ] - [DEBUG] Conducting Regular Expression check
2020-04-22 11:44:32,069 [base.py:471 - check_regex() ] - [DEBUG] Detected regex stream 'stdout' so reading output file: /mxg-hpc/users/ssi29/tutorials/.buildtest/vecadd/run/vecadd_gnu_04-22-2020-11-44.out
2020-04-22 11:44:32,069 [base.py:482 - check_regex() ] - [DEBUG] Applying re.search with exp: ^final result: 1.000000$
2020-04-22 11:44:32,069 [base.py:572 - run_tests() ] - [INFO] ReturnCode Match: True Regex Match: True

```

Finally, if status is not defined buildtest works as normal, here is an example run without any status check. We are running the test https://github.com/buildtesters/tutorials/blob/master/apps/serial/hello.yml

```
(buildtest-framework) ssi29@ag-mxg-hulk090> buildtest build -b apps/serial/hello.yml

            Discovered Buildspecs

/mxg-hpc/users/ssi29/tutorials/apps/serial/hello.yml



hello                          fortan_hello                   PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/hello.yml
hello                          c_hello                        PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/hello.yml
hello                          cplusplus_hello                PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/hello.yml


                        Test summary
Executed 3 tests
Passed Tests: 3/3 Percentage: 100.000%
Failed Tests: 0/3 Percentage: 0.000%

```

I have tested locally by tweaking returncode and regex to see if PASS/FAIL is working and it was successful. Feel free to test this by cloning the ``tutorials`` repo and set ``WORK`` environment to parent directory where you cloned tutorials and it should work.

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -447,10 +447,7 @@ class BuilderBase:
             :rtype: bool
         """
 
-        if recipe_returncode == result_returncode:
-            return True
-
-        return False
+        return recipe_returncode == result_returncode
 
     def check_regex(self, regex):
         """ This method conducts a regular expression check using 're.search' with regular
@@ -555,21 +552,22 @@ class BuilderBase:
             # if no regex is defined we set this to True since we do a logical AND
             regex_match = True
 
-            # check returncode from result matches value defined in Buildspec recipe, self.check_returncode returns a bool
             if "returncode" in status:
                 self.logger.debug("Conducting Return Code check")
                 self.logger.debug(
                     "Status Return Code: %s   Result Return Code: %s"
                     % (status["returncode"], result["RETURN_CODE"])
                 )
+                # self.check_returncode checks if test returncode matches returncode specified in Buildspec.
+                # The return value is a bool
                 returncode_match = self.check_returncode(
                     status["returncode"], result["RETURN_CODE"]
                 )
 
-            # check regex expression in Buildspec with output or error stream. self_check_regex returns a boolean (True/False)
-            # by using re.search to check expression
             if "regex" in status.keys():
                 self.logger.debug("Conducting Regular Expression check")
+                # self.check_regex  applies regular expression check specified in Buildspec with output or error
+                # stream. self.check_regex returns a boolean (True/False) by using re.search
                 regex_match = self.check_regex(status["regex"])
 
             self.logger.info(
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -476,13 +476,12 @@ class BuilderBase:
             content = read_file(self.metadata["errfile"])
 
         # convert list to string
-        content = "\n".join(content)
-        self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
-        # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
-        if re.search(regex["exp"], content):
-            return True
+        # content = "\n".join(content)
 
-        return False
+        self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
+
+        # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
+        return re.search(regex["exp"], content) != None
 
     def run_tests(self, testfile):
         """The shared _run function will run a test file, which must be
@@ -540,7 +539,7 @@ class BuilderBase:
 
         status = self.recipe.get("status")
 
-        test_state = ""
+        test_state = "FAIL"
 
         # if status is defined in Buildspec, then check for returncode and regex
         if status:
@@ -564,7 +563,7 @@ class BuilderBase:
                     status["returncode"], result["RETURN_CODE"]
                 )
 
-            if "regex" in status.keys():
+            if "regex" in status:
                 self.logger.debug("Conducting Regular Expression check")
                 # self.check_regex  applies regular expression check specified in Buildspec with output or error
                 # stream. self.check_regex returns a boolean (True/False) by using re.search
@@ -577,15 +576,12 @@ class BuilderBase:
 
             if returncode_match and regex_match:
                 test_state = "PASS"
-            else:
-                test_state = "FAIL"
 
         # if status is not defined we check test returncode, by default 0 is PASS and any other return code is a FAIL
         else:
             if command.returncode == 0:
                 test_state = "PASS"
-            else:
-                test_state = "FAIL"
+
 
         # this variable is used later when counting all the pass/fail test in buildtest/menu/build.py
         result["TEST_STATE"] = test_state
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -582,7 +582,6 @@ class BuilderBase:
             if command.returncode == 0:
                 test_state = "PASS"
 
-
         # this variable is used later when counting all the pass/fail test in buildtest/menu/build.py
         result["TEST_STATE"] = test_state
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -526,7 +526,7 @@ class BuilderBase:
         self.metadata["errfile"] = run_output_file + ".err"
 
         # write output of test to .out file
-        write_file(self.metadata["outfile"],out)
+        write_file(self.metadata["outfile"], out)
 
         self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
 
@@ -553,7 +553,7 @@ class BuilderBase:
             regex_match = True
 
             # check returncode from result matches value defined in Buildspec recipe, self.check_returncode returns a bool
-            if "returncode" in status.keys():
+            if "returncode" in status:
                 self.logger.debug("Conducting Return Code check")
                 self.logger.debug(
                     "Status Return Code: %s   Result Return Code: %s"
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -433,12 +433,40 @@ class BuilderBase:
         return result
 
     def check_returncode(self, recipe_returncode, result_returncode):
+        """ This method checks if return code from Buildspec recipe matches
+            return code from test. The return type is a bool either 'True'/'False'
+            depending if the return codes match
+
+            Parameters:
+
+            :param recipe_returncode: return code specified in recipe
+            :type recipe_returncode: int, required
+            :param result_returncode: return code from test
+            :type result_returncode: int, required
+            :return: A boolean return True/False based on conditional equality
+            :rtype: bool
+        """
+
         if recipe_returncode == result_returncode:
             return True
 
         return False
 
+
     def check_regex(self, regex):
+        """ This method conducts a regular expression check using 're.search' with regular
+            expression defined in Buildspec. User must specify an output stream (stdout, stderr)
+            to select when performing regex. In buildtest, this would read the .out or .err file
+            based on stream and run the regular expression to see if there is a match.
+
+            Parameters:
+
+            :param regex: Regular expression object defined in Buildspec file
+            :type regex: str, required
+            :return:  A boolean return True/False based on if re.search is successful or not
+            :rtype: bool
+        """
+
         if regex["stream"] == "stdout":
             self.logger.debug(
                 f"Detected regex stream 'stdout' so reading output file: {self.metadata['outfile']}"
@@ -458,6 +486,7 @@ class BuilderBase:
             return True
 
         return False
+
 
     def run_tests(self, testfile):
         """The shared _run function will run a test file, which must be
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -526,14 +526,17 @@ class BuilderBase:
         self.metadata["errfile"] = run_output_file + ".err"
 
         # write output of test to .out file
-        write_file(self.metadata["outfile"], out)
 
+        out = "\n".join(out)
+        err = "\n".join(err)
+
+        write_file(self.metadata["outfile"], out)
         self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
 
         # write error from test to .err file
-        write_file(self.metadata["errfile"], out)
-
+        write_file(self.metadata["errfile"], err)
         self.logger.debug(f"Writing run error to file: {self.metadata['errfile']}")
+
         self.logger.debug(f"Return code: {command.returncode} for test: {testfile}")
         result["RETURN_CODE"] = command.returncode
         result["END_TIME"] = self.get_formatted_time("end_time")
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -475,9 +475,6 @@ class BuilderBase:
             )
             content = read_file(self.metadata["errfile"])
 
-        # convert list to string
-        # content = "\n".join(content)
-
         self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
 
         # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
@@ -526,12 +523,12 @@ class BuilderBase:
         out = "\n".join(out)
         err = "\n".join(err)
 
-        write_file(self.metadata["outfile"], out)
         self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
+        write_file(self.metadata["outfile"], out)
 
         # write error from test to .err file
-        write_file(self.metadata["errfile"], err)
         self.logger.debug(f"Writing run error to file: {self.metadata['errfile']}")
+        write_file(self.metadata["errfile"], err)
 
         self.logger.debug(f"Return code: {command.returncode} for test: {testfile}")
         result["RETURN_CODE"] = command.returncode
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -29,7 +29,7 @@ from buildtest.defaults import (
     variable_sections,
 )
 from buildtest.utils.command import BuildTestCommand
-from buildtest.utils.file import create_dir, is_file, is_dir, resolve_path
+from buildtest.utils.file import create_dir, is_dir, resolve_path, read_file, write_file
 
 known_sections = variable_sections + build_sections
 
@@ -470,15 +470,16 @@ class BuilderBase:
             self.logger.debug(
                 f"Detected regex stream 'stdout' so reading output file: {self.metadata['outfile']}"
             )
-            with open(self.metadata["outfile"], "r") as fd:
-                content = fd.read()
+            content = read_file(self.metadata["outfile"])
+
         elif regex["stream"] == "stderr":
             self.logger.debug(
                 f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}"
             )
-            with open(self.metadata["errfile"], "r") as fd:
-                content = fd.read()
+            content = read_file(self.metadata["errfile"])
 
+        # convert list to string
+        content = "\n".join(content)
         self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
         # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
         if re.search(regex["exp"], content):
@@ -525,12 +526,13 @@ class BuilderBase:
         self.metadata["errfile"] = run_output_file + ".err"
 
         # write output of test to .out file
-        with open(self.metadata["outfile"], "w") as fd:
-            fd.write("\n".join(out))
+        write_file(self.metadata["outfile"],out)
+
         self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
+
         # write error from test to .err file
-        with open(self.metadata["errfile"], "w") as fd:
-            fd.write("\n".join(err))
+        write_file(self.metadata["errfile"], out)
+
         self.logger.debug(f"Writing run error to file: {self.metadata['errfile']}")
         self.logger.debug(f"Return code: {command.returncode} for test: {testfile}")
         result["RETURN_CODE"] = command.returncode
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -436,24 +436,28 @@ class BuilderBase:
         if recipe_returncode == result_returncode:
             return True
 
-        return  False
+        return False
 
-    def check_regex(self,regex):
+    def check_regex(self, regex):
         if regex["stream"] == "stdout":
-            self.logger.debug(f"Detected regex stream 'stdout' so reading output file: {self.metadata['outfile']}")
+            self.logger.debug(
+                f"Detected regex stream 'stdout' so reading output file: {self.metadata['outfile']}"
+            )
             with open(self.metadata["outfile"], "r") as fd:
                 content = fd.read()
         elif regex["stream"] == "stderr":
-            self.logger.debug(f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}")
+            self.logger.debug(
+                f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}"
+            )
             with open(self.metadata["errfile"], "r") as fd:
                 content = fd.read()
 
         self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
         # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
-        if re.search(regex["exp"],content):
-            return  True
+        if re.search(regex["exp"], content):
+            return True
 
-        return  False
+        return False
 
     def run_tests(self, testfile):
         """The shared _run function will run a test file, which must be
@@ -494,19 +498,18 @@ class BuilderBase:
         self.metadata["errfile"] = run_output_file + ".err"
 
         # write output of test to .out file
-        with open(run_output_file + ".out", "w") as fd:
+        with open(self.metadata["outfile"], "w") as fd:
             fd.write("\n".join(out))
-        self.logger.debug(f"Writing run output to file: {run_output_file+'.out'}")
+        self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
         # write error from test to .err file
-        with open(run_output_file + ".err", "w") as fd:
+        with open(self.metadata["errfile"], "w") as fd:
             fd.write("\n".join(err))
-        self.logger.debug(f"Writing run error to file: {run_output_file + '.err'}")
+        self.logger.debug(f"Writing run error to file: {self.metadata['errfile']}")
         self.logger.debug(f"Return code: {command.returncode} for test: {testfile}")
         result["RETURN_CODE"] = command.returncode
         result["END_TIME"] = self.get_formatted_time("end_time")
 
         status = self.recipe.get("status")
-
 
         test_state = ""
 
@@ -523,8 +526,13 @@ class BuilderBase:
             # check returncode from result matches value defined in Buildspec recipe, self.check_returncode returns a bool
             if "returncode" in status.keys():
                 self.logger.debug("Conducting Return Code check")
-                self.logger.debug("Status Return Code: %s   Result Return Code: %s" % (status["returncode"], result["RETURN_CODE"]))
-                returncode_match = self.check_returncode(status["returncode"], result["RETURN_CODE"])
+                self.logger.debug(
+                    "Status Return Code: %s   Result Return Code: %s"
+                    % (status["returncode"], result["RETURN_CODE"])
+                )
+                returncode_match = self.check_returncode(
+                    status["returncode"], result["RETURN_CODE"]
+                )
 
             # check regex expression in Buildspec with output or error stream. self_check_regex returns a boolean (True/False)
             # by using re.search to check expression
@@ -532,20 +540,24 @@ class BuilderBase:
                 self.logger.debug("Conducting Regular Expression check")
                 regex_match = self.check_regex(status["regex"])
 
-            self.logger.info("ReturnCode Match: %s Regex Match: %s " % (returncode_match, regex_match))
+            self.logger.info(
+                "ReturnCode Match: %s Regex Match: %s "
+                % (returncode_match, regex_match)
+            )
 
             if returncode_match and regex_match:
                 test_state = "PASS"
             else:
                 test_state = "FAIL"
-        # if status is not defined we check reference returncode for PASS is 0,
+
+        # if status is not defined we check test returncode, by default 0 is PASS and any other return code is a FAIL
         else:
             if command.returncode == 0:
                 test_state = "PASS"
             else:
                 test_state = "FAIL"
 
-        #this variable is used later when counting all the pass/fail test in buildtest/menu/build.py
+        # this variable is used later when counting all the pass/fail test in buildtest/menu/build.py
         result["TEST_STATE"] = test_state
 
         print(
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -82,6 +82,7 @@ class BuildspecParser:
            :param buildspec: the pull path to the configuration file, must exist.
            :type buildspec: str, required
         """
+
         # if invalid input for buildspec
         if not buildspec:
             sys.exit("Invalid input type for Buildspec, must be of type 'string'.")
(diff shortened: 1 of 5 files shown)


buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,60 @@ def run(self):
...
+            :type regex: str, required
+            :return:  A boolean return True/False based on if re.search is successful or not
+            :rtype: bool
+        """
+
+        if regex["stream"] == "stdout":
+            self.logger.debug(
+                f"Detected regex stream 'stdout' so reading output file: {self.metadata['outfile']}"
+            )
+            with open(self.metadata["outfile"], "r") as fd:
```
vsoch: Would it be worth to have a utils. read_file and write_file to be used here? They are relatively simple functions, but would be one line instead of two :)

buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,60 @@ def run(self):
...
+
+            Parameters:
+
+            :param regex: Regular expression object defined in Buildspec file
+            :type regex: str, required
+            :return:  A boolean return True/False based on if re.search is successful or not
+            :rtype: bool
+        """
+
+        if regex["stream"] == "stdout":
```
vsoch, line 466: Can you be sure "stream" is a field of regex? What if it's 1. not there, or 2. there but not matching?

buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,60 @@ def run(self):
...
+        elif regex["stream"] == "stderr":
+            self.logger.debug(
+                f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}"
+            )
+            with open(self.metadata["errfile"], "r") as fd:
+                content = fd.read()
+
+        self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
+        # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
+        if re.search(regex["exp"], content):
```
vsoch: This one too - how are we sure this is provided?

buildtest/buildsystem/base.py
```diff
@@ -466,31 +521,77 @@ def run_tests(self, testfile):
         # Keep an output file
         run_output_file = os.path.join(self.metadata.get("rundir"), self.build_id)
 
+        self.metadata["outfile"] = run_output_file + ".out"
+        self.metadata["errfile"] = run_output_file + ".err"
+
         # write output of test to .out file
-        with open(run_output_file + ".out", "w") as fd:
+        with open(self.metadata["outfile"], "w") as fd:
```
vsoch: We could use the utils functions here too :) And it would be good to add tests.

buildtest/buildsystem/base.py
```diff
@@ -466,31 +521,77 @@ def run_tests(self, testfile):
...
+
+            # returncode_match is boolean to check if reference returncode matches return code from test
+            returncode_match = True
+
+            # regex_match is boolean to check if output/error stream matches regex defined in Buildspec,
+            # if no regex is defined we set this to True since we do a logical AND
+            regex_match = True
+
+            # check returncode from result matches value defined in Buildspec recipe, self.check_returncode returns a bool
+            if "returncode" in status.keys():
```
vsoch: You don't need "keys" here, you can just say if key in dict.

vsoch (changes requested):
Looking good! I addressed some issues with the function, and specifically let's:
 - add util functions for writing / reading files with tests
 - can we add tests for these return statuses / what you showed?

Just to verify in some of the output you showed - there was an __init__ function showed in the output for the second that wasn't shown for the first and third. Was this just an extra copy paste line or some actual difference / error? (related to the schema found but seemed like an incomplete message).

buildtest/buildsystem/base.py
shahzebsiddiqui: If i understand you want two method in buildtest/utils/file.py ``read_file`` and ``write_file``. that makes sense, 1 one line is better than 2. I can get this implemented in this PR.

buildtest/buildsystem/base.py
shahzebsiddiqui: I made sure ``stream`` and ``exp`` are required https://github.com/buildtesters/schemas/blob/master/script/script-v0.0.1.schema.json#L73 
Also enum set for ``stream`` so its not possible https://github.com/buildtesters/schemas/blob/master/script/script-v0.0.1.schema.json#L69
And this is also tested in the following

https://github.com/buildtesters/schemas/blob/master/.github/tests/invalid/script/0.0.1/invalid_script_v0.0.1.yml#L38
https://github.com/buildtesters/schemas/blob/master/.github/tests/invalid/script/0.0.1/invalid_script_v0.0.1.yml#L47

That should clarify any doubts :)

buildtest/buildsystem/base.py
vsoch: Yes! You've used it twice in this PR, so it makes sense to add. You can probably grep for other instances of open( to find other occurrences too.

buildtest/buildsystem/base.py
vsoch: Awesome! Yes, clarified.

buildtest/buildsystem/base.py
shahzebsiddiqui: I think this is resolved with my response in https://github.com/buildtesters/buildtest/pull/286#discussion_r413178872 

There is a test for this too https://github.com/buildtesters/schemas/blob/master/.github/tests/invalid/script/0.0.1/invalid_script_v0.0.1.yml#L47

I believe the ``validate`` method will complain if this key is not present. Right now i checked if ``stream`` is passed without ``exp`` i didn't check the other way around but i take it that would be redundant test for ``required``.

buildtest/utils/file.py
```diff
@@ -143,3 +143,80 @@ def resolve_path(path):
...
+        Parameters:
+
+        :param filepath: file name to read
+        :type filepath: str, required
+        :return: return content of file
+        :rtype: list
+    """
+
+    # type check filepath to ensure its a string, if not return None
+    if not isinstance(filepath,str):
```
vsoch: I think this will need black formatting, should be a space after filepath.

buildtest/utils/file.py
```diff
@@ -143,3 +143,80 @@ def resolve_path(path):
...
+    # resolve_path will handle shell and user expansion and account for any symlinks and check for file existence.
+    # if resolve_path does not return gracefully it implies file does not exist and will return None
+    filepath = resolve_path(filepath)
+
+    # if it's invalid file let's return None
+    if not filepath:
+        return None
+
+    content = []
+    fd = open(filepath, "r")
```
vsoch: This is overly complicated - we just need:
```
with open(filepath, 'r') as fd:
    content = fd.read()
```
This parsing of lines doesn't make sense. If we wanted to read lines we could do readline() but it's better to return the raw content of the file and let the parsing function decide what to do. For example, if we used this and piped into a function to parse yaml, it would be all kinds of weird to get a list of strings.

buildtest/utils/file.py
```diff
@@ -143,3 +143,80 @@ def resolve_path(path):
...
+        :return: return content of file
+        :rtype: list
+    """
+
+    # type check filepath to ensure its a string, if not return None
+    if not isinstance(filepath,str):
+        return None
+
+    # if content was passed as a string, let's convert to list
+    if isinstance(content, str):
```
vsoch: This also doesn't make sense - just take the content given from the user, and check that it's a string, and write it to the file. If the user wants newlines, etc., this is up for him/her to pass.

buildtest/buildsystem/base.py
shahzebsiddiqui: Good call, i fixed this now in https://github.com/shahzebsiddiqui/buildtest-framework/commit/1ba7ae328775f96715281dfb79887b4b195e6dad.

buildtest/utils/file.py
```diff
@@ -143,3 +143,80 @@ def resolve_path(path):
...
+        content = content.splitlines(True)
+
+    # shown below is equivalent to running 'resolve_path' but we can't invoke it since we expect this method
+    # to write to a new file. resolve_path assumes filepath already exists and returns realpath otherwise returns
+    # None.
+    filepath = os.path.expanduser(filepath)
+    filepath = os.path.expandvars(filepath)
+    filepath = os.path.realpath(filepath)
```
vsoch: This is also not scoped to the function - we simply need to check if it's a string, and if the directory name to write it to exists, and then write it :) These functions should be 3-6 lines at most!

buildtest/utils/file.py
```diff
@@ -143,3 +143,80 @@ def resolve_path(path):
...
+
+    # if filepath is an actual file, let's not write to file and return None.
+    # also if filepath is a directory let's also return None
+    if is_file(filepath) or is_dir(filepath):
+        return None
+    
+    fd = open(filepath, "w")
+    # process each line and write to file
```
vsoch: Don't forget that it's best practice to use with:

vsoch (changes requested):

buildtest/utils/file.py
shahzebsiddiqui: ahhh, i was kinda following the principles of read and write by line similar to ``BuildTestCommand`` . If you wanted a simple as that, i could have saved some time :(
Would you like the return type for ``read_file`` be a string or a list.

vsoch: Sorry for the misunderstanding! I literally meant changing those 2 lines into a function since we re-use, and adding a check (for read) that the file exists first, and (for write) that the directory to write exists first.

Generally, we don't expect special / custom parsing, so you'd want to use read() and just return raw content, same idea for write.

buildtest/utils/file.py
shahzebsiddiqui: that's strange ill fix this, i have pre-commit hook enabled but i run into these issues. Don't worry if you see a failure on black. Ill fix those in subsequent commits if it happens.

buildtest/utils/file.py
shahzebsiddiqui: just to clarify we should only accept ``str`` type to ``write_file`` when passing the content of file. If it's not a string we can return ``None``

vsoch: I think it's fine to assume that a string is passed, and let the error be raised if something else is. I would just check that the filename exists for read, and that the dirname exists for write.

buildtest/utils/file.py
shahzebsiddiqui: ok if i understand i will create two methods ``read`` and ``write`` with the suggestion you proposed.

buildtest/utils/file.py
shahzebsiddiqui: So i initially allowed an input of list because ``BuildTestCommand.execute`` returns output, error as a ``list``. These are then invoked to ``write_file``.
Just to confirm you would want input content to be in ``string``, if it's list would you want ``write_test`` to do nothing or convert to string and try to write file. Btw, i initially assumed input could be string or list so i convert it.

vsoch: `write_file("\n".join(output))`

shahzebsiddiqui: Unrelated to this, but i was testing locally something is wrong with ``BuildTestCommand`` it was working before now it's only getting first line of file and i can reproduce this on the client. The input content ``out`` and ``err`` are invoked via ``write_file`` so i was trying to figure out if this was an issue with my function, seems like it's coming from ``BuildTestCommand``

```
>>> from buildtest.utils.command import BuildTestCommand
>>> cmd = BuildTestCommand("python .buildtest/python-circle/circle_area.py")
>>> out,err = cmd.execute()
>>> out
['Circle Radius  2\n']

>>> os.system("python .buildtest/python-circle/circle_area.py")
Circle Radius  2
Area of circle  12.566370614359172
Block
0

```

buildtest/utils/file.py
```diff
@@ -143,3 +143,76 @@ def resolve_path(path):
 
     if os.path.exists(real_path):
         return real_path
+
+
+def read(filepath):
```
vsoch: Please call this "read_file" and then "write_file."

vsoch (changes requested):

buildtest/utils/file.py
shahzebsiddiqui: This issue is present in ``devel`` branch, ill open an issue basically its reading first line and stopping.

buildtest/utils/file.py
shahzebsiddiqui: this is fixed now in https://github.com/shahzebsiddiqui/buildtest-framework/commit/1fbe34ca33da9f65698e68e153a7315491a02979

buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,61 @@ def run(self):
...
+
+            :param recipe_returncode: return code specified in recipe
+            :type recipe_returncode: int, required
+            :param result_returncode: return code from test
+            :type result_returncode: int, required
+            :return: A boolean return True/False based on conditional equality
+            :rtype: bool
+        """
+
+        if recipe_returncode == result_returncode:
```
vsoch: You can just return `recipe_returncode == result_returncode`

buildtest/buildsystem/base.py
```diff
@@ -466,31 +522,81 @@ def run_tests(self, testfile):
...
+                self.logger.debug("Conducting Return Code check")
+                self.logger.debug(
+                    "Status Return Code: %s   Result Return Code: %s"
+                    % (status["returncode"], result["RETURN_CODE"])
                 )
+                returncode_match = self.check_returncode(
+                    status["returncode"], result["RETURN_CODE"]
+                )
+
+            # check regex expression in Buildspec with output or error stream. self_check_regex returns a boolean (True/False)
```
vsoch: this line / documentation should go with the check_regex function.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: got it, i fixed this locally.

buildtest/buildsystem/base.py
shahzebsiddiqui: fixed locally.

buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,58 @@ def run(self):
...
+            self.logger.debug(
+                f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}"
+            )
+            content = read_file(self.metadata["errfile"])
+
+        # convert list to string
+        content = "\n".join(content)
+        self.logger.debug(f"Applying re.search with exp: {regex['exp']}")
+        # perform a regex search based on value of 'exp' key defined in Buildspec with content file (output or error)
+        if re.search(regex["exp"], content):
```
vsoch: This could also be shortened to return `re.search(..) != None` if you want to use fewer lines :)

buildtest/buildsystem/base.py
```diff
@@ -466,31 +519,82 @@ def run_tests(self, testfile):
...
         result["END_TIME"] = self.get_formatted_time("end_time")
 
-        # Print the test result for the user
-        if command.returncode == 0:
-            print(
-                "{:<30} {:<30} {:<30} {:<30}".format(
-                    self.config_name, self.name, "PASSED", self.buildspec
+        status = self.recipe.get("status")
+
+        test_state = ""
```
vsoch: Any reason to not have it start as FAIL up here, and then just update to PASS given passing conditions? Right now there is an if -> else flow, and the else always sets it to FAIL - why not just start like that? A default state is to FAIL until there is evidence of passing. That will also simplify some of the logic below.

buildtest/buildsystem/base.py
```diff
@@ -466,31 +519,82 @@ def run_tests(self, testfile):
...
+
+            self.logger.info(
+                "ReturnCode Match: %s Regex Match: %s "
+                % (returncode_match, regex_match)
             )
+
+            if returncode_match and regex_match:
+                test_state = "PASS"
+            else:
+                test_state = "FAIL"
```
vsoch: If this is default (FAIL), you don't need the else: fail. It will already be set and you just need to set to be success given the first condition. It's the same for line ~587.

buildtest/utils/file.py
```diff
@@ -143,3 +143,66 @@ def resolve_path(path):
...
+
+        :param filepath: file name to read
+        :type filepath: str, required
+        :return: return content of file
+        :rtype: list
+    """
+
+    # type check filepath to ensure its a string, if not return None
+    if not isinstance(filepath, str):
+        return None
```
vsoch: You have the return type here a list? I would also add a note that returning None indicates some problem with the file. Is there any reason not to exit on error given the return is None? We want to make sure that any Nones returned are properly handled.

buildtest/utils/file.py
```diff
@@ -143,3 +143,66 @@ def resolve_path(path):
...
+
+        :param filepath: file name to write
+        :type filepath: str, required
+        :param content: content to write to file
+        :type content: str, required
+        :return: return content of file
+        :rtype: list
+    """
+
+    # type check filepath to ensure its a string, if not return None
```
vsoch: You can combine these first two conditions into one if --> return statement.

buildtest/utils/file.py
```diff
@@ -143,3 +143,66 @@ def resolve_path(path):
...
+    if not isinstance(filepath, str):
+        return None
+
+    # ensure content is of type string
+    if not isinstance(content, str):
+        return None
+
+    # if filepath is an actual file, let's not write to file and return None.
```
vsoch: This isn't expected behavior - I would want to give a write file function a file (that already might exist) and have it over-written. If you are concerned about this, then there could be a check in the calling function, but it shouldn't be here with the util (and then silently not update the file that exists, that sounds like a bad silent bug to me!)

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: good call, I see your point, set **test_state="FAIL"** and remove the else conditions. Ill fix this.

shahzebsiddiqui:
Okay, i am having hard time testing all of these feedbacks given that BuildTestCommand is not working #287.  The ``read_file`` and ``write_file`` is not something i'd like to refactor now, it's bit hard to understand what exactly how you want these methods to operate and it would take some time for me to properly check all the regtest. I'd suggest we come to resolution and merge as is and not have endless back-forth.

I am unable to properly test the regex stream with all the edge case, but I got the SELINUX test https://github.com/buildtesters/tutorials/blob/master/system/selinux.yml partly because it was one line of output. I tried another test for vecAdd https://github.com/buildtesters/tutorials/blob/master/apps/openacc/vecadd.yml for just doing regex match and this didnt work because output file was empty which should not be the case.

vsoch:
>  I'd suggest we come to resolution and merge as is and not have endless back-forth.

This back and forth is part of the review process. If you ask me to review a PR, I am going to give it my best review, and not merge based on some subjective thing like the number of comments we've already had.

If you are having trouble testing, then I'd suggest working on that issue first before coming back here. The solution is not to disregard testing and merge anyway. :/

buildtest/buildsystem/base.py
shahzebsiddiqui: okay i believe this is fixed in https://github.com/shahzebsiddiqui/buildtest-framework/commit/b4750fdab1dd4367d448f7c51dd21e3098c3e92d

buildtest/utils/file.py
shahzebsiddiqui: this is checking is to ensure ``filepath``   is of type string. I could with System Error on this if that makes sense.

buildtest/utils/file.py
shahzebsiddiqui: I am not following, you want me to change the comment or are you referring to the two conditions

```
    # type check filepath to ensure its a string, if not return None
    if not isinstance(filepath, str):
        return None

    # ensure content is of type string
    if not isinstance(content, str):
        return None
```

vsoch:
```python
if not isinstance(filepath, str) and not isinstance(content, str):
    return None
```

buildtest/utils/file.py
vsoch: That makes sense, yes.

shahzebsiddiqui:
Understood, I will try my best to address your feedback, but i can't read your thoughts so please be patient if i dont get it right the first time.

buildtest/utils/file.py
shahzebsiddiqui: I changed the behavior with meaningful exception messages, i kept them separate so i can get coverage for each if condition with the appropriate test.

vsoch:
@shahzebsiddiqui I only respond when you comment (and I am notified by email) so if you don't want me to respond again, just don't post a comment for me to respond to (or explicitly say in it that no feedback / response is needed).

vsoch:
btw, I get a LOT of these notifications (this is only from April 10th!) so if you want to be more concise and decrease that number, it would be greatly appreciated :)

image

vsoch:
What I'll do if you want less discussion is wait until you explicitly ask for review. That means I can make a lot of comments and it will only trigger one email. Cheers!

buildtest/utils/file.py
shahzebsiddiqui: this is resolved but with a SystemError message see https://github.com/shahzebsiddiqui/buildtest-framework/commit/e4ad6f2188b00e085e92d94b225b8587e1c03cdf

shahzebsiddiqui:
What i do is click **Resolve Conversation** once it has been done, i find that effective way to track what needs to be done. Sorry for all those notifications, i get the same amount of emails too. 

Anyhow, i made some changes ``read_file`` and ``write_file``  to not return None but an error message.

buildtest/utils/file.py
shahzebsiddiqui: Okay i removed this check so you can overwrite to file.  This is added in https://github.com/shahzebsiddiqui/buildtest-framework/commit/e4ad6f2188b00e085e92d94b225b8587e1c03cdf#diff-b94bc4259405fb65c0cd50bf50dc3bc1L200

vsoch:
I didn't mean to complain, I was responding to:

> Understood, I will try my best to address your feedback, but i can't read your thoughts so please be patient if i dont get it right the first time.

Because you must infer that my responses are "me not being patient" but they are in fact "me responding to a notification." I have a lot of patience and I will not want to merge anything before it is properly finished, tested, and ready.

shahzebsiddiqui:
@vsoch this is pending your review. The only unresolved conversation that needs your attention are regarding ``read_file`` and ``write_file``. Once this is complete i think we can resolve this task.

buildtest/buildsystem/base.py
```diff
@@ -431,6 +432,57 @@ def run(self):
...
+            )
+            content = read_file(self.metadata["outfile"])
+
+        elif regex["stream"] == "stderr":
+            self.logger.debug(
+                f"Detected regex stream 'stderr' so reading error file: {self.metadata['errfile']}"
+            )
+            content = read_file(self.metadata["errfile"])
+
+        # convert list to string
```
vsoch: This should be removed, generally it’s not good practice to leave commented code.

buildtest/buildsystem/base.py
```diff
@@ -466,31 +518,78 @@ def run_tests(self, testfile):
...
         # write output of test to .out file
-        with open(run_output_file + ".out", "w") as fd:
-            fd.write("\n".join(out))
-        self.logger.debug(f"Writing run output to file: {run_output_file+'.out'}")
+
+        out = "\n".join(out)
+        err = "\n".join(err)
+
+        write_file(self.metadata["outfile"], out)
+        self.logger.debug(f"Writing run output to file: {self.metadata['outfile']}")
```
vsoch, line 526: You’d want this line before attempting the write in case something goes wrong it will appear in the log.

buildtest/buildsystem/base.py
```diff
@@ -466,31 +518,78 @@ def run_tests(self, testfile):
...
+
         # write error from test to .err file
-        with open(run_output_file + ".err", "w") as fd:
-            fd.write("\n".join(err))
-        self.logger.debug(f"Writing run error to file: {run_output_file + '.err'}")
+        write_file(self.metadata["errfile"], err)
+        self.logger.debug(f"Writing run error to file: {self.metadata['errfile']}")
```
vsoch, line 530: Same here.

buildtest/buildsystem/base.py
```diff
@@ -466,31 +518,78 @@ def run_tests(self, testfile):
...
+            regex_match = True
+
+            if "returncode" in status:
+                self.logger.debug("Conducting Return Code check")
+                self.logger.debug(
+                    "Status Return Code: %s   Result Return Code: %s"
+                    % (status["returncode"], result["RETURN_CODE"])
                 )
+                # self.check_returncode checks if test returncode matches returncode specified in Buildspec.
+                # The return value is a bool
```
vsoch, line 558: Why is there an entire function to check for equality of two values (one line)?

buildtest/utils/file.py
```diff
@@ -143,3 +138,62 @@ def resolve_path(path):
...
+
+        Parameters:
+
+        :param filepath: file name to read
+        :type filepath: str, required
+        :return: return content of file
+        :rtype: str
+    """
+
+    # type check filepath to ensure its a string, if not return None
```
vsoch: Docstring here needs update.

buildtest/utils/file.py
```diff
@@ -143,3 +138,62 @@ def resolve_path(path):
...
+
+        Parameters:
+
+        :param filepath: file name to write
+        :type filepath: str, required
+        :param content: content to write to file
+        :type content: str, required
+    """
+
+    # type check filepath to ensure its a string, if not return None
```
vsoch: Docstring needs update.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: I foresee we can invoke or override method in future for example slurm job codes would need to be retrieved first before comparing with value in Buildspec. Some of these methods in BuilderBase could be better suited to be defined in ExecutorBase  class.

vsoch: I think this method would make sense only in the case that a return code was retrieved internally. Since we have to provide both values, there is no benefit /use to doing this as the user will always need to provide those two values (and just as easily could compare them.) For your use case that you mention, I would suggest having the check for success happen on the level of a specific kind of executor (e.g., slurm). If the function is to be meaningful here, it should use something internally. If this is truly needed / to be overridden by a custom executor, that should be added when the executor is written, and not before hand (now.)

shahzebsiddiqui:
Okay i made all the changes for now please check again.

vsoch (approved):
If you are ok with decrease in coverage, I have differences in opinion on some of the changes, but it passes the bar for basic review, so feel free to merge.

shahzebsiddiqui:
thanks for review @vsoch i'll merge this for now.

PR #289, opened 2020-04-27
Fix/command out err

vsoch:
I'm unable to reproduce this error, but logically based on the general execution envronment that @shahzebsiddiqui is describing, I suspect there is an issue with identifying the newline character and so this is the first thing I'd want to try for the fix.

> When stdout or stderr are pipes and universal_newlines is True then all line endings will be converted to '\n' as described for the universal newlines 'U' mode argument to open().

The fix is not to revert back to a less secure method, but to properly debug the actual issue. This will close #287

Issue #284, opened 2020-04-19
resurrect buildtest build report


shahzebsiddiqui:
In ``devel`` we removed ``buildtest build report`` and it's reporting features that are still present in ``master`` branch. See https://buildtest.readthedocs.io/en/latest/build_subcommand/build_status.html#build-report-buildtest-build-report for details.

We first need to capture every build and metadata associated to Buildspec and produced output, error, log file into a json file. Maybe we have one json file per build or have one json file for all builds. Once we have the data captured, we should have a command line option to retrieve each build or test id.

vsoch:
I think the output structure, and lock file functionality needs to be developed before this. You can look at snakemake source code if you need ideas.

vsoch:
I can’t find the thread now because I have like 1000 emails from you, but for the shebang is there any reason that we can’t always use /usr/bin/env followed by whatever shall they choose? And then of course honor that for the actual execution?

shahzebsiddiqui:
This is now resolved with `buildtest report` https://github.com/buildtesters/buildtest/pull/322

PR #290, opened 2020-04-28
Refactor BuilderBase to account for changes to shell, shebang key in script schema

shahzebsiddiqui:
sync script schema from https://github.com/buildtesters/schemas/blob/master/script/script-v0.0.1.schema.json
modify implementation on shell detection, environment variable, and the
command to execute the test. We can now support passing shell options on
the command line. The shebang can be customized via 'shebang' key if
it's not found it will use the default shebang based on shell type. If
it's python it will resort to path of 'python' wrapper.
When running in dry mode we were getting output header of table of
results that were not suppose to show up.
Add more log details for BuilderBase

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -500,10 +500,8 @@ class BuilderBase:
         os.chdir(self._get_testdir())
         self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
-        # command to run the test
-        cmd = []
         # build the run command that includes the shell path, shell options and path to test file
-        cmd += [self.shell.path, self.shell.opts, testfile]
+        cmd = [self.shell.path, self.shell.opts, testfile]
         self.metadata["command"] = " ".join(cmd)
         self.logger.debug(f"Running Test via command: {self.metadata['command']}")
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -281,7 +281,9 @@ class BuilderBase:
 
         # if shell program (bash, sh, python) is not present in system raise an error
         if not shutil.which(self.shell_name):
-            raise BuildTestError(f"Can't find executable {self.shell_name}, please check your $PATH or install the appropriate package")
+            raise BuildTestError(
+                f"Can't find executable {self.shell_name}, please check your $PATH or install the appropriate package"
+            )
 
         self.shebang = self.recipe.get("shebang")
 
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -247,7 +247,8 @@ class BuilderBase:
             os.getcwd(), ".buildtest", self.config_name
         )
         self.logger = logging.getLogger(__name__)
-
+        self.logger.debug(f"Processing Buildspec: {self.buildspec}")
+        self.logger.debug(f"Processing Buildspec section: {self.name}")
         # A builder is required to define the type attribute
         if not hasattr(self, "type"):
             sys.exit(
@@ -265,6 +266,36 @@ class BuilderBase:
                 "Mismatch in type. Builder expects %s but found %s."
                 % (self.type, self.recipe.get("type"))
             )
+
+        self.shell = self.recipe.get("shell")
+        # if 'shell' key not found in Buildspec, let's set it to BUILDTEST_SHELL and log a message
+        if not self.shell:
+            self.logger.debug(
+                f"shell not found in Buildspec, setting shell to {BUILDTEST_SHELL} "
+            )
+            self.shell = BUILDTEST_SHELL
+
+        self.shell_name = self.shell.split()[0]
+        self.shell_opts = self.shell.split()[1:]
+
+        self.shebang = self.recipe.get("shebang")
+        # if shebang is not found in recipe let's set it to #!/bin/bash
+        if not self.shebang:
+            # when shell: python we set shebang to location of 'python' wrapper
+            if "python" in self.shell_name:
+                self.shebang = "#!" + shutil.which("python")
+            # otherwise we set to #!/bin/bash
+            else:
+                self.shebang = "#!" + BUILDTEST_SHELL
+
+            self.logger.debug(
+                f"shebang not found in Buildspec, setting shebang to {self.shebang} "
+            )
+
+        self.logger.debug(f"shell: {self.shell}")
+        self.logger.debug(f"shell name: {self.shell_name}")
+        self.logger.debug(f"shell opts: {self.shell_opts}")
+        self.logger.debug(f"shebang: {self.shebang}")
 
     def __str__(self):
         return "[builder-%s-%s]" % (self.type, self.name)
@@ -302,10 +333,11 @@ class BuilderBase:
            :rtype: str
         """
 
-        shell = self.get_shell()
-        if "python" in shell:
+        if "python" in self.shell_name:
+            self.logger.debug("Setting test extension to 'py'")
             return "py"
 
+        self.logger.debug("Setting test extension to 'sh'")
         return "sh"
 
     def get_environment(self):
@@ -319,8 +351,7 @@ class BuilderBase:
 
         env = []
         pairs = self.recipe.get("env")
-        shell = self.get_shell()
-
+        shell = self.shell_name
         # If we are using a python shell, we need to import os first
         if "python" in shell:
             env.append("import os")
@@ -393,7 +424,8 @@ class BuilderBase:
                 self.metadata[known_section] = self.recipe.get(known_section)
 
         # Get the shell (sh, bash) for writing testscript. A Buildspec could specify this via ``shell`` key
-        self.metadata["shell"] = self.get_shell()
+        self.metadata["shell"] = self.shell_name
+        self.metadata["shell_opts"] = self.shell_opts
 
         # Derive the path to the test script
         self.metadata["testpath"] = "%s.%s" % (
@@ -498,8 +530,14 @@ class BuilderBase:
         os.chdir(self._get_testdir())
         self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
-        # Run the test file using the shell
-        cmd = [self.get_shell(), testfile]
+        # build the command to run the test-script
+        cmd = []
+        # add the shell (/bin/bash, /bin/sh, sh, bash, python)
+        cmd += [self.shell_name]
+        # add shell opts if set
+        if self.shell_opts:
+            cmd += self.shell_opts
+        cmd += [testfile]
 
         self.logger.debug(f"Running Test via command: {' '.join(cmd)}")
 
@@ -653,13 +691,8 @@ class BuilderBase:
            :rtype: list
         """
 
-        lines = []
-        shell = shutil.which(self.get_shell())
-
-        if not shell:
-            shell = BUILDTEST_SHELL
-
-        lines += [f"#!{shell}"]
+        # start of each test should have the shebang
+        lines = [self.shebang]
 
         # Add environment variables
         lines += self.get_environment()
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -28,6 +28,7 @@ from buildtest.defaults import (
 from buildtest.exceptions import BuildTestError
 from buildtest.utils.command import BuildTestCommand
 from buildtest.utils.file import create_dir, is_dir, resolve_path, read_file, write_file
+from buildtest.utils.shell import Shell
 
 known_sections = variable_sections + build_sections
 
@@ -267,43 +268,19 @@ class BuilderBase:
                 "Mismatch in type. Builder expects %s but found %s."
                 % (self.type, self.recipe.get("type"))
             )
+        # The default shell will be bash
+        shell = Shell()
+        # if 'shell' key  is defined in Buildspec let's override the default shell
+        if self.recipe.get("shell"):
+            shell = Shell(self.recipe.get("shell"))
 
-        self.shell = self.recipe.get("shell")
-        # if 'shell' key not found in Buildspec, let's set it to BUILDTEST_SHELL and log a message
-        if not self.shell:
-            self.logger.debug(
-                f"shell not found in Buildspec, setting shell to {BUILDTEST_SHELL} "
-            )
-            self.shell = BUILDTEST_SHELL
+        self.shell = shell.get()
 
-        self.shell_name = self.shell.split()[0]
-        self.shell_opts = self.shell.split()[1:]
+        # set shebang to value defined in Buildspec, if not defined then get one from Shell class
+        self.shebang = self.recipe.get("shebang") or self.shell["shebang"]
 
-        # if shell program (bash, sh, python) is not present in system raise an error
-        if not shutil.which(self.shell_name):
-            raise BuildTestError(
-                f"Can't find executable {self.shell_name}, please check your $PATH or install the appropriate package"
-            )
-
-        self.shebang = self.recipe.get("shebang")
-
-        # if shebang is not found in recipe let's set it to #!/bin/bash
-        if not self.shebang:
-            # when shell: python we set shebang to location of 'python' wrapper
-            if "python" in self.shell_name:
-                self.shebang = "#!" + shutil.which("python")
-            # otherwise we set to #!/bin/bash
-            else:
-                self.shebang = "#!" + BUILDTEST_SHELL
-
-            self.logger.debug(
-                f"shebang not found in Buildspec, setting shebang to {self.shebang} "
-            )
-
-        self.logger.debug(f"shell: {self.shell}")
-        self.logger.debug(f"shell name: {self.shell_name}")
-        self.logger.debug(f"shell opts: {self.shell_opts}")
-        self.logger.debug(f"shebang: {self.shebang}")
+        self.logger.debug(f"Shell Details: {self.shell}")
+        self.logger.debug(f"Shebang: {self.shebang}")
 
     def __str__(self):
         return "[builder-%s-%s]" % (self.type, self.name)
@@ -341,7 +318,7 @@ class BuilderBase:
            :rtype: str
         """
 
-        if "python" in self.shell_name:
+        if "python" in self.shell["name"]:
             self.logger.debug("Setting test extension to 'py'")
             return "py"
 
@@ -359,10 +336,7 @@ class BuilderBase:
 
         env = []
         pairs = self.recipe.get("env")
-        shell = self.shell_name
-        # If we are using a python shell, we need to import os first
-        if "python" in shell:
-            env.append("import os")
+        shell = self.shell["name"]
 
         # Parse environment depending on expected shell
         if pairs:
@@ -371,21 +345,12 @@ class BuilderBase:
             if re.search("(bash|sh)$", shell):
                 [env.append("%s=%s" % (k, v)) for k, v in pairs.items()]
 
-            # python interpreter can be a shell
-            elif "python" in shell:
-                [env.append("os.putenv('%s','%s')" % (k, v)) for k, v in pairs.items()]
-
             else:
                 self.logger.warning(
                     f"{shell} is not supported, skipping environment variables."
                 )
 
         return env
-
-    def get_shell(self):
-        """Return the shell defined in the recipe, or default to bash."""
-
-        return self.recipe.get("shell", BUILDTEST_SHELL)
 
     def run_wrapper(func):
         """The run wrapper will execute any prepare_run and finish_run 
@@ -432,8 +397,7 @@ class BuilderBase:
                 self.metadata[known_section] = self.recipe.get(known_section)
 
         # Get the shell (sh, bash) for writing testscript. A Buildspec could specify this via ``shell`` key
-        self.metadata["shell"] = self.shell_name
-        self.metadata["shell_opts"] = self.shell_opts
+        self.metadata["shell"] = self.shell
 
         # Derive the path to the test script
         self.metadata["testpath"] = "%s.%s" % (
@@ -538,18 +502,14 @@ class BuilderBase:
         os.chdir(self._get_testdir())
         self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
-        # build the command to run the test-script
+        # command to run the test
         cmd = []
-        # add the shell (/bin/bash, /bin/sh, sh, bash, python)
-        cmd += [self.shell_name]
-        # add shell opts if set
-        if self.shell_opts:
-            cmd += self.shell_opts
-        cmd += [testfile]
-
-        self.logger.debug(f"Running Test via command: {' '.join(cmd)}")
-
-        command = BuildTestCommand(cmd)
+        # build the run command that includes the shell path, shell options and path to test file
+        cmd += [self.shell["path"], self.shell["opts"], testfile]
+        self.metadata["command"] = " ".join(cmd)
+        self.logger.debug(f"Running Test via command: {self.metadata['command']}")
+        
+        command = BuildTestCommand(self.metadata["command"])
         out, err = command.execute()
 
         # Record the ending time
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -269,18 +269,18 @@ class BuilderBase:
                 % (self.type, self.recipe.get("type"))
             )
         # The default shell will be bash
-        shell = Shell()
-        # if 'shell' key  is defined in Buildspec let's override the default shell
-        if self.recipe.get("shell"):
-            shell = Shell(self.recipe.get("shell"))
-
-        self.shell = shell.get()
+        self.shell = Shell(self.recipe.get("shell", "bash"))
 
         # set shebang to value defined in Buildspec, if not defined then get one from Shell class
-        self.shebang = self.recipe.get("shebang") or self.shell["shebang"]
+        self.shebang = self.recipe.get("shebang") or self.shell.shebang
 
-        self.logger.debug(f"Shell Details: {self.shell}")
-        self.logger.debug(f"Shebang: {self.shebang}")
+        self.logger.debug(f"Shell Details: ")
+        self.logger.debug(f"Shell Name: {self.shell.name}")
+        self.logger.debug(f"Shell Opts: {self.shell.opts}")
+        self.logger.debug(f"Shell Path: {self.shell.path}")
+        self.logger.debug(f"Shell Shebang: {self.shell.shebang}")
+
+        self.logger.debug(f"Shebang used for test: {self.shebang}")
 
     def __str__(self):
         return "[builder-%s-%s]" % (self.type, self.name)
@@ -318,7 +318,7 @@ class BuilderBase:
            :rtype: str
         """
 
-        if "python" in self.shell["name"]:
+        if "python" in self.shell.name:
             self.logger.debug("Setting test extension to 'py'")
             return "py"
 
@@ -335,20 +335,20 @@ class BuilderBase:
         """
 
         env = []
-        pairs = self.recipe.get("env")
-        shell = self.shell["name"]
+        pairs = self.recipe.get("env", [])
+        shell = self.shell.name
 
         # Parse environment depending on expected shell
-        if pairs:
+        # if pairs:
 
-            # Handles bash and sh
-            if re.search("(bash|sh)$", shell):
-                [env.append("%s=%s" % (k, v)) for k, v in pairs.items()]
+        # Handles bash and sh
+        if re.search("(bash|sh)$", shell):
+            [env.append("%s=%s" % (k, v)) for k, v in pairs.items()]
 
-            else:
-                self.logger.warning(
-                    f"{shell} is not supported, skipping environment variables."
-                )
+        else:
+            self.logger.warning(
+                f"{shell} is not supported, skipping environment variables."
+            )
 
         return env
 
@@ -395,9 +395,6 @@ class BuilderBase:
         for known_section in known_sections:
             if known_section in self.recipe:
                 self.metadata[known_section] = self.recipe.get(known_section)
-
-        # Get the shell (sh, bash) for writing testscript. A Buildspec could specify this via ``shell`` key
-        self.metadata["shell"] = self.shell
 
         # Derive the path to the test script
         self.metadata["testpath"] = "%s.%s" % (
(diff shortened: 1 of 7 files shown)


shahzebsiddiqui:
@vsoch i have tested this locally for all the shell types and options

I will just give an example for python when ``examples/script/circle-area.py``. Shown below is the generated script, note that the python interpreter is set to path of 'python' and environment is also working.

```
(buildtest) ssi29@ag-mxg-hulk090> cat .buildtest/python-circle/circle_area.py
#!/u/users/ssi29/.local/share/virtualenvs/buildtest-0NCSLNEz/bin/python
import os

os.putenv("foo", "bar")
import math

radius = 2
area = math.pi * radius * radius
print("Circle Radius ", radius)
print("Area of circle ", area)

```

For the other bash/sh examples i updated the tutorials repo https://github.com/buildtesters/tutorials/commit/19ef0a228521fe5221bb0167ed63a0da9d8ef07b where some of these tests can be run.

buildtest/buildsystem/base.py
```diff
@@ -266,6 +267,36 @@ def __init__(self, name, recipe, buildspec=None, testdir=None):
                 % (self.type, self.recipe.get("type"))
             )
 
+        self.shell = self.recipe.get("shell")
+        # if 'shell' key not found in Buildspec, let's set it to BUILDTEST_SHELL and log a message
+        if not self.shell:
```
vsoch: Do you want to make self.shell a property instead? I can see this being reused.

buildtest/buildsystem/base.py
```diff
@@ -266,6 +267,36 @@ def __init__(self, name, recipe, buildspec=None, testdir=None):
...
+
+        self.shell_name = self.shell.split()[0]
+        self.shell_opts = self.shell.split()[1:]
+
+        self.shebang = self.recipe.get("shebang")
+        # if shebang is not found in recipe let's set it to #!/bin/bash
+        if not self.shebang:
+            # when shell: python we set shebang to location of 'python' wrapper
+            if "python" in self.shell_name:
+                self.shebang = "#!" + shutil.which("python")
```
vsoch: This will trigger an error if Python isn’t found and returned as None.

shahzebsiddiqui:
I have attached an example buildtest.log for this run. The system directory contains a few examples of the shell and shebang

```
(buildtest) ssi29@ag-mxg-hulk090> buildtest build -b system/

            Discovered Buildspecs

/u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/systemd.yml
/u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/selinux.yml
/u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/ulimits.yml
/u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/disk_usage.yml



systemd                        systemd_default_target         PASS                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/systemd.yml
selinux                        selinux_disable                PASS                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/selinux.yml
ulimits                        ulimit_filelock                PASS                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/ulimits.yml
ulimits                        ulimit_cputime                 PASS                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/ulimits.yml
ulimits                        ulimit_stacksize               FAIL                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/ulimits.yml
disk_usage                     root_disk_usage                PASS                           /u/users/ssi29/.buildtest/site/github.com/buildtesters/tutorials/system/disk_usage.yml


                        Test summary
Executed 6 tests
Passed Tests: 5/6 Percentage: 83.333%
Failed Tests: 1/6 Percentage: 16.667%

```
[buildtest.log](https://github.com/buildtesters/buildtest/files/4547521/buildtest.log)

buildtest/buildsystem/base.py
vsoch: Wouldn’t there be a more elegant way to do this instead of hard coding python?

buildtest/buildsystem/base.py
```diff
@@ -319,8 +351,7 @@ def get_environment(self):
 
         env = []
         pairs = self.recipe.get("env")
-        shell = self.get_shell()
-
+        shell = self.shell_name
```
vsoch: This would use the self.shell property.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: Yea if python is not found we can raise an error. Also ``shell: python`` adapts to both python2 or python3 but it's not clear to user when they write in ``run`` section and test is not portable since it depends on user python environment. So if it works for me it may not necessarily work for you.

Right now we don't have ability to pass system python let's say ``shell: /usr/bin/python`` or ``shell: /usr/bin/python3`` similarly.

I particularly don't like having ``env`` support when shell is python, it causes some confusion to user when it sees buildtest modify the python script. I was thinking of removing this support but didnt for time being

buildtest/buildsystem/base.py
shahzebsiddiqui: Can you describe with example how you plan to reuse the shell decorator. I have been getting up to speed on decorators but haven't found a use-case. 

I am thinking you want to invoke BuilderBase class directly if i understand. I think changing shell from ``bash`` --> ``sh`` or ``sh`` --> ``bash`` would be okay for most part, but there are some minor differences between the two shells that can cause an error depending on how ``run`` section is setup. I was able to find this link https://www.gnu.org/software/bash/manual/html_node/Major-Differences-From-The-Bourne-Shell.html outlining some of the differences between ``bash`` and ``sh`` 

I don't see a usecase where we can allow a shell to change from ``bash`` or ``sh`` --> ``python`` since that would cause run section to break pretty dramatically.

buildtest/buildsystem/base.py
shahzebsiddiqui: I don't think it's possible for ``python`` to not be found when running buildtest. It would cause an error running buildtest. In the ``master`` branch we had the **buildtest** https://github.com/buildtesters/buildtest/blob/master/bin/buildtest as a shell script that checks for the python client as a requirement. That has been removed in ``devel`` but i can see it being useful having these checks at start of buildtest.

buildtest/buildsystem/base.py
vsoch: All the times in the code we do some action like "self.recipe.get("shell") we would want to just reference self.shell. And I was thinking this morning, since there are now a bunch of different parameters around a shell, would it make sense to have it be a class? E.g.,

```Python
self.shell = Shell(self.recipe.get('shell', '/bin/bash'))
self.shell.opts
self.shell.shebang
```
etc.

buildtest/buildsystem/base.py
vsoch: We only have python3 on sherlock, which is what made me think of this. I think (given that there can be multiple shells allowed) instead of hard coding them in specifically, it should be done with a single list, the schema validates, but then the code extracts (whatever the valid thing is). That's a much more elegant implementation than hard coding python into the code (and all the other future cases).

buildtest/buildsystem/base.py
shahzebsiddiqui: I have been thinking about the ``shell`` key this is specific to ``LocalExecutor`` behavior since we are running test locally. This wont be relevant when running test with ``SlurmExector``. It's true we still need to detect the shell for writing the test, get test extension,  but execution of test is depended on the executor type. We also have ``executor`` key https://github.com/buildtesters/buildtest/blob/devel/buildtest/buildsystem/schemas/script/script-v0.0.1.schema.json#L39  which we have not implemented during the build stage. Right now everything defaults to LocalExecutor but again this is out of scope for this PR but worth pointing out.

buildtest/buildsystem/base.py
shahzebsiddiqui: Are you referring to modifying the schema line https://github.com/buildtesters/schemas/blob/master/script/script-v0.0.1.schema.json#L44 

Do you want add python3 and system python (/bin/python, /bin/python3) into the pattern

```
"pattern": "^(/bin/bash|/bin/sh|sh|bash|python|python3|/bin/python|/bin/python3).*"
```

The python interpreter will be whatever python wrapper was used when user is invoking buildtest so it's not possible to not have python in $PATH. I think having the distinction between python3 and python2 would be useful so user knows how to write their script in ``run`` section.

vsoch: I think I've mentioned this before - in that shell is an executable, I don't think we should limit the user to some known set. I should be able to run bash, sh (or a specific location, /bin/sh) or python3, or python4 when it comes out. We shouldn't need to pre-define the executables that the use can specify, only look for them with shutil.which() and exit with a clear error if they chose something not findable.

shahzebsiddiqui: If i understand you like to run ``shutil.which()`` on ``self.shell_name`` and if its ``None`` raise an error.  This would be much earlier in the code, but note that we have ``BuildTestCommand`` raise this error at https://github.com/buildtesters/buildtest/blob/devel/buildtest/utils/command.py#L46 which would eventually catch this exception.

vsoch: Yep.

shahzebsiddiqui: See https://github.com/buildtesters/buildtest/pull/290/commits/922756693cf0320010121c16e3deb243b2c756ba

buildtest/buildsystem/base.py
shahzebsiddiqui: I dont feel comfortable making this change, i think this relates to discussion  https://github.com/buildtesters/buildtest/pull/290#discussion_r417504156

vsoch: Yep happy to review this again in another PR.

vsoch:
Since this PR is directly relevant to factoring in changes to shell, I think the previous discussion about a class for the Shell (and a much simpler interface here) is scoped within this PR, and I'd like to see those changes here to clean up the currently long logic to just load the shell and get it's opts, etc.

shahzebsiddiqui:
Okay i will try to implement ``Shell`` class tomorrow.
A few questions 
1. Any preference where the ``Shell`` class should be declared with respect to modules. Should this a new file let's say ``buildtest/utils/shell.py``, or part of ``buildtest/buildsystem/base.py`` or somewhere else?
2. I don't think ``self.shell.shebang`` should be available in Shell class based on https://github.com/buildtesters/buildtest/pull/290#discussion_r416881599. The shell script with a shebang ``#!/bin/bash`` will run with the ``/bin/bash`` as the shell interpreter. The script can also be run directly but when we need to pass shell options then we need to do something like ``bash [option] script.sh``. I think we need to make a distinction between ``shell`` and ``shebang`` key. The shebang key is used to tweak the test script shebang line whereas the shell key  is used to tweak how to run the test. The more i think about shell key this looks more suitable for LocalExecutor behavior.

vsoch:
> Any preference where the Shell class should be declared with respect to modules. Should this a new file let's say buildtest/utils/shell.py, or part of buildtest/buildsystem/base.py or somewhere else?

Yes! I would even go as far to say that (this is cool enough) that someone might want to use the class as a helper library even outside of build test (meaning it could exist as it's own Python module) but since this is very specific to buildtest, unlike schemas, let's keep it inside the buildtest module. I think we should put it where it intuitively would be found, so akin to a BuildTestCommand, I would do Shell under:

```
buildtest/utils/shell.py
```

so we would do:

```python
from buildtest.utils.shell import Shell
```

I don't think there is anything special about the shell, but if you think it's more consitent you could also do 

```python
from buildtest.utils.shell import BuildTestShell
```
Even though I'm not sure there is a thing called a "build test shell." I'll leave that up to you.

> I don't think self.shell.shebang should be available in Shell class based on #290 (comment). The shell script with a shebang #!/bin/bash will run with the /bin/bash as the shell interpreter. The script can also be run directly but when we need to pass shell options then we need to do something like bash [option] script.sh. I think we need to make a distinction between shell and shebang key. The shebang key is used to tweak the test script shebang line whereas the shell key is used to tweak how to run the test. The more i think about shell key this looks more suitable for LocalExecutor behavior.

Sure, that works. If you want to have a function that will generate a shebang based on a shell, that could be provided, and perhaps used if the user hasn't set anything. Again, totally up to you, and I'm happy to review what you decide.

buildtest/utils/shell.py
```diff
@@ -0,0 +1,67 @@
+import shutil
+from buildtest.exceptions import BuildTestError
+
+
+class Shell:
```
vsoch, line 5: Perhaps a default to bash?

buildtest/utils/shell.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # enforce input argument 'shell' to be a string
+        if not isinstance(shell, str):
+            raise BuildTestError(
+                f"Invalid type for input: {shell} must be of type 'str'"
+            )
+
+        self.name = shell.split()[0]
+        self.opts = " ".join(shell.split()[1:])
+
```
vsoch, line 25: The function opts would immediately override the class variable of the same name.

buildtest/utils/shell.py
```diff
@@ -0,0 +1,67 @@
...
+           >>> shell.name="sh"
+           >>> shell.path
+           '/usr/bin/sh'
+
+        """
+
+        path = shutil.which(self.name)
+
+        # raise an exception if shell program is not found
+        if not path:
```
vsoch, line 74: These should be done in init.

shahzebsiddiqui:
Hey @vsoch,

So i have created the Shell class which can be invoked directly so i will share how this works before i integrate into Builderbase

This is a simple call to invoke a bash shell and use the get method to retrieve all the shell attributes
```
>>> shell = Shell("bash")
>>> shell.get()
{'name': 'bash', 'opts': '', 'path': '/usr/bin/bash', 'shebang': '#!/usr/bin/bash'}
```

We can retrieve the shell path by invoking 

```
>>> shell.path
'/usr/bin/bash'
```

The class automatically updates shell path if the ``name`` attribute is changed and if i understood correctly you wanted to use decorators for this purpose

Note i change ``shell.name="sh"`` and ``shell.path ``and ``shell.shebang`` also changed
```
>>> shell.name="sh"
>>> shell.get()
{'name': 'sh', 'opts': '', 'path': '/usr/bin/sh', 'shebang': '#!/usr/bin/sh'}

```

We can also update the opts which would be useful during a shell change 

```
>>> shell.opts = "-c"
>>> shell.get()
{'name': 'sh', 'opts': '-c', 'path': '/usr/bin/sh', 'shebang': '#!/usr/bin/sh'}
```

If user specify invalid shell we raise an exception such as this example

```
>>> shell.name = "bash1"
>>> shell.get()
  File "<stdin>", line 1, in <module>
  File "/mxg-hpc/users/ssi29/buildtest/buildtest/utils/shell.py", line 65, in get
    self.shell["path"] = self.path
  File "/mxg-hpc/users/ssi29/buildtest/buildtest/utils/shell.py", line 53, in path
    raise BuildTestError(f"Can't find program: {self.name}")
buildtest.exceptions.BuildTestError: "Can't find program: bash1"
```

I assumed the input in Shell constructor must be a string so i raise an exception if it's not enforced.

buildtest/utils/shell.py
```diff
@@ -0,0 +1,67 @@
...
+            raise BuildTestError(f"Can't find program: {self.name}")
+
+        # shebang is formed by adding the char '#!' with path to program
+        self.shebang = f"#!{path}"
+        return path
+
+    def get(self):
```
vsoch, line 84: Test this out - most objects already can be serialized as dict so it’s about setting properties.

vsoch (changes requested):

buildtest/utils/shell.py
shahzebsiddiqui: i can do that so make argument ``shell`` as optional argument so that we can invoke the class as
`` shell = Shell()`` which defaults to bash.

buildtest/utils/shell.py
shahzebsiddiqui: so what are you suggesting we do? Remove ``opts`` method?

buildtest/utils/shell.py
shahzebsiddiqui: So how is this

```
>>> shell = Shell()
>>> shell.get()
{'name': 'bash', 'opts': '', 'path': '/usr/bin/bash', 'shebang': '#!/usr/bin/bash'}
>>> sh = Shell("sh")
>>> sh.get()
{'name': 'sh', 'opts': '', 'path': '/usr/bin/sh', 'shebang': '#!/usr/bin/sh'}

```

buildtest/utils/shell.py
shahzebsiddiqui: ok i have done this in init also, but found that i also need this check in ``path`` method when there is a change in name attribute meaning ``shell.name = "python1"`` we report an error otherwise the path and shebang will be ``None``

buildtest/utils/shell.py
vsoch: It would be appropriate to parse original opts in the init, but then have this function more properly named based on what it does, e.g., shell.set_opts()

buildtest/utils/shell.py
vsoch: Just create a function to handle these operations and call it in init. If the user of the Shell class wants to call it again after, they are free to do that.

shahzebsiddiqui:
I am having hard time understanding based on your comments on what you want. This would be where you can issue a PR to my branch with how you plan to use the Shell class. I guess once this class is done i will update BuilderBase to use the Shell class.

vsoch:
All set. Take a look!

shahzebsiddiqui:
It's not perfect but i tested it out and didn't intend user to update the shell path which will update the name since that is quite hard to do

For example in this example the name get's same as path.
```
>>> shell.path="/usr/bin/python"
>>> shell.get()
{'name': '/usr/bin/python', 'opts': '', 'path': '/usr/bin/python', 'shebang': '#!/usr/bin/python'}
```

Another thing is if user changes the name, the path and shebang are not updated automatically as shown below

```
>>> shell.name="sh"
>>> shell.get()
{'name': 'sh', 'opts': '', 'path': '/usr/bin/python', 'shebang': '#!/usr/bin/python'}
```

At the moment, i don't really care as buildtest will invoke Shell once and get the appropriate shell. I kinda liked my implementation before, where path and shebang were only changed based on name change.

vsoch:
I can only implement what you’ve clearly communicated as being important. Please make a list of billets for how you want it to work and I’ll update the class.

shahzebsiddiqui:
Features for Shell class
- [x] User can invoke ``Shell()`` which should default to bash 
- [x] User can update the class attribute ``opts`` so something like ``shell.opts="-v"`` this may be useful to override options. 
- [x] Catch exception when invalid path detected during init i.e ``shell = Shell("python4")``
- [] Catch exception when invalid path is result of change in attribute ``name`` (i.e self.name="python4")
```
>>> shell.name="python4"
>>> shell.get()
{'name': 'python4', 'opts': '', 'path': '/usr/bin/bash', 'shebang': '#!/usr/bin/bash'}
```
- [] User can update class attribute ``name`` let's assume shell is an instance of Shell we can do ``shell.name="python"`` which will update the path and shebang automatically (not working now)

```
>>> shell = Shell()
>>> shell.get()
{'name': 'bash', 'opts': '', 'path': '/usr/bin/bash', 'shebang': '#!/usr/bin/bash'}
>>> shell.name="sh"
>>> shell.get()
{'name': 'sh', 'opts': '', 'path': '/usr/bin/bash', 'shebang': '#!/usr/bin/bash'}

```

- [] Remove the feature where the class attribute path can be changed once object is called so remove ``@path.setter`` https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_shell_logic/buildtest/utils/shell.py#L66
This results in weird things where user can do something really bizarre

```
>>> shell.path="hostname"
>>> shell.get()
{'name': 'hostname', 'opts': '', 'path': '/usr/bin/hostname', 'shebang': '#!/usr/bin/hostname'}
```

vsoch:
@shahzebsiddiqui you can't have path update name and vice versa - that leads to an infinite loop :)

vsoch:
The way that shell is designed now works as expected to instantiate the Shell, and I don't see a use case being that a user would need to go in and start customizing random things. So I'd suggest that you keep as is and open these issues when they are actually proven to be issues.

shahzebsiddiqui:
> The way that shell is designed now works as expected to instantiate the Shell, and I don't see a use case being that a user would need to go in and start customizing random things. So I'd suggest that you keep as is and open these issues when they are actually proven to be issues.

Agreed, right now i don't have a use-case of using Shell class directly I think this was suggestion you made. Anyhow, i integrate this into BuilderBase and is ready for final review.

buildtest/buildsystem/base.py
```diff
@@ -265,6 +268,19 @@ def __init__(self, name, recipe, buildspec=None, testdir=None):
                 "Mismatch in type. Builder expects %s but found %s."
                 % (self.type, self.recipe.get("type"))
             )
+        # The default shell will be bash
+        shell = Shell()
+        # if 'shell' key  is defined in Buildspec let's override the default shell
+        if self.recipe.get("shell"):
+            shell = Shell(self.recipe.get("shell"))
+
```
vsoch: These 5 lines can just be

```python
shell = Shell(self.recipe.get("shell", "bash"))
```

buildtest/buildsystem/base.py
```diff
@@ -265,6 +268,19 @@ def __init__(self, name, recipe, buildspec=None, testdir=None):
...
+        self.shell = shell.get()
+
+        # set shebang to value defined in Buildspec, if not defined then get one from Shell class
+        self.shebang = self.recipe.get("shebang") or self.shell["shebang"]
```
vsoch: The reason to have the self.shell as the shell class is that we would do:

```
self.shell = Shell()
```
and then refer to `self.shell.shebang` and not need to do shell.get() (this doesn't make sense to me). If the shebang can be included with the shell init and then referenced in the same way, this would also be cleaner code.

buildtest/buildsystem/base.py
```diff
@@ -319,11 +336,7 @@ def get_environment(self):
 
         env = []
         pairs = self.recipe.get("env")
-        shell = self.get_shell()
-
-        # If we are using a python shell, we need to import os first
-        if "python" in shell:
-            env.append("import os")
+        shell = self.shell["name"]
```
vsoch: This would be `self.shell.name`

buildtest/buildsystem/base.py
```diff
@@ -319,11 +336,7 @@ def get_environment(self):
 
         env = []
         pairs = self.recipe.get("env")
```
vsoch: If you do self.recipe.get("env", []) you don't need to do `if pairs` but can just loop through them still (an empty list).

buildtest/buildsystem/base.py
```diff
@@ -393,7 +397,7 @@ class data structures that are used by both run and dry_run
                 self.metadata[known_section] = self.recipe.get(known_section)
 
         # Get the shell (sh, bash) for writing testscript. A Buildspec could specify this via ``shell`` key
-        self.metadata["shell"] = self.get_shell()
+        self.metadata["shell"] = self.shell
```
vsoch: Why is this added to metadata?

buildtest/buildsystem/base.py
```diff
@@ -498,12 +502,14 @@ def run_tests(self, testfile):
...
         self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
-        # Run the test file using the shell
-        cmd = [self.get_shell(), testfile]
-
-        self.logger.debug(f"Running Test via command: {' '.join(cmd)}")
+        # command to run the test
+        cmd = []
+        # build the run command that includes the shell path, shell options and path to test file
+        cmd += [self.shell["path"], self.shell["opts"], testfile]
```
vsoch: This again should reference properties and not dictionary items, given that you are interacting with the class.

buildtest/utils/command.py
```diff
@@ -54,27 +116,26 @@ def execute(self):
...
-
-            out = self.decode(process.stdout.readline())
-            err = self.decode(process.stderr.readline())
-
-            # Append output and error
-            if out:
-                self.out.append(out)
-            if err:
-                self.err.append(out)
+        # Capturing provides temporary output and error files
```
vsoch: Why are these changes showing up here when I already made them in another PR? I wrote these lines...

vsoch (changes requested):

buildtest/utils/command.py
shahzebsiddiqui: I was testing something, this was an accident my bad ill undo this, ill copy the file from devel since i didnt make any change to the file.

buildtest/buildsystem/base.py
shahzebsiddiqui: Ill remove it for now, it was not clear how this metadata will be used going forward one thought was to use the metadata to write to a json file for reporting (``buildtest build report``) #284

buildtest/buildsystem/base.py
shahzebsiddiqui: This is now fixed along with the other stuff mentioned, feel free to resolve what you think is addressed.

buildtest/buildsystem/base.py
```diff
@@ -265,6 +268,19 @@ def __init__(self, name, recipe, buildspec=None, testdir=None):
                 "Mismatch in type. Builder expects %s but found %s."
                 % (self.type, self.recipe.get("type"))
             )
+        # The default shell will be bash
+        self.shell = Shell(self.recipe.get("shell", "bash"))
```
vsoch, line 272: This whole section is much cleaner than it was before - good job!

buildtest/buildsystem/base.py
```diff
@@ -318,36 +335,23 @@ def get_environment(self):
...
+        pairs = self.recipe.get("env", [])
+        shell = self.shell.name
 
         # Parse environment depending on expected shell
-        if pairs:
-
-            # Handles bash and sh
-            if re.search("(bash|sh)$", shell):
-                [env.append("%s=%s" % (k, v)) for k, v in pairs.items()]
+        # if pairs:
```
vsoch: This is commented out and should not be there.

buildtest/utils/command.py
```diff
@@ -1,7 +1,69 @@
+import os
 import locale
 import subprocess
 import shlex
 import shutil
+import tempfile
```
vsoch: This was written by me, and shouldn't appear like this. It suggests that you didn't pull the merged commits from my contribution. It's not really okay to take credit for my work, so this needs to be fixed.

shahzebsiddiqui:
Travis is failing because ``shell`` is part of ``metadata`` in the test https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_shell_logic/tests/buildsystem/test_base.py#L54 which i dont understand exactly why this was being done in the test. I am guessing we are checking if each env variable is present.

 since i removed ``shell`` it i can put it back in but it would be ``self.metadata["shell"]=self.shell.name`` Originally it was ``self.metadata["shell"]=self.get_shell()`` which was returning the name of the shell

buildtest/buildsystem/base.py
shahzebsiddiqui: See https://github.com/buildtesters/buildtest/pull/290#commitcomment-38882276. I introduced this bug by commenting this out. I guess we still need the condition ``if pairs`` when a Buildspec doesn't define ``env`` section we just skip the whole thing

shahzebsiddiqui:
I have added ``shell`` back into ``self.metadata`` i didn't realize this was used in https://github.com/shahzebsiddiqui/buildtest-framework/blob/update_shell_logic/tests/buildsystem/test_base.py#L54 i don't really like the way this was setup as ``self.metadata["shell"]`` is not used anywhere in the code except the line i mentioned in test_base.py
The test should be passing, if you dont like this i would suggest to remove the check in ``test_base.py``

vsoch:
Please address why you have the commits from my Capturing class, already commit, that are included in this PR as your own. This absolutely cannot be merged in that state, as you would be stealing my work.

buildtest/utils/command.py
shahzebsiddiqui: I am not sure what happened i do see your git commit **Fix/command out err (#289)** in my commit history https://github.com/shahzebsiddiqui/buildtest-framework/commits/update_shell_logic 
The only other explanation could be that when i copy-paste to fix this i applied black so it's showing a diff.
I am bit afraid to do a merge from devel now

shahzebsiddiqui:
> Please address why you have the commits from my Capturing class, already commit, that are included in this PR as your own. This absolutely cannot be merged in that state, as you would be stealing my work.

I am trying to figure this out see https://github.com/buildtesters/buildtest/pull/290#discussion_r418695856

vsoch:
If you have integrated devel and this persists, then you must have done something (copy paste??) to have messed that up. Please start fresh with devel, and then cherry pick the commits from here, not including the one that re-does the work that I already did. I don't know what you did with git, but it's not including my work, but counting it here as new work (authored by you) which is really not ok :(

shahzebsiddiqui:
Let me try to merge from devel and see if it fixes the problem i have done it locally now checking if there were any merge conflicts before i push.

vsoch:
Just don't merge this PR, whatever you do.

vsoch:
Ah so that was the issue? You had copy pasted my file without merging (rebasing with devel?) Just curious... why?

shahzebsiddiqui:
Okay i think the merge https://github.com/shahzebsiddiqui/buildtest-framework/commit/4a1b7168092f4db2928d6156388d84a046cac6c9  fixed the commit diff for ``buildtest/utils/command.py``  so now you don't see all the changes i made due to copy-paste sorry about that. I have no clue how this happened.

buildtest/buildsystem/base.py
shahzebsiddiqui: This is fixed now

buildtest/buildsystem/base.py
```diff
@@ -498,12 +500,14 @@ def run_tests(self, testfile):
         os.chdir(self._get_testdir())
         self.logger.debug(f"Changing to directory {self._get_testdir()}")
 
-        # Run the test file using the shell
-        cmd = [self.get_shell(), testfile]
+        # command to run the test
+        cmd = []
+        # build the run command that includes the shell path, shell options and path to test file
```
vsoch, line 503: If the cmd is created here and then has lines added to it, why do we need to create it as an empty list first? E.g., just do:

```python
cmd = [self.shell.path, self.shell.opts, testfile]
```

shahzebsiddiqui: This is fixed now in https://github.com/shahzebsiddiqui/buildtest-framework/commit/b18e1c53fbb42e081b6e016daa911d63677a9295

shahzebsiddiqui:
@vsoch i have added regression test for Shell class in ``test/utils/test_shell.py``  in https://github.com/shahzebsiddiqui/buildtest-framework/commit/b18e1c53fbb42e081b6e016daa911d63677a9295

Good news coverage is 100% for the initial implementation.

vsoch (approved):

PR #299, opened 2020-05-08
Sync schemas and add support for viewing global schema

shahzebsiddiqui:
This PR is adding the following support
- Sync global and script schema 
- Option ``buildtest show schema --global`` to view the global schema. I figure this would be special schema since it doesn't have version and up a directory as pose to script schema
- Set --name to use choices field that allows argparse to handle to input choice which makes sense to me.
- Lastly we increase coverage to 100% for buildtest/menu/show.py

### DIFF
--- a/buildtest/menu/__init__.py
+++ b/buildtest/menu/__init__.py
@@ -189,11 +189,10 @@ class BuildTestParser:
             help="show schema by name (e.g., script)",
             choices=supported_schemas,
         )
-        # can't use attribute args.global so setting dest=main for argparse
         parser_schema.add_argument(
             "-g",
             "--global",
-            dest="main",
+            dest="_global",
             help="show global schema",
             action="store_true",
         )
--- a/buildtest/menu/__init__.py
+++ b/buildtest/menu/__init__.py
@@ -6,6 +6,7 @@ interact with a global configuration for buildtest.
 import argparse
 
 from buildtest import BUILDTEST_VERSION
+from buildtest.defaults import supported_schemas
 from buildtest.menu.build import func_build_subcmd
 from buildtest.menu.get import func_get_subcmd
 from buildtest.menu.config import (
@@ -183,6 +184,17 @@ class BuildTestParser:
             "-v", "--version", help="choose a specific version of schema to show.",
         )
         parser_schema.add_argument(
-            "-n", "--name", help="show schema by name (e.g., script)",
+            "-n",
+            "--name",
+            help="show schema by name (e.g., script)",
+            choices=supported_schemas,
+        )
+        # can't use attribute args.global so setting dest=main for argparse
+        parser_schema.add_argument(
+            "-g",
+            "--global",
+            dest="main",
+            help="show global schema",
+            action="store_true",
         )
         parser_schema.set_defaults(func=show_schema_layout)
--- a/buildtest/menu/show.py
+++ b/buildtest/menu/show.py
@@ -1,28 +1,38 @@
 import json
-import sys
+import os
 
-from buildtest.defaults import supported_schemas
 from buildtest.buildsystem.schemas.utils import (
     load_schema,
     get_schemas_available,
     get_schema_fullpath,
+    here,
 )
 
 
 def show_schema_layout(args):
-    """Implements method ``buildtest show schema``"""
+    """This method implements command ``buildtest show schema`` which displays json schemas
+       supported by buildtest. The input ``args`` is an instance of argparse class that contains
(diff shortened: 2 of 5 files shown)


shahzebsiddiqui:
I have already taken the changes  for script schema from https://github.com/buildtesters/schemas/pull/35 so they should be in sync

buildtest/menu/__init__.py
```diff
@@ -183,6 +184,17 @@ def show_menu(self):
             "-v", "--version", help="choose a specific version of schema to show.",
         )
         parser_schema.add_argument(
-            "-n", "--name", help="show schema by name (e.g., script)",
+            "-n",
+            "--name",
+            help="show schema by name (e.g., script)",
+            choices=supported_schemas,
+        )
+        # can't use attribute args.global so setting dest=main for argparse
```
vsoch: Maybe just do args._global instead? It would still be intuitive , remove the need for this comment, and not trigger the error.

buildtest/menu/show.py
```diff
@@ -1,28 +1,38 @@
...
+       ``buildtest show schema --name <NAME>``. The result is an output of the json schema on the console.
+
+       Parameters:
+
+       :param args: instance of argparse class
+       :type args: <class 'argparse.Namespace'>
+       :result: output of json schema on console
+    """
+
+    # implements ``buildtest show schema --global``. Since we can't use args.global we set this to main
```
vsoch: Same here, args._global is clear.

buildtest/menu/show.py
```diff
@@ -1,28 +1,38 @@
...
+    if args.main:
+        global_schema = os.path.join(here, "global.schema.json")
+        # check if file exists
+        assert global_schema
+        recipe = load_schema(global_schema)
```
vsoch: This should be refactored to get some config schema depending on the args provided, and then one line at the end to print. If you see a pattern that has redundant code and an early return it’s usually an indication that you can implement it more succinctly.

vsoch (changes requested):

buildtest/menu/show.py
shahzebsiddiqui: I realized that I couldn't use ``get_schemas_available`` https://github.com/shahzebsiddiqui/buildtest-1/blob/schema_updates/buildtest/buildsystem/schemas/utils.py#L104 since it's relied on a specific directory structure like ``script/script-v0.0.1.schema.json``.

My thought was to introduce a separate option ``--global`` instead of using ``--name global``. I could separate out the functionality into separate methods one for name schema (``compiler``, ``script``) and the other for global schema.

Unrelated to this but i see that settings schema is in a different path https://github.com/buildtesters/buildtest/blob/devel/buildtest/settings/settings.schema.json  apart from other schemas and it was intentional.  This means the implementation depends on the directory structure and where all the schemas need to be placed.

I had some other ideas in mind but didn't introduce them yet for example
- Display settings schema https://github.com/buildtesters/buildtest/blob/devel/buildtest/settings/settings.schema.json maybe it could be ``buildtest show schema --settings``. I wasn't sure if it made sense to group this feature in ``buildtest config`` with some option like ``buildtest config schema``
- Add support to fetch development schemas such as https://buildtesters.github.io/schemas/script/script-v0.0.1.schema.json and others.

vsoch: You are misunderstanding. I'm not talking about how it's loaded, but your instantiation of the client here.

vsoch: You are doing:
```
if this evaluates to true:
    set file to be this
    load
    print
    return

set file to be this
load
print
```

and what you should do for less redundant code is

```
determine file based on input args, set file to be this
load
print
```
Hopefully that pseudo code helps :)

shahzebsiddiqui: Yea i see your point thanks for the hint, i will make some adjustment and get back to  you.

vsoch: Great! Design details like this are important, you will over time become a better programmer. :)

buildtest/menu/show.py
shahzebsiddiqui: just to be clear you want to change ``dest=name`` to ``dest=_global`` to resemble the name of the option. I couldn't understand is ``args.global`` a reserved attribute in argparse

shahzebsiddiqui:
@vsoch i have refactored the code as you suggested see https://github.com/shahzebsiddiqui/buildtest-1/commit/b6934e7d488372264ed9628b77819f8b0ee2c13a

vsoch (approved):

PR #300, opened 2020-05-12
first attempt at implementing compiler schema in Buildtest

### LINKED ISSUE #296: compiler support
It would be nice to have buidtest support compiler types which picks the appropriate compiler wrapper and just alters variable ``cc``, ``fc``, ``cxx``, ``cflags``, ``cxxflags``, ``ldflags``. For initial support we can try to support ``gnu``, ``intel``, ``pgi``, ``cuda`` and ``clang``.  

The way compilers could work is take a single test and allow user to compile with different compilers so that test doesn't have to be duplicated.

Here is a typical example for compiling vector addition with different compilers

```
version: 0.0.1
hello_gnu:
  type: script
  description: vector addition using gnu compiler
  env: 
    cc: gcc
    cflags: -O2
    ldflags: -lm
    srcdir: apps
    exe: vecAdd
  run: |
    module load gcc
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
 
hello_intel:
  type: script
  description: vector addition using intel compiler
  env: 
    cc: icc
    cflags: -O2
    ldflags: -lm
    srcdir: apps
    exe: vecAdd
  run: |
    module load intel
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
```

Given that above code is repetitive, we should allow some form of compiler selection that sets these variable such that they can be referenced in ``run`` section. I think buildtest should use the ``compiler`` key to just set the variables similar to what is done in ``env`` section that declares environment variable. 

Here is a proposal schema that supports gnu and intel compiler. The required key is ``cc``, ``fc`` or ``cxx`` so user can specify which one they want based on source file they compiling. We have a ``module`` key to allow user to specify a module to load based on the compiler. Down side with this approach is we can't specify multiple gnu compilers to test.

```
version: 0.0.1
hello_compiler_example:
  type: script
  description: vector addition using gnu and intel compiler
  compiler:
    gnu:
      module: |
          module purge
          module load gcc/7.0
      cc: gcc
      cflags: -O2
      ldflags: -lm
    intel:
      module: |
        module purge
        module load intel/18     
      cc: icc
      cflags: -O2
      ldflags: -lm
  env: 
    srcdir: apps
    exe: vecAdd
  run: |
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
```

Here is an approach for a more nested compiler structure. Inside ``gnu`` and ``intel`` we have a list of objects where each object is a pattern type such as ``gcc/7.0`` and ``gcc/6.0``. These names would be tacked on to the test or part of the output report.

With this approach we can support multiple compiler instances but top level key like ``gnu``, ``intel`` would have to be pre-defined.

```
version: 0.0.1
hello_compiler_example:
  type: script
  description: vector addition using gnu and intel compiler
  compiler:
    gnu:
    - gcc/7.0: 
        module: |
            module purge
            module load gcc/7.0
        cc: gcc
        cflags: -O2
        ldflags: -lm

   - gcc/6.0:
       module: |
            module purge
            module load gcc/6.0
       cc: gcc
       cflags: -O1 -Wall
       ldflags: -lm

    intel:
      intel/18: 
        module: |
          module purge
          module load intel/18     
        cc: icc
        cflags: -O2
        ldflags: -lm
  env: 
    srcdir: apps
    exe: vecAdd
  run: |
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
```

Here is an extension to previous design but with ability to avoid redefining variables if they are same across compiler group (``gnu``) or all compilers via use of ``commons`` key. 

```
version: 0.0.1
hello_compiler_example:
  type: script
  description: vector addition using gnu and intel compiler
  compiler:
    gnu:
      commons:
        cc: gcc
        cflags: -O2
        ldflags: -lm
      spec:
      - gcc/7.0: 
           module: |
             module purge
             module load gcc/7.0
     - gcc/6.0:
          module: |
            module purge
            module load gcc/6.0       
          cflags: -O1 -Wall

    intel:
      spec:
      - intel/18: 
          module: |
            module purge
            module load intel/18     
          cc: icc
          cflags: -O2
          ldflags: -lm
  env: 
    srcdir: apps
    exe: vecAdd
  run: |
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
```

The ``commons`` is a shared across group of gnu or intel it's optional, but user can override commons in the case of ``gcc/6.0`` where ``cflags`` was set to ``-O1 -Wall``. 

In this case we had to introduce a new key in this case ``spec`` which is under each compiler group which is a  list of objects.

I realize this can get more complex with what it already is, but this is an idea what i had in mind. The ``module`` key is used to specify a list of modules to load as is. This would get injected into the ``run`` section but at the moment user has no control in what order this is placed. Generally modules are placed right after ``#SBATCH`` directives but user is not aware of this.

Last one is an example but with shared common across all compilers

```
version: 0.0.1
hello_compiler_example:
  type: script
  description: vector addition using gnu and intel compiler
  compiler:
    commons:
      cflags: -O2
      ldflags: -lm
    gnu:
      commons:
        cc: gcc
      spec:
      - gcc/7.0: 
           module: |
             module purge
             module load gcc/7.0    

    intel:
      spec:
      - intel/18: 
          module: |
            module purge
            module load intel/18     
          cc: icc
  env: 
    srcdir: apps
    exe: vecAdd
  run: |
    $cc $cflags -o $exe $srcdir/vecAdd.c $ldflags
    ./$exe 
```

shahzebsiddiqui:
This PR does the following
- Added class CompilerBuilder which is a subclass of BuilderBase that implements the compiler schema.
 - Add support for viewing compiler schema via 'buildtest show schema'
 - Modify global schema since we ran into validation error when specifying maintainer key. This was due to expected type.
 - Add compiler schema to buildtest with some additional key's including ``cxxflags``, ``fflags``, ``cppflags`` ``exec_args``

### DIFF
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -180,10 +180,25 @@ class BuildspecParser:
                         ScriptBuilder(name, recipe, self.buildspec, testdir=testdir)
                     )
                 elif recipe["type"] == "compiler":
-                    builders.append(
-                        CompilerBuilder(name, recipe, self.buildspec, testdir=testdir)
-                    )
-                    launch_compiler(recipe)
+                    if recipe["compiler"].get("name") == "gnu":
+                        builders.append(
+                            GNUCompiler(name, recipe, self.buildspec, testdir=testdir)
+                        )
+                    elif recipe["compiler"].get("name") == "intel":
+                        builders.append(
+                            IntelCompiler(name, recipe, self.buildspec, testdir=testdir)
+                        )
+                    elif recipe["compiler"].get("name") == "pgi":
+                        builders.append(
+                            PGICompiler(name, recipe, self.buildspec, testdir=testdir)
+                        )
+                    elif recipe["compiler"].get("name") == "cray":
+                        builders.append(
+                            GNUCompiler(name, recipe, self.buildspec, testdir=testdir)
+                        )
+                    else:
+                        continue
+
                 else:
                     print(
                         "%s is not recognized by buildtest, skipping." % recipe["type"]
@@ -724,6 +739,15 @@ class CompilerBuilder(BuilderBase):
     def get_ldflags(self):
         return self.ldflags
 
+    def get_path(self):
+        """This method returns the full path for GNU Compilers: ``gcc``, ``g++``, ``gfortran``"""
+        path = {
+            self.cc: shutil.which(self.cc),
+            self.cxx: shutil.which(self.cxx),
+            self.fc: shutil.which(self.fc),
+        }
+        return path
+
     def resolve_source(self, source):
         """This method resolves full path to source file, it checks for absolute path first before checking relative
            path that is relative to Buildspec recipe.
@@ -792,39 +816,23 @@ class CompilerBuilder(BuilderBase):
 
         return "%s.exe" % os.path.basename(self.sourcefile)
 
-    def compiler_setup(self):
-        pass
+    def setup(self):
 
-    def setup(
-        self,
-        cc=None,
-        cxx=None,
-        fc=None,
-        cflags=None,
-        cxxflags=None,
-        fflags=None,
-        ldflags=None,
-        cppflags=None,
-    ):
+        self.compiler_recipe = self.recipe.get("compiler")
+        self.sourcefile = self.resolve_source(self.compiler_recipe["source"])
 
-        self.cc = cc
-        self.cxx = cxx
-        self.fc = fc
-        self.cflags = cflags
-        self.fflags = fflags
-        self.cxxflags = cxxflags
-        self.ldflags = ldflags
-        self.cppflags = cppflags
+        # set executable name and assign to self.executable
+        self.executable = self.set_executable_name()
+        self.lang = self.detect_lang(self.sourcefile)
+        self.compile_cmd = self.generate_compile_cmd()
 
-        print(
-            "setup:",
-            self.cc,
-            self.cflags,
-            self.fc,
-            self.fflags,
-            self.cxx,
-            self.cxxflags,
-        )
+        self.run_cmd = self.build_run_cmd(self.compiler_recipe.get("exec_args"))
+
+        self.cflags = self.compiler_recipe.get("cflags")
+        self.fflags = self.compiler_recipe.get("fflags")
+        self.cxxflags = self.compiler_recipe.get("cxxflags")
+        self.ldflags = self.compiler_recipe.get("ldflags")
+        self.cppflags = self.compiler_recipe.get("cppflags")
 
     def _build_testcontent(self):
         """This method will build the test content from a Buildspec that uses compiler schema. We need a 'compiler'
@@ -838,24 +846,8 @@ class CompilerBuilder(BuilderBase):
         """
 
         self.compiler_setup()
-        # self.setup()
-        self.compiler_recipe = self.recipe.get("compiler")
-        self.sourcefile = self.resolve_source(self.compiler_recipe["source"])
-        print(
-            "_build_testcontent:",
-            self.cc,
-            self.cflags,
-            self.fc,
-            self.fflags,
-            self.cxx,
-            self.cxxflags,
-        )
-        # set executable name and assign to self.executable
-        self.executable = self.set_executable_name()
-        self.lang = self.detect_lang(self.sourcefile)
-        self.compile_cmd = self.generate_compile_cmd()
+        self.setup()
 
-        self.run_cmd = self.build_run_cmd(self.compiler_recipe.get("exec_args"))
         # every test starts with shebang line
         lines = [self.shebang]
 
@@ -934,41 +926,23 @@ class GNUCompiler(CompilerBuilder):
     cxx = "g++"
     fc = "gfortran"
 
-    def __init__(self):
-        pass
 
-    def compiler_setup(self, recipe):
-        self.cflags = recipe.get("cflags")
-        self.fflags = recipe.get("fflags")
-        self.cxxflags = recipe.get("cxxflags")
-        self.ldflags = recipe.get("ldflags")
-        self.cppflags = recipe.get("cppflags")
+class IntelCompiler(CompilerBuilder):
 
-        self.setup(
-            cc=self.cc,
-            cxx=self.cxx,
-            fc=self.fc,
-            cflags=self.cflags,
-            fflags=self.fflags,
-            cxxflags=self.cxxflags,
-            ldflags=self.ldflags,
-            cppflags=self.cppflags,
-        )
-
-    def get_path(self):
-        """This method returns the full path for GNU Compilers: ``gcc``, ``g++``, ``gfortran``"""
-        path = {
-            "gcc": shutil.which(self.cc),
-            "g++": shutil.which(self.cxx),
-            "gfortran": shutil.which(self.fc),
-        }
-        return path
+    cc = "icc"
+    cxx = "icpc"
+    fc = "ifort"
 
 
-def launch_compiler(recipe):
+class PGICompiler(CompilerBuilder):
 
-    compiler_recipe = recipe.get("compiler")
-    if compiler_recipe.get("gnu"):
-        gnu_recipe = compiler_recipe["gnu"]
-        gnu = GNUCompiler()
-        gnu.compiler_setup(gnu_recipe)
+    cc = "pgcc"
+    cxx = "pgc++"
+    fc = "pfortran"
+
+
+class CrayCompiler(CompilerBuilder):
+
+    cc = "cc"
+    cxx = "CC"
+    fc = "ftn"
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -9,6 +9,7 @@ import datetime
 import logging
 import os
 import re
+import shutil
 import stat
 import sys
 
@@ -668,13 +669,6 @@ class CompilerBuilder(BuilderBase):
         ".FTN": "Fortran",
         ".ftn": "Fortran",
     }
-    # compiler table based on Language types
-    compiler_table = {
-        "gnu": {"C": "gcc", "C++": "g++", "Fortran": "gfortran"},
-        "intel": {"C": "icc", "C++": "icpc", "Fortran": "ifort"},
-        "pgi": {"C": "pgcc", "C++": "pgc++", "Fortran": "pgfortran"},
-        "cray": {"C": "cc", "C++": "CC", "Fortran": "ftn"},
-    }
 
     cc = None
     cxx = None
@@ -715,12 +709,40 @@ class CompilerBuilder(BuilderBase):
         self.executable = "%s.exe" % os.path.basename(self.source)
 
         self.compiler_name = None
-        if "gnu" in self.compiler_recipe:
-            self.compiler_name = "gnu"
-        elif "intel" in self.compiler_recipe:
-            self.compiler_name = "intel"
+        self.lang = self.detect_lang(self.source)
 
-        self.detect_compiler(self.source)
+        # Get GNU Compiler if 'gnu' defined in compiler properties.
+        if self.compiler_recipe.get("gnu"):
+            self.compiler_name = "gnu"
+            gnu = GNUCompiler()
+            # get_compilers() will return a list that returns C, C++, and Fortran GNU wrapper
+            self.cc, self.cxx, self.fc = gnu.get_compilers()
+
+        # Get Intel Compiler
+        elif self.compiler_recipe.get("intel"):
+            self.compiler_name = "intel"
+            intel = IntelCompiler()
+            self.cc, self.cxx, self.fc = intel.get_compilers()
+
+        # Get PGI Compiler
+        elif self.compiler_recipe.get("pgi"):
+            self.compiler_name = "pgi"
+            pgi = PGICompiler()
+            self.cc, self.cxx, self.fc = pgi.get_compilers()
+
+        # Get Cray Compiler
+        elif self.compiler_recipe.get("cray"):
+            self.compiler_name = "cray"
+            cray = CrayCompiler()
+            self.cc, self.cxx, self.fc = cray.get_compilers()
+
+        # Set ldflags, cppflags, cflags, cxxflags, fflags if defined in Buildspec
+        self.ldflags = self.compiler_recipe[self.compiler_name].get("ldflags")
+        self.cppflags = self.compiler_recipe[self.compiler_name].get("cppflags")
+        self.cflags = self.compiler_recipe[self.compiler_name].get("cflags")
+        self.fflags = self.compiler_recipe[self.compiler_name].get("fflags")
+        self.cxxflags = self.compiler_recipe[self.compiler_name].get("cxxflags")
+
         cmd = self.generate_compile_cmd()
 
         # every test starts with shebang line
@@ -746,52 +768,21 @@ class CompilerBuilder(BuilderBase):
 
         # add run line to test
         lines.append(" ".join(run))
-
         return lines
 
-    def detect_compiler(self, sourcefile):
-        """This method will detect the compiler wrapper based on the sourcefile and compiler name. This method supports
-           compiler detection for 'gnu' and 'intel'. The compiler detection is based on file extension of source file
-           and name of compiler name.
-        """
+    def detect_lang(self, sourcefile):
+        """This method will return the Programming Language based by looking up  file extension of source file."""
 
         ext = os.path.splitext(sourcefile)[1]
 
-        # variable used to store name of Programming Language (C, C++, Fortran)
-        self.lang = None
-
-        # if ext not in lang_ext_table then raise an error. This table consist of all file extensions that map to a Programming Language
+        # if ext not in self.lang_ext_table then raise an error. This table consist of all file extensions that map to a Programming Language
         if ext not in self.lang_ext_table:
             raise BuildTestError(
                 f"Unable to detect Program Language based on extension: {ext} in file {sourcefile}"
             )
         # Set Programming Language based on ext. Programming Language could be (C, C++, Fortran)
-        self.lang = self.lang_ext_table[ext]
-
-        if self.lang == "C":
-            self.cc = self.compiler_table[self.compiler_name].get(self.lang)
-        elif self.lang == "C++":
-            self.cxx = self.compiler_table[self.compiler_name].get(self.lang)
-        elif self.lang == "Fortran":
-            self.fc = self.compiler_table[self.compiler_name].get(self.lang)
-
-        # if we fail to set any C, C++, Fortran compiler we raise an error since we can't proceed
-        if not any([self.cc, self.cxx, self.fc]):
-
-            print("{:>15} | {:<10}".format("Extension", "Programming Language"))
-            print("{:_<60}".format(""))
-            for k, v in self.lang_ext_table.items():
-                print("{:>15} | {:<10}".format(k, v))
-
-            raise BuildTestError(
-                "Fail to set any C, C++, or Fortran compiler. Please check if you have valid file extension or select the appropriate compiler"
-            )
-
-        # Set ldflags, cppflags, cflags, cxxflags, fflags if defined in Buildspec
-        self.ldflags = self.compiler_recipe[self.compiler_name].get("ldflags")
-        self.cppflags = self.compiler_recipe[self.compiler_name].get("cppflags")
-        self.cflags = self.compiler_recipe[self.compiler_name].get("cflags")
-        self.fflags = self.compiler_recipe[self.compiler_name].get("fflags")
+        lang = self.lang_ext_table[ext]
+        return lang
 
     def generate_compile_cmd(self):
         """This method generates the compilation line and returns the output as a list. The compilation line depends
@@ -836,3 +827,91 @@ class CompilerBuilder(BuilderBase):
             ]
         # remove any None from list
         return list(filter(None, cmd))
+
+
+class GNUCompiler(CompilerBuilder):
+
+    cc = "gcc"
+    cxx = "g++"
+    fc = "gfortran"
+
+    def __init__(self):
+        pass
+
+    def get_compilers(self):
+        return [self.cc, self.cxx, self.fc]
+
+    def get_path(self):
+        """This method returns the full path for GNU Compilers: ``gcc``, ``g++``, ``gfortran``"""
+        path = {
+            "gcc": shutil.which(self.cc),
+            "g++": shutil.which(self.cxx),
+            "gfortran": shutil.which(self.fc),
+        }
+        return path
+
+
+class IntelCompiler(CompilerBuilder):
+
+    cc = "icc"
+    cxx = "icpc"
+    fc = "ifort"
+
+    def __init__(self):
+        pass
+
+    def get_compilers(self):
+        return [self.cc, self.cxx, self.fc]
+
+    def get_path(self):
+        """This method returns the full path for Intel Compilers: ``icc``, ``icpc``, ``ifort``"""
+        path = {
+            "icc": shutil.which(self.cc),
+            "icpc": shutil.which(self.cxx),
+            "ifort": shutil.which(self.fc),
+        }
+        return path
+
+
+class PGICompiler(CompilerBuilder):
+
+    cc = "pgcc"
+    cxx = "pgc++"
+    fc = "pgfortran"
+
+    def __init__(self):
+        pass
+
+    def get_compilers(self):
+        return [self.cc, self.cxx, self.fc]
+
+    def get_path(self):
+        """This method returns the full path for PGI Compilers: ``pgcc``, ``pgc++``, ``pgfortran``"""
+        path = {
+            "pgcc": shutil.which(self.cc),
+            "pgc++": shutil.which(self.cxx),
+            "pgfortran": shutil.which(self.fc),
+        }
+        return path
+
+
+class CrayCompiler(CompilerBuilder):
+
+    cc = "cc"
+    cxx = "CC"
+    fc = "ftn"
+
+    def __init__(self):
+        pass
+
+    def get_compilers(self):
+        return [self.cc, self.cxx, self.fc]
+
+    def get_path(self):
+        """This method returns the full path for Cray Compilers: ``cc``, ``CC``, ``ftn``"""
+        path = {
+            "cc": shutil.which(self.cc),
+            "CC": shutil.which(self.cxx),
+            "ftn": shutil.which(self.fc),
+        }
+        return path
--- a/buildtest/buildsystem/base.py
+++ b/buildtest/buildsystem/base.py
@@ -647,31 +647,54 @@ class ScriptBuilder(BuilderBase):
 
 class CompilerBuilder(BuilderBase):
     type = "compiler"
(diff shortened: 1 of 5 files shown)


shahzebsiddiqui:
@vsoch this is in development and i wanted to get some feedback before this get's too complex.

I have added few examples see https://github.com/buildtesters/tutorials/commit/823986e75c53072dbce62c70e68a5e3047c8cd47 that i was able to run successfully


```
(buildtest) ssi29@ag-mxg-hulk090> buildtest build -b compiler_schema_hello.yml

            Discovered Buildspecs

/mxg-hpc/users/ssi29/tutorials/apps/serial/compiler_schema_hello.yml



compiler_schema_hello          hello_f                        PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/compiler_schema_hello.yml
compiler_schema_hello          hello_c                        PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/compiler_schema_hello.yml
compiler_schema_hello          hello_cplusplus                PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/compiler_schema_hello.yml
compiler_schema_hello          passing_args                   PASS                           /mxg-hpc/users/ssi29/tutorials/apps/serial/compiler_schema_hello.yml


                        Test summary
Executed 4 tests
Passed Tests: 4/4 Percentage: 100.000%
Failed Tests: 0/4 Percentage: 0.000%
```

For now i decided not to pursue module permutation with a test since that complicates things quite a bit. 

The content of the generated scripts are as follows. There is one example for each compiler language, one of them includes a module example and another includes passing arguments.

```
(buildtest) ssi29@ag-mxg-hulk090> cat .buildtest/compiler_schema_hello/hello_cplusplus.sh
#!/usr/bin/bash
g++  -Wall -o hello.cpp.exe /mxg-hpc/users/ssi29/tutorials/apps/serial/src/hello.cpp
hello.cpp.exe
```

```
(buildtest) ssi29@ag-mxg-hulk090> cat .buildtest/compiler_schema_hello/hello_f.sh
#!/usr/bin/bash
module purge && module load GCC
gfortran  -Wall -o hello.f90.exe /mxg-hpc/users/ssi29/tutorials/apps/serial/src/hello.f90
hello.f90.exe
```

```
(buildtest) ssi29@ag-mxg-hulk090> cat .buildtest/compiler_schema_hello/hello_c.sh
#!/usr/bin/bash
gcc  -Wall -o hello.c.exe /mxg-hpc/users/ssi29/tutorials/apps/serial/src/hello.c
hello.c.exe
```

```
(buildtest) ssi29@ag-mxg-hulk090> cat .buildtest/compiler_schema_hello/passing_args.sh
#!/usr/bin/bash
gcc  -Wall -o argc.c.exe /mxg-hpc/users/ssi29/tutorials/apps/serial/src/argc.c
argc.c.exe 1 2 3
```

shahzebsiddiqui:
Also don't panic since i made some changes to schema that are not pushed to ``schema`` repo. I plan to do that at same time as this PR is merged. First i wanted to focus on the implementation and correctness of the generated test content.

vsoch:
My main feedback here is that right now I see you doing a lot of "if this compiler" or "if this then.."  and it's very complicated quickly. I think that suggests that instead of a having a global "compiler" base, you might consider having this be an abstract class (for shared attributes, functions, etc.) but then having a Builder base for each of the compilers you see, e.g.,

```
FortranCompilerBuilder
IntelCompilerBuilder
GNUCompilerBuilder
```
That way, you can avoid this redundancy of if, else, then that could be captured with these clean classes.

shahzebsiddiqui:
Thanks for input, i can change the if/else logic for compiler detection to a lookup table. I don't think it's feasible for creating a class per compiler that would be too many classes given that there are many compilers (Intel, PGI, GNU, CUDA, Clang, NAG) and compiler detection comes down to setting ``cc`` ``fc`` and ``cxx`` variables to the appropriate variables. That's how Make works and i can see this easily extend to ``MakeBuilder``, not something i am committing to yet. 

I will get back with some second round changes.

shahzebsiddiqui:
I made my second round of changes by refactoring the if/else conditions so it should be much simpler now.

vsoch:
I would not implement it in this way, so I think you should find someone else to review. I don't think I would approve this structure, it just won't scale.

shahzebsiddiqui:
Right now it's only me and you. So i can't review myself. Can you comment on specifics on what you think is the issue. I understand ``BuilderBase`` class is quite complex itself and some of the methods need to be moved to appropriate location. One that comes to mind are all the **run** related methods should be part of Executor class. I think the scope of BuilderBase should be to process Buildspec and generate the test script. 
Right now i didn't want to do a rearchitect of BuilderBase class that seems out of scope while adding this new support.

vsoch:
In the long run a more modular design, even for many compilers, is the right way to go - you don’t see it now but there will be subtle differences that are easy to work on as modular classes, and messy at best trying to shove into one class.

shahzebsiddiqui:
I am going to make this modular, the way i think it could be is as follows as per hierarchy of classes by Compiler is as follows. I will limit to these compilers for now and add more as needed.
 
CompilerBuilder 
  |_____ GNUCompiler 
  |_____ IntelCompiler
  |_____ PGICompiler 
  |_____ CrayCompiler

vsoch:
Great! I think this approach is the right way to go 👍

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+
+
+class CompilerBuilder(BuilderBase):
+    type = "compiler"
+
+    # Fortran Extensions Links:
+    # https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-developer-guide-and-reference/top/compiler-setup/using-the-command-line/understanding-file-extensions.html
+    # Fortran Extensions: http://fortranwiki.org/fortran/show/File+extensions
+    lang_ext_table = {
+        ".c": "C",
```
vsoch, line 667: These compiler-specific variables should belong with the appropriate class.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        ".f03": "Fortran",
+        ".f": "Fortran",
+        ".F": "Fortran",
+        ".FOR": "Fortran",
+        ".for": "Fortran",
+        ".FTN": "Fortran",
+        ".ftn": "Fortran",
+    }
+
+    cc = None
```
vsoch, line 684: Same with these.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        if not self.source:
+            sys.exit(f"Failed to resolve path specified by key 'source': {sourcefile}")
+
+        # name of generated executable. Add the .exe extension at end of source file name.
+        self.executable = "%s.exe" % os.path.basename(self.source)
+
+        self.compiler_name = None
+        self.lang = self.detect_lang(self.source)
+
+        # Get GNU Compiler if 'gnu' defined in compiler properties.
```
vsoch: These if else elif should be broken into the appropriate CompilerBuilder base - it doesn't make sense to have to repeat them all here. If there are shared operations before this step, create a common function (for the CompilerBuilder base) that is run/shared for all child classes that need it.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+            self.cc, self.cxx, self.fc = pgi.get_compilers()
+
+        # Get Cray Compiler
+        elif self.compiler_recipe.get("cray"):
+            self.compiler_name = "cray"
+            cray = CrayCompiler()
+            self.cc, self.cxx, self.fc = cray.get_compilers()
+
+        # Set ldflags, cppflags, cflags, cxxflags, fflags if defined in Buildspec
+        self.ldflags = self.compiler_recipe[self.compiler_name].get("ldflags")
```
vsoch: This is also compiler specific.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        # add run line to test
+        lines.append(" ".join(run))
+        return lines
+
+    def detect_lang(self, sourcefile):
+        """This method will return the Programming Language based by looking up  file extension of source file."""
+
+        ext = os.path.splitext(sourcefile)[1]
+
+        # if ext not in self.lang_ext_table then raise an error. This table consist of all file extensions that map to a Programming Language
```
vsoch, line 868: You shouldn't need this complicated logic if the compilers are easily separated - the compiler will know a set of languages, and exit if there is mismatch.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        lang = self.lang_ext_table[ext]
+        return lang
+
+    def generate_compile_cmd(self):
+        """This method generates the compilation line and returns the output as a list. The compilation line depends
+           on the the language detected that is stored in variable ``self.lang``.
+        """
+        cmd = []
+
+        # Generate C compilation line
```
vsoch, line 883: Another example here of if else elif that will be addressed by separation into the appropriate classes. I shouldn't need to edit the Abstract Compiler Base if I want to tweak the FortranCompiler, for example. This is what I'd have to do now.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+                "-o",
+                self.executable,
+                self.source,
+                self.ldflags,
+            ]
+        # remove any None from list
+        return list(filter(None, cmd))
+
+
+class GNUCompiler(CompilerBuilder):
```
vsoch, line 922: This is a great start! This is exactly what I had in mind - see if you can simplify the CompilerBuilder base so that it doesn't have all that if else elif logic and you can just put shared functions there, and specifics in each of the classes here.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: The file extension for C, C++, Fortran are compiler agnostic and not belong to any compiler class. You may be getting confused, the compiler class are not C, Fortran, C++ compiler but rather name of the compiler ``GNU``, ``Intel``, ``PGI``, etc.. Each of these compilers come with a C, C++, Fortran wrapper. That's why i kept ``lang_ext_table`` in CompilerBuilder, the main purpose is to use the extensions to detect the programming language (``self.lang``) which is also compiler agnostic.

vsoch: Ah ok. Is this going to be a common practice - to have a huge table of extensions? Can you think of a simpler way to do this, maybe regular expressions that can be shared across functions and classes instead of a hard coded lookup just with the Compiler class?

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+           and 'source' key which specifies the source files to compile. We resolve the source file path which can
+           be an absolute value or relative path with respect to Buildspec. The file extension of sourcefile is used
+           to detect the Programming Language which is used to lookup the compiler wrapper based on Language + Compiler.
+           During compiler detection, we set class variables ``self.cc``, ``self.cxx``. ``self.fc``, ``self.cflags``,
+           ``self.cxxflags``, ``self.fflags``, ``self.cppflags``. ``self.ldflags``. Finally we generate the compile
+           command and add each instruction to ``lines`` which contains content of test. Upon completion, we return
+           a list that contains content of the test.
+        """
+
+        self.compiler_recipe = self.recipe.get("compiler")
```
vsoch: Is the compiler key guaranteed to be defined?

vsoch: Same question for source.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        sourcefile = self.compiler_recipe["source"]
+        # The 'source' key can accept an absolute path or relative path with respect to Buildspec file.
+        # resolve_path returns full path to file or None if not found. We first check absolute path if not found we
+        # check for relative path.
+        self.source = resolve_path(sourcefile) or resolve_path(
+            os.path.join(
+                os.path.dirname(self.buildspec), self.compiler_recipe["source"]
```
vsoch: If you defined `self.compiler_recipe["source"]` to go into `sourcefile` why don't you use `sourcefile` here as well?

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+        # name of generated executable. Add the .exe extension at end of source file name.
+        self.executable = "%s.exe" % os.path.basename(self.source)
+
+        self.compiler_name = None
+        self.lang = self.detect_lang(self.source)
+
+        # Get GNU Compiler if 'gnu' defined in compiler properties.
+        if self.compiler_recipe.get("gnu"):
+            self.compiler_name = "gnu"
+            gnu = GNUCompiler()
```
vsoch: Okay here is where I still think it needs work - right now you are passing _all_ kinds of compiler specs into the Compiler class, and doing this complicated if/then/else thing. The executors should be modular (and use different names as necessary) so we start, off the bat, with the recipe mapped to the right compiler. The Compiler class does not exist to be a jump off point to determine which other class to instantiate, but instead an Abstract class to implement common functions or variables. Take a look at snakemake.executors if you need a good example of modularity with parent/child classes.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +644,274 @@ def _build_testcontent(self):
...
+            self.cc, self.cxx, self.fc = cray.get_compilers()
+
+        # Set ldflags, cppflags, cflags, cxxflags, fflags if defined in Buildspec
+        self.ldflags = self.compiler_recipe[self.compiler_name].get("ldflags")
+        self.cppflags = self.compiler_recipe[self.compiler_name].get("cppflags")
+        self.cflags = self.compiler_recipe[self.compiler_name].get("cflags")
+        self.fflags = self.compiler_recipe[self.compiler_name].get("fflags")
+        self.cxxflags = self.compiler_recipe[self.compiler_name].get("cxxflags")
+
+        cmd = self.generate_compile_cmd()
```
vsoch: The rest of this recipe is hugely long, and should be broken into functions that can be mapped to steps. I would suggest taking these steps:

 - first refactor the logic to start with a compiler class, removing a lot of the code above with if / else because a Compiler class is just abstract to provide common functions.
 - then for each, determine what steps are common and can be shared. Write those into the Compiler parent class, and just call from the child classes. Leave the specific commands that are compiler specific.

The goal is that shared functionality is represented in neatly defined steps held by the parent Compiler Class. The children then call those steps, and it can be read from top to bottom to easily understand what is going on (it's not a big dump of spaghetti code as it is now).  Then of course each child will have some custom steps / functionality that is more prominent in the class.  I'll continue my review after this change.

vsoch (changes requested):

buildtest/buildsystem/base.py
shahzebsiddiqui: Fortran is the culprit when it comes to many extensions, i converted the if/else into a lookup table which simplified the logic, the fact is you can write any C, C++ program without a file extension and it would still work so buildtest is trying to enforce some rules when user specify source file. 

Eventually i wanted to add ``.cu`` to the lookup table for CUDA language, it seems easier to add key, value so i can map extension to a language name. Another option was to have a separate list per language and then search in list. Either way it doesn't make much difference, I prefer a lookup table. 

```
c_ext = ['.c']
cplus_ext = ['.cpp', '.cc', ...]
fortran_ext = [.'f', ...]
```

buildtest/buildsystem/base.py
shahzebsiddiqui: both compiler and source are required see
https://github.com/shahzebsiddiqui/buildtest-1/blob/compiler_schema/buildtest/buildsystem/schemas/compiler/compiler-v0.0.1.schema.json#L6

https://github.com/shahzebsiddiqui/buildtest-1/blob/6fecf4c42212ab16aec9b18ba35d73c41cf5f3ac/buildtest/buildsystem/schemas/compiler/compiler-v0.0.1.schema.json#L33

The ``source`` key is required with oneOf the compiler names at the moment. For now this was what we discussed in #296

buildtest/buildsystem/base.py
vsoch: Makes sense, let’s have this lookup table (since it’s shared between all compilers) stay as is for now, it will be inherited by child classes.

buildtest/buildsystem/base.py
shahzebsiddiqui: So the way C, C++, Fortran compilation works for single program is as follows, this is true for any compiler to the best of my knowledge.

```
$cc $CPPFLAGS $CFLAGS -o $EXEC $SOURCE $LDFLAGS
$cxx $CPPFLAGS $CXXFLAGS -o $EXEC $SOURCE $LDFLAGS
$fc $CPPFLAGS $FFLAGS -o $EXEC $SOURCE $LDFLAGS
```

The $cc is your C compiler with CFLAGS as option to C compiler, the same is true with Fortran (FFLAGS) and C++ (CXXFLAGS). 
That's how Makefiles work if you take look at this example, you just need to change cc and CFLAGS to map to the compiler you want but the pattern for compilation remains the same.
https://www.cs.swarthmore.edu/~newhall/unixhelp/howto_makefiles.html

That's why i implemented this in Base class. The one issue we have right now is CompilerBuilder is base class of BuilderBase in order to inherit the build logic.  The compilation pattern can change if user wants to build source into object files (.o) and link each object into executable. 

Here is some example https://www3.ntu.edu.sg/home/ehchua/programming/cpp/gcc_make.html for compilation

vsoch: Thank you for your explanation of how compilers work - I'm aware :) This is different from the refactor needed to clean up the base class.

shahzebsiddiqui:
@vsoch i have tried to address some of your concerns you mentioned. Based on what i understood, i took out the logic where CompilerBuilder was directing which compiler class to invoke, instead i put this in separate method ``launch_compiler``.  I tried to implement all the common methods in CompilerBuilder and pass in compiler parameters from one of the Compiler Classes. For now i removed all the compiler class except for GNUCompiler for testing purposes, I will add them back once i get one compiler working. 

The issue i have right now is i am unable to pass ``cc`` and other compiler variables to ``_build_testcontent`` method in CompilerBuilder and it's mainly due to how i have designed the structure of the class and this where i need some help. 

When i build my tests, I don't get the compiler variables passed so ``gcc`` is not coming in the compile cmd line. 


```
(buildtest) siddiq90@DOE-7086392 serial % buildtest build -b compiler_schema_hello.yml

            Discovered Buildspecs              

/Users/siddiq90/Documents/tutorials/apps/serial/compiler_schema_hello.yml



setup: gcc None gfortran -Wall g++ None
setup: gcc -Wall gfortran None g++ None
setup: gcc None gfortran None g++ -Wall
setup: gcc -Wall gfortran None g++ None
_build_testcontent: None None None None None None
['./hello.f90.exe']
['#!/bin/bash', 'module purge && module load GCC', '-o hello.f90.exe /Users/siddiq90/Documents/tutorials/apps/serial/src/hello.f90', './hello.f90.exe']
compiler_schema_hello          hello_f                        FAIL                           /Users/siddiq90/Documents/tutorials/apps/serial/compiler_schema_hello.yml
_build_testcontent: None None None None None None
['./hello.c.exe']
['#!/bin/bash', '-o hello.c.exe /Users/siddiq90/Documents/tutorials/apps/serial/src/hello.c', './hello.c.exe']
compiler_schema_hello          hello_c                        FAIL                           /Users/siddiq90/Documents/tutorials/apps/serial/compiler_schema_hello.yml
_build_testcontent: None None None None None None
['./hello.cpp.exe']
['#!/bin/bash', '-o hello.cpp.exe /Users/siddiq90/Documents/tutorials/apps/serial/src/hello.cpp', './hello.cpp.exe']
compiler_schema_hello          hello_cplusplus                FAIL                           /Users/siddiq90/Documents/tutorials/apps/serial/compiler_schema_hello.yml
_build_testcontent: None None None None None None
['./argc.c.exe', '1 2 3']
['#!/bin/bash', '-o argc.c.exe /Users/siddiq90/Documents/tutorials/apps/serial/src/argc.c', './argc.c.exe 1 2 3']
compiler_schema_hello          passing_args                   FAIL                           /Users/siddiq90/Documents/tutorials/apps/serial/compiler_schema_hello.yml
```

vsoch:
My advice is the same - you should instantiate the GNUCompiler off the bat and not need a launch compiler function.

shahzebsiddiqui:
Can you be more specific, right now i am looking at way we add builders  which invoke the Compiler class on line 
https://github.com/shahzebsiddiqui/buildtest-1/blob/compiler_schema/buildtest/buildsystem/base.py#L184

vsoch:
I mean exactly to use the class, here so it's clear:
```python
elif recipe["type"] == "gnucompiler":
                    builders.append(
                        GNUCompiler(name, recipe, self.buildspec, testdir=testdir)
                    )
```
If you want to create a "compiler namespace" then you could also do something like:
```python
elif recipe["type"] == "compiler.gnu":
                    builders.append(
                        GNUCompiler(name, recipe, self.buildspec, testdir=testdir)
                    )
```

shahzebsiddiqui:
Thanks for the clarification, I like the compiler namespace as pose to a new compiler type. Will a new namespace have any implications to schema. I mean right now we only accept 
**type**  to be **compiler** so adding new options would be just modifying the type field and everything else stay same.
I guess we could remove **gnu** property and move all the attributes up one level. Does that make sense?

vsoch:
Yes I think a flatter representation is better, and that was my suggestion with having compiler.gnu or compiler-gnu or gnu-computer (name the type as you please!)

shahzebsiddiqui:
good news, i was able to get the test working, but i had to introduce a new key ``name``. I couldn't use ``type: compiler.gnu`` or something else since this was used by BuildSpecParser in self.lookup to fetch the schemas and lookup the name and version check. This caused key error since name of schema (compiler) is same value assigned to type. Hope that makes sense.

In this case, i decided not to drop the compiler namespace suggestion because that would  break existing functionality hence I added ``name`` key to compiler properties. The ``compiler`` key must take source and name as required fields. I have some examples for gnu and intel to illustrate what that looks like i haven't pushed this to tutorials yet.

```
(buildtest) siddiq90@DOE-7086392 serial % cat compiler_schema_hello.yml intel_hello.yml 
version: 0.0.1
hello_f:
  type: compiler
  description: "Hello World Fortran Compilation"
  module: 
    - "module purge && module load GCC"
  compiler:
    name: gnu
    source: "src/hello.f90"
    fflags: -Wall

hello_c:
  type: compiler
  description: "Hello World C Compilation"
  compiler:
    name: gnu
    source: "src/hello.c"
    cflags: -Wall

hello_cplusplus:
  type: compiler
  description: "Hello World C++ Compilation"
  compiler:
    name: gnu
    source: "src/hello.cpp"
    cxxflags: -Wall

passing_args:
  type: compiler
  description: Passing arguments example
  compiler:
    name: gnu
    source: "src/argc.c"
    exec_args: "1 2 3"
    cflags: -Wall

maintainers:
  - "@shahzebsiddiqui"
version: 0.0.1
hello_f:
  type: compiler
  description: "Hello World Fortran Compilation"
  module: 
    - "module purge && module load GCC"
  compiler:
    name: intel
    source: "src/hello.f90"
    fflags: -Wall

hello_c:
  type: compiler
  description: "Hello World C Compilation"
  compiler:
    name: intel
    source: "src/hello.c"
    cflags: -Wall

hello_cplusplus:
  type: compiler
  description: "Hello World C++ Compilation"
  compiler:
    name: intel
    source: "src/hello.cpp"
    cxxflags: -Wall
```

shahzebsiddiqui:
I will update the schema  PR https://github.com/buildtesters/schemas/pull/37 stay tuned. We can merge both of these at same time once this PR is approved. Feel free to review now, i am done.

buildtest/buildsystem/base.py
```diff
@@ -640,3 +655,294 @@ def _build_testcontent(self):
...
+
+        :param workdir: relative path to binary from working directory, if not specified it's assumed executable
+                        is in working directory.
+        :type workdir: str, optional
+        :return: A list containing path to executable
+        :rtype: list
+        """
+
+        if workdir:
+            return [f"{workdir}/{self.executable}"]
```
vsoch: os.path.join here

vsoch (changes requested):
Looks good! One tiny path issue, then with the schema in this is a solid start.

buildtest/buildsystem/base.py
shahzebsiddiqui: good call. Ill fix this

buildtest/buildsystem/base.py
```diff
@@ -787,7 +787,7 @@ def simple_run(self, workdir=None):
         """
 
         if workdir:
-            return [f"{workdir}/{self.executable}"]
+            return [os.path.join(workdir / self.executable)]
```
vsoch: Why not `os.path.join(workdir, self.executable)` ?

shahzebsiddiqui: the reason why i put as a list is in future there could be different types of run cmd right now its a single line, as a standalone executable or running executable with arguments. Running test via mpi, or debugging and profiling code could be multiline commands that  are post build. It would be easier to add items to list, and the ``build_run_cmd`` could be expanded to encapsulate different run scenarios. There could be a combination of YAML and similar to ``run`` key from ``script`` to allow for more customized tests.

vsoch: Sorry I wasn’t asking about the list - I’m asking about the /

shahzebsiddiqui: My thought was to allow flexibility in future if user want's to run a test not relative to their working directory, they can provide a relative directory. For example, building a binary with different flags  needs to have a unique executable name, user could have way to organize where executable is generated and other artifacts as a result of execution of test.

Right now it's possible two sections in Buildspec that reference same sourcefile generate the same executable. This is because the default name of executable is related to source file. One possibility is to have a key for user to specify name of executable, but that is unrelated to your comment.

vsoch: I’ll try again - why did you use a slash (/) instead of an expected comma for two args in is.path.join?

shahzebsiddiqui: wow that's embarrassing this is not suppose to be a / it is suppose to be a comma, i will fix this.

shahzebsiddiqui:
Anything else you would like me to address for this first draft. Once this is approved i will update docs and examples.

vsoch (approved):
If the schema is updated from the last PR, then I think this looks ready (for a first draft). I suspect you might change some things when you work more on it!

shahzebsiddiqui:
Yea i realized that the first implementation of compiler schema was to get the initial schema and examples validated in https://github.com/buildtesters/schemas/pull/36 but practically adding this to framework led to other changes. I think going forward, any schema updates should be coupled with a change in framework so that work is in sync and implemented in framework. 

Anyhow, if there aren't any further changes feel free to merge

PR #318, opened 2020-06-11
add getter methods for extracting repo entry and repo destination path

shahzebsiddiqui:
via get_repo_paths and active_repos.
Fix bug when adding repo where .git extension was added to entry, this
needed to be removed.
Remove searchpath key from settings.schema.json
Create slurm and ssh executor examples in separate .yml files and remove
these from settings.yml which is the default settings file for buildtest.
This was done because we expect buildtest to provide local executor as
default while other executors would require further customization by site.


shahzebsiddiqui:
This PR also address schema change in https://github.com/buildtesters/schemas/pull/43

PR #1066, opened 2022-06-02
Update urlchecker.yml

vsoch:
This one too!

Also @shahzebsiddiqui you can put two triggers in the same file to avoid the redundant workflows.


shahzebsiddiqui:
Thanks @vsoch i appreciate your help. I see there was a change in version numbers in according to the tag release https://github.com/urlstechie/urlchecker-action/tags.

PR #1069, opened 2022-06-03
Add summary field to schema

### LINKED ISSUE #1059: Add a 'summary' field in buildspec
The `summary` field can can be used to describe a test which can be used as extended description. Currently the `description` field is limited to 80 char defined in https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/definitions.schema.json

```
    "description": {
      "type": "string",
      "description": "The ``description`` field is used to document what the test is doing",
      "maxLength": 80
    },
```

The idea would be to include similar field without any `maxLength`. This would be applied to all schemas including `script`, `compiler`, and `spack`.

A typical test could be something like

```
buildspecs:
  summary_example:
    type: script
    description: Short description limited to 80 characters
    summary: |
      This is a long description of test that 
      can exceed 80 characters and be multiline 
   executor: generic.local.bash
   run: pwd
```

- [x] Update schema  https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/definitions.schema.json and add property to other schemas such as script schema https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/script.schema.json
- [x] Add an example test in https://github.com/buildtesters/buildtest/tree/devel/buildtest/schemas/examples to ensure schema validates with `summary` field. 
- [x] Ensure test pass via regression test. Run via `python $BUILDTEST_ROOT/buildtest/tools/unittest.py -c` or `buildtest unittests -c`. See https://buildtest.readthedocs.io/en/devel/contributing/regression_testing.html
- [x] Add example buildspec in https://github.com/buildtesters/buildtest/tree/devel/tutorials for documentation purposes. 
- [x] Document this field in https://github.com/buildtesters/buildtest/blob/devel/docs/buildspecs/buildspec_overview.rst

shahzebsiddiqui:


buildtest/schemas/definitions.schema.json
```diff
@@ -80,6 +80,10 @@
       "description": "The ``description`` field is used to document what the test is doing",
       "maxLength": 80
     },
+    "summary": {
+      "type": "string",
+      "description": "The ``description`` field is used to document what the test is doing"
```
shahzebsiddiqui: Change description to `summary ` that is surrounded in double tick marks

buildtest/schemas/examples/script.schema.json/invalid/examples.yml
```diff
@@ -41,6 +41,21 @@ buildspecs:
       - "Multi Line description"
       - "is not accepted"
 
+
+  invalid_summary_1:
+    type: script
```
shahzebsiddiqui, line 46: This test is valid so we should remove this one

buildtest/schemas/examples/script.schema.json/invalid/examples.yml
```diff
@@ -41,6 +41,21 @@ buildspecs:
...
+    executor: generic.local.bash
+    description: "test"
+    summary: "test"
+
+
+  invalid_summary_2:
+    type: script
+    executor: generic.local.bash
+    description: 1
```
shahzebsiddiqui: For description field I would provide a description of this test. We are not testing description field in this example.

Say for instance "summary field must be a string "

buildtest/schemas/examples/script.schema.json/valid/examples.yml
```diff
@@ -234,6 +234,15 @@ buildspecs:
     skip: true
     run: hostname
 
+  summary_example:
+    type: script
+    executor: generic.local.bash
+    description: Short description limited to 80 characters
```
shahzebsiddiqui: Change this description to 

"The summary field can be a multi-line string and exceed 80 char"

Note this test is not meant to test description its for summary. There is some tests above for description if you look in file above

shahzebsiddiqui (commented):
There are a few changes I'd like you to make. Overall this looks pretty good.

PR #1086, opened 2022-06-13
update bash completion and add aliases

shahzebsiddiqui:
This PR will add alias `buildtest debug` and `buildtest test`


Xiangs18 (approved):
It looks good to me!!

PR #1106, opened 2022-06-21
update section for test names section that includes a dash

shahzebsiddiqui:


shahzebsiddiqui:
@Xiangs18 adding documentation since we added the dash character in property names.

PR #1108, opened 2022-06-22
add build stamp "-Experimental" in CDASH upload

shahzebsiddiqui:
@zackgalbreath

Adding the `-Experimental` and tried pushing a cdash test https://my.cdash.org//build/2181181. Let me know if this is the field, i dont understand what are the valid types for build stamp and how this impacts cdash upload. Is there user documentation that can help clarify this so we can make informed decision in the buildtest codebase.

Screen Shot 2022-06-22 at 4 53 02 PM

### DIFF
--- a/buildtest/cli/cdash.py
+++ b/buildtest/cli/cdash.py
@@ -255,6 +255,7 @@ def upload_test_cdash(
     console.print(f"Uploading {len(tests)} tests")
 
     build_stamp = build_starttime.strftime(output_datetime_format)
+    build_stamp += "-Experimental"
 
     filename = "Test.xml"
     site_element = ET.Element(


zackgalbreath:
LGTM!

The important bit in the screenshot here is "Type: Experimental". It doesn't _really_ matter what you put here, as long as it's consistent from one build to the next.

Prior to this change, CDash was detecting a portion of the date string as the type, so it would change every time, and we lost the ability to navigate from one build to the next.

shahzebsiddiqui:
So previously I updated the build stamp by removing the `Experimental` and I suspect that may have caused some data to be lost. 

Are there are reserved words for this Type field what values should we use. Is this documented anywhere?

zackgalbreath:
> Are there are reserved words for this Type field what values should we use. Is this documented anywhere?

The closest documentation we currently have to this behavior is for the `MODEL` and `GROUP` options to  [`ctest_start()`](https://cmake.org/cmake/help/latest/command/ctest_start.html).

The default values that CTest/CDash know about for this field are 'Nightly', 'Experimental', and 'Continuous'. You can also create your own custom buildgroups on CDash and insert their name into this field as well.

PR #1121, opened 2022-06-28
add regression test for buildtest stats

### LINKED ISSUE #1075: add regression test for buildtest stats
We added support for new command ``buildtest stats`` see https://github.com/buildtesters/buildtest/pull/1074. We need to add regression test for this.

This can be done by creating a test in https://github.com/buildtesters/buildtest/tree/devel/tests/cli with a filename `test_stats.py`.

For this task one needs to check regression test results to make sure we get enough coverage o file.

Xiangs18:
Still need to pass **python $BUILDTEST_ROOT/buildtest/tools/unittests.py -c**

Screen Shot 2022-06-28 at 5 41 31 PM

### DIFF
--- a/buildtest/cli/stats.py
+++ b/buildtest/cli/stats.py
@@ -1,8 +1,7 @@
-from statistics import StatisticsError, mean, variance
+from statistics import mean, variance
 
 from buildtest.cli.report import Report
 from buildtest.defaults import console
-from buildtest.exceptions import BuildTestError
 
 
 def stats_cmd(name, report_file=None):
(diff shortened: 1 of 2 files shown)


Xiangs18:
Hi @shahzebsiddiqui, please review and see if further changes are required.
Screen Shot 2022-06-29 at 12 43 22 PM

shahzebsiddiqui (approved):
Thanks @Xiangs18 this looks good as discussed i think this gets rid of the exception when running `buildtest stats` on a test with single run so for now this is good.

PR #1130, opened 2022-06-30
limit output of buildtest report

### LINKED ISSUE #1128: [FEATURE]: limit output of buildtest report

The `buildtest report` command will show all test results and the output can be quite long so for this feature we would like to limit the output by limiting the number of rows that get printed. 

We can have an option `buildtest report --count=<INT>` where `--count` expects a positive integer that limits the query to X number of rows.  

- [x] Add arg parse argument
- [x] Add implementation
- [x] Add bash completion
- [x] Add regression test
- [x] Add documentation
- [x] Add help config section

### Post question in Slack

Xiangs18:
Hi @shahzebsiddiqui, please see if further changes are required. Thank you!

Screen Shot 2022-06-30 at 11 35 06 AM


shahzebsiddiqui:
@Xiangs18 this looks good. Can you share the output i think this looks perfect.

I presume this won't work with terse mode `buildtest rt --terse --count=1`

Xiangs18:
Hi @shahzebsiddiqui, this won't work with terse mode buildtest rt --terse --count=1
However, it works with pager mode buildtest rt --pager --count 1
If you wanna make it work with terse mode as well, I can include it.

Xiangs18:
Hi @shahzebsiddiqui, now it works with terse mode as well. Thank you!

tests/cli/test_report.py
```diff
@@ -232,3 +232,25 @@ class args:
...
+        report_file = None
+        pager = None
+        terse = None
+        no_header = None
+        report_subcommand = None
+        helpfilter = False
+        helpformat = False
+        count = 5
+
+    report_cmd(args)
```
shahzebsiddiqui: Can you add another test with terse mode so just create another invocation to `args` . Alternately we may want to rewrite this test to directly call `Report` class directly. Either way works

shahzebsiddiqui (changes requested):

tests/cli/test_report.py
Xiangs18: Screen Shot 2022-06-30 at 1 24 12 PM

Xiangs18:
Screen Shot 2022-06-30 at 1 51 20 PM

shahzebsiddiqui (approved):
LGTM

PR #1143, opened 2022-07-01
add regression test for editing buildspecs

### LINKED ISSUE #1102: add regression test for 'buildtest buildspec edit-file' and 'buildtest buildspec edit-test'
@JeffreyFG here is the next task for you. 

According to codecov https://app.codecov.io/gh/buildtesters/buildtest/blob/devel/buildtest/cli/buildspec.py we dont have coverage in regression test for the following methods `edit_buildspec_test`  and `edit_buildspec_file`. Lines in red are missing coverage

Screen Shot 2022-06-17 at 12 34 41 PM

Screen Shot 2022-06-17 at 12 34 54 PM

Since these methods expect opening file in editor we can't run these during regression test however we can have an optional argument to each method called `open=True` to open file which is the default and `open=False` is set during regression test which will not open file. 

The logic of code would be changed such that you can do 

```
if not open:
  return
```

in the regression test we would call the method like `edit_buildspec_file(buildspecs, configuration, open=False)`

TODO
- [ ] Add  regression test in https://github.com/buildtesters/buildtest/blob/devel/tests/cli/test_buildspec.py. Define new method `test_edit_test` and `test_edit_file`

JeffreyFG:
…or the regression testing from test_edit_test and test_edit_file
.
Me and @Xiangs18 had an interesting discussion on the best way to call edit_buildspec_file and edit_buildspec_test in the regression test. I left both positional arguments editor and buildspecs null because I knew that the open=False would stop that from mattering. @Xiangs18 was rightfully concerned about this being very un standard. @shahzebsiddiqui Do you think what I did was appropriate. And if I should create builspecs for this file what would they be.


coverage.json
```diff
@@ -0,0 +1 @@
...ches": 22, "num_partial_branches": 2, "covered_branches": 20, "missing_branches": 2}, "missing_lines": [30, 62], "excluded_lines": [194, 195, 197, 198]}, "buildtest/utils/timer.py": {"executed_lines": [1, 4, 5, 8, 9, 10, 11, 13, 15, 18, 20, 22, 25, 26, 29, 30], "summary": {"covered_lines": 15, "num_statements": 17, "percent_covered": 84.0, "percent_covered_display": "84.00", "missing_lines": 2, "excluded_lines": 0, "num_branches": 8, "num_partial_branches": 2, "covered_branches": 6, "missing_branches": 2}, "missing_lines": [16, 23], "excluded_lines": []}, "buildtest/utils/tools.py": {"executed_lines": [1, 4, 5, 12], "summary": {"covered_lines": 3, "num_statements": 3, "percent_covered": 100.0, "percent_covered_display": "100.00", "missing_lines": 0, "excluded_lines": 1, "num_branches": 2, "num_partial_branches": 0, "covered_branches": 2, "missing_branches": 0}, "missing_lines": [], "excluded_lines": [12]}}, "totals": {"covered_lines": 4258, "num_statements": 5484, "percent_covered": 75.7527311484146, "percent_covered_display": "75.75", "missing_lines": 1226, "excluded_lines": 175, "num_branches": 2022, "num_partial_branches": 222, "covered_branches": 1428, "missing_branches": 594}}
```
shahzebsiddiqui: Please remove this file and please add this in .gitignore so it's excluded

tests/cli/test_buildspec.py
```diff
@@ -150,6 +152,19 @@ def test_buildspec_find_invalid():
     cache.print_invalid_buildspecs(error=False, terse=True, header=True)
 
 
+@pytest.mark.cli
+def test_edit_test():
+    # cache.print
+    buildspecs = []
```
shahzebsiddiqui: No need to declare here u can call it directly in function call `edit_buildspec_test(buildspecs=[], ...)`

shahzebsiddiqui (changes requested):
Overall this looks correct.

Can u also update the docstring to include the new argument so it shows up in the API docs.

shahzebsiddiqui:
@JeffreyFG just one tip when opening PR please update the title of PR to something meaningful since github uses the commit message and that doesn't make sense.

Also if u see a PR address an issue u can also link that in PR.

buildtest/cli/buildspec.py
```diff
@@ -953,7 +953,10 @@ def edit_buildspec_test(test_names, configuration, editor):
         test_names (list): A list of test names to open in editor
         configuration (buildtest.config.SiteConfiguration): An instance of SiteConfiguration class
         editor (str): Path to editor to use when opening file
+        open (bool): Variable allows for more accurate coverage tracking in regression testing.
```
shahzebsiddiqui: I think the docstring is a bit misleading you can state the following

```
open (bool): A boolean to determine whether to open file for editing, by default `open=True` indicates file will be open in editor. If its False this method will return immediately which is useful when running regression test.
```

shahzebsiddiqui:
@JeffreyFG i realize that the regression test will not test coverage for the entire method simply because we have it only set to `open=False` if we set to `open=True` it won't work right now since it will just hang indefinitely.

One idea i had in mind, we would need to refactor the original implementation by changing the `subprocess.call` to `BuildTestCommand` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/buildspec.py#L946 

Screen Shot 2022-07-06 at 9 41 25 PM


Note that this class *execute* method has an optional argument `timeout` which can be set to something like 1 or 2 second during the regression test.

https://github.com/buildtesters/buildtest/blob/7eba9872f0ee5ea3a71b8ee5355f3410fc72f793/buildtest/utils/command.py#L88

We would need to pass the `timeout` argument to the method `edit_buildspec_test` which by default should be `timeout=None`. I hope that makes sense. Please try this out and run the regression test locally and see if the coverage report captures entire method.

tests/cli/test_buildspec.py
```diff
@@ -150,6 +152,16 @@ def test_buildspec_find_invalid():
     cache.print_invalid_buildspecs(error=False, terse=True, header=True)
 
 
+@pytest.mark.cli
+def test_edit_test():
+    edit_buildspec_test(configuration, buildspecs=[], editor="", test=True)
```
shahzebsiddiqui: You can try calling a valid buildspec you can do something like `buildspecs = [os.path.join(BUILDTEST_ROOT,"tutorials","vars.yml")]`

tests/cli/test_buildspec.py
```diff
@@ -150,6 +152,24 @@ def test_buildspec_find_invalid():
     cache.print_invalid_buildspecs(error=False, terse=True, header=True)
 
 
+@pytest.mark.cli
+def test_edit_test():
+    edit_buildspec_test(
+        buildspecs=[os.path.join(BUILDTEST_ROOT, "tutorials", "vars.yml")],
```
shahzebsiddiqui: so this method doesn't accept `buildspecs` but instead you should use one of the valid test names. See the method signature, you need to use `test_names` to pass the test name which takes the argument for name of test when you say `buildtest bc edit-test hello_world` the name of test is `hello_world`.

shahzebsiddiqui: i believe `hello_world` is a test name that should exist you can call this method like
```
    edit_buildspec_test(
        test_names=["hello_world",],
        configuration=configuration,
        editor='',
    )
```

JeffreyFG: This was the code snipet that you gave to use in this spot. So instead i should just give the method call a test name like hello world? 
IG: edit_buildspec_test(buildspecs="hello_world",configuration=configuration,editor="")?

PR #1147, opened 2022-07-05
update black version in pre-commit configuration and pass --verbose in CI check

shahzebsiddiqui:


shahzebsiddiqui:
@JeffreyFG @Xiangs18 @prathmesh4321 please install the latest version of black (22.6.0) in your python environment as per https://github.com/buildtesters/buildtest/pull/1146 PR by dependabot. 

This will affect the pre-commit configuration when you do `pre-commit install` it will use the same version of black.

PR #1149, opened 2022-07-05
Issue uploading results to CodeCov

### LINKED ISSUE #1063: issue uploading results to CodeCov
There is an issue uploading test results to codecov. If you take a look at https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true you will see the following message

```
Run codecov/codecov-action@v3.1.0
==> macos OS detected
https://uploader.codecov.io/latest/macos/codecov.SHA2[5](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:6)[6](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:7)SUM
==> SHASUM file signed by key id 806bb28aed[7](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:8)79[8](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:9)69
==> Uploader SHASUM verified (8d3709d957c7115610e764621569728be102d213fee15bc1d1aa9d465eb2c258  codecov)
==> Running version latest
==> Running version v0.2.3
/Users/runner/work/_actions/codecov/codecov-action/v3.1.0/dist/codecov -n  -Q github-action-3.1.0 -C 6[9](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:10)8d95d6cc37213f9bdd6189606c378071ca885a
[2022-05-27T20:09:16.962Z] ['info'] 
     _____          _
    / ____|        | |
   | |     ___   __| | ___  ___ _____   __
   | |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
   | |___| (_) | (_| |  __/ (_| (_) \ V /
    \_____\___/ \__,_|\___|\___\___/ \_/

  Codecov report uploader 0.2.3
[2022-05-27T20:09:[16](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:17).977Z] ['info'] => Project root located at: /Users/runner/work/buildtest/buildtest
[2022-05-27T20:09:16.983Z] ['info'] -> No token specified or token is empty
[2022-05-27T20:09:[17](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:18).[20](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:21)0Z] ['info'] Searching for coverage files...
[20[22](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:23)-05-27T20:09:17.[24](https://github.com/buildtesters/buildtest/runs/6631262688?check_suite_focus=true#step:7:25)7Z] ['info'] Warning: Some files located via search were excluded from upload.
[2022-05-27T20:09:17.247Z] ['info'] If Codecov did not locate your files, please review https://docs.codecov.com/docs/supported-report-formats
[2022-05-27T20:09:17.247Z] ['error'] There was an error running the uploader: No coverage files located, please try use `-f`, or change the project root with `-R`
```

This github action is coming from https://github.com/codecov/codecov-action and is used in https://github.com/buildtesters/buildtest/blob/devel/.github/workflows/regression.yml

Possibly bring this up with developer by reporting an issue at https://github.com/codecov/codecov-action/issues or try a new version (https://github.com/codecov/codecov-action/releases). Currently we are using 3.1.0 which is the latest release as of now. Not sure if downgrading to 3.0.0 will work.

Xiangs18:


shahzebsiddiqui:
@Xiangs18 I see you have tried several things it is still not working. So it is most likely we still need to use the latest version of codecov action - codecov-action@v3

According to their docs https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage they use `pytest --cov .` to get the coverage which i used to do however we scripted this up.

If you want to try the XML route you need to update the python regression test script to use `xml_report()` see https://coverage.readthedocs.io/en/6.4.1/api_coverage.html#coverage.Coverage.xml_report. 

Previously codecov had the bash uploaded script but this was deprecated therefore they recommend using codecov uploader see https://docs.codecov.com/docs/codecov-uploader. This can be a script one can download to upload the codecov results. For public projects token is not required however if you need one i can help get that to you.

According to https://docs.codecov.com/docs/supported-languages they support several languages i tried looking at bash and python example repos. This is their example https://github.com/codecov/example-python/blob/main/.github/workflows/ci.yml however i can see they have a working upload script https://github.com/codecov/example-python/runs/6824315539?check_suite_focus=true 

If you need help please loop in their developer, i am kinda out of ideas. I was thinking of testing the codecov uploader script manually and then try it in the CI job but it doesn't make any sense since the coverage file is already generated and it should just work.

Xiangs18:
So basically this is the problem: 

extra line of - uses: actions/checkout@v2 will delete dirs we created to avoid conflicts, including all reports. 

remove actions/checkout@v2 immediately before upload would do the job.

.github/workflows/regression.yml
```diff
@@ -16,10 +16,10 @@ jobs:
         python-version: ['3.7', '3.8', '3.9', '3.10']
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
```
shahzebsiddiqui: why are we using checkout@v2 why not v3 this is the latest version

.github/workflows/regression.yml
```diff
@@ -16,10 +16,10 @@ jobs:
...
 
     - name: Setup Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3.1.2
+      uses: actions/setup-python@v2
```
shahzebsiddiqui: why are we using `v2` here according to https://github.com/actions/setup-python we are at v4

.github/workflows/regression.yml
Xiangs18: I did not try v3. v2 is the version used in this public official demo: https://github.com/codecov/example-python/blob/0e76edc4dae15ead94098ab331931cc36eb8f03d/.github/workflows/ci.yml#L8

shahzebsiddiqui (changes requested):
can u try the latest version and if they work we should be good to go otherwise we may need to go down some version. The dependabot will notify us with version changes but i understand some of these github action may have some requirement on other versions.

.github/workflows/regression.yml
Xiangs18: I think v3 also works

shahzebsiddiqui: i think u should keep this to `@v3` unless it breaks. I think they haven't updated their CI job at time of writing their job. That's what dependabot will do by sending PRs when each of these actions has a new version. Sometimes they have breaking changes we need to see release notes.

Xiangs18: Now is the latest.

shahzebsiddiqui:
thanks @Xiangs18 i am merging this in. I appreciate your help this bug was annoying to fix.

PR #1166, opened 2022-07-12
Add Spelling check CI

### LINKED ISSUE #1164: Spelling check CI
Everyone makes spelling mistakes whether it's documentation, comments in code, etc...

In that case let's fix all of these. Found a githuv action https://github.com/marketplace/actions/run-pyspelling-as-a-github-action let's integrate this into the workflow

Xiangs18:


Xiangs18:
Hi @shahzebsiddiqui, currently spellcheck runner has been integrated as part of Github Action workflow. However, is still failing because PySpelling detected a lot of [false positive cases](https://github.com/buildtesters/buildtest/runs/7310665849?check_suite_focus=true). We can either add all these words into .wordlist.txt so that these cases can be ignored or we need to kinda go through them one by one to decide.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  spellcheck:
+    name: runner / spellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code.
+      uses: actions/checkout@v3
+    - name: Run PySpelling as a GitHub action
+      uses: sbates130272/spellcheck-github-actions@v0.7.0
```
shahzebsiddiqui: lets change this workflow with the new github action https://github.com/marketplace/actions/github-spellcheck-action

Xiangs18:
Hi @shahzebsiddiqui, workflow is now working with the new github action. Currently, we only check on README.rst and CHANGELOG.md files. Please let me know if you want to include other folders/docs. Thank you!

.wordlist.txt
```diff
@@ -0,0 +1,63 @@
...
+cori
+csh
+devel
+doi
+ef
+gh
+github
+herokuapp
+hpcbuildtest
+hpcbuildtest
```
shahzebsiddiqui: why is this mentioned twice

shahzebsiddiqui:
Overall this looks good and checks are passing. Let's try to add the python directory for the code-base `buildtest` and then let's see what errors come up.

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,24 @@
+name: Spellcheck Action
+
+on:
+  push:
+    branches:
+      - master
+      - devel
```
shahzebsiddiqui: i think we can disable the checks on push just limit to PRs is fine. So you can remove this section

Xiangs18: sounds good. will fix

.wordlist.txt
Xiangs18: will fix. Thank you!

.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  build:
+    name: Spellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: rojopolis/spellcheck-github-actions@0.25.0
+      name: Spellcheck
+      with:
+        source_files: README.rst CHANGELOG.md
```
shahzebsiddiqui: its `CHANGELOG.rst` not `CHANGELOG.md` in the repo. This may cause some errors in spell check since its a really big file

Xiangs18:
Hi @shahzebsiddiqui, I limited scope to README.rst and docs/ folder now. There are a few spelling errors that you can see here: https://github.com/buildtesters/buildtest/runs/7523477908?check_suite_focus=true. We can either add all/part of these words into wordlist.txt or correct docs instead. Please let me know your thoughts.

shahzebsiddiqui:
I am trying to see the useful of this CI check i think for most part there are a ton of errors that are not useful however i do see a few such as this one below where there is a spelling error `recieved` -> `received`

In that case we can fix the errors later and refine this list with the words such that we can avoid as much of these errors. We can let this check fail and over time the spelling error will cut down. Overall i think this check can still be useful even with alot of noise at first. 

Screen Shot 2022-07-26 at 12 45 56 PM

.spellcheck.yml
```diff
@@ -0,0 +1,40 @@
...
+  - pyspelling.filters.python:
+      comments: false
+      docstrings: false
+      group_comments: false
+      decode_escapes: true
+      strings: false
+      string_types: fu
+  sources:
+  - '**/*.py'
+  default_encoding: utf-8
```
shahzebsiddiqui: I think we should not check any python scripts it is bound to failure and i dont think anyone would care if there is spelling errors if its in comments or docstrings. In that case can we remove this section to avoid any further noise

Xiangs18:
Hi @shahzebsiddiqui, I have made made changes based on your suggestions. Please review again!

.spellcheck.yml
```diff
@@ -0,0 +1,20 @@
...
+  pipeline:
+  - pyspelling.filters.markdown:
+  - pyspelling.filters.html:
+      comments: false
+      ignores:
+      - code
+      - pre
+  sources:
+  - 'README.rst'
+  - 'docs/*.rst'
```
shahzebsiddiqui, line 19: i think this regex is an issue this will only search the top-level directory for `*.rst` but we may have more sub-directories. 

Can you just specify the directory so say `- docs`

Xiangs18: Hi @shahzebsiddiqui, neither docs nor docs/ works on my end. I added 'docs/**/*.rst' instead. Thank you!

Xiangs18:
Hi @shahzebsiddiqui, updated. Please review again!

shahzebsiddiqui (approved):

PR #1174, opened 2022-07-26
rename options for buildtest last log

shahzebsiddiqui:
@Xiangs18 @prathmesh4321 please take a look. I renamed some of the options since it was bit too long we can presume these options will print last log we have in the help message.

I caught an exception when logfile is not found instead of reading the file.

Also i made sure we exit program by `return` which i noticed wasn't set in that case one could do something like `buildtest --view-log bd -t pass` and run some commands after viewing log.


Xiangs18 (approved):
Hi @shahzebsiddiqui, LGTM!

prathmesh4321 (approved):
Hi @shahzebsiddiqui, looks good to me!

PR #1175, opened 2022-07-26
Automatically test compiler instances via module load test

### LINKED ISSUE #1159: [FEATURE]: Automatically test compiler instances via module load test

Add a new command `buildtest config compilers test` to allow one to test the compilers with the corresponding `modules` if set. We should also check the full path to option `cc`, `fc`, `cxx` to ensure that path is valid since one can specify arbitrary path. This should be relatively easy but running a `shutil.which` should do the trick.

Note we already test the `module load` operation prior to generating the configuration when running `buildtest config compilers find` based on regular expression. This logic is implemented in https://github.com/buildtesters/buildtest/blob/7eba9872f0ee5ea3a71b8ee5355f3410fc72f793/buildtest/cli/compilers.py#L130-L239

This command should iterate over all compilers and perform the `module load` test and show an output of each compiler with a `PASS` or `FAIL` next to each compiler section. The command `buildtest config compilers test` should not take any additional argument, it just test all the compilers.

If no compilers are set we return immediately.

The implementation of module test is done via lmodule see https://lmodule.readthedocs.io/en/latest/index.html

### Post question in Slack

Xiangs18:


Xiangs18:
Hi @shahzebsiddiqui, does this table display meet your requirement? Please review and advise. Thank you!

Screen Shot 2022-07-28 at 2 31 39 PM

buildtest/cli/compilers.py
```diff
@@ -34,6 +40,42 @@ def compiler_cmd(args, configuration):
         bc.print_yaml()
 
 
+def compiler_test(args, configuration):
+    """This method implements ``buildtest config compilers test`` which tests
+    the compilers with the corresponding modules if set. This command iterates
+    over all compilers and perform the module load test and show an output of
+    each compiler with a PASS or FAIL next to each compiler section.
+    """
```
shahzebsiddiqui: Please add the `args` for the docstring

buildtest/cli/compilers.py
```diff
@@ -34,6 +40,42 @@ def compiler_cmd(args, configuration):
...
+    table.add_column("Compiler Name", style="red")
+    table.add_column("Status", justify="right")
+
+    count_id = 1
+    for compiler_cat in bc.compiler_modules_lookup_fail:
+        for compiler in bc.compiler_modules_lookup_fail[compiler_cat]:
+            table.add_row(str(count_id), compiler, "❌")
+            count_id += 1
+
+    console.print(table)
```
shahzebsiddiqui: add if statement saying `if table.row_count` then print the table https://rich.readthedocs.io/en/stable/reference/table.html#rich.table.Table.row_count

buildtest/cli/compilers.py
```diff
@@ -34,6 +40,42 @@ def compiler_cmd(args, configuration):
         bc.print_yaml()
 
 
+def compiler_test(args, configuration):
+    """This method implements ``buildtest config compilers test`` which tests
+    the compilers with the corresponding modules if set. This command iterates
+    over all compilers and perform the module load test and show an output of
+    each compiler with a PASS or FAIL next to each compiler section.
```
shahzebsiddiqui: you can remove `with a PASS or FAIL next to each compiler section.` since we are showing a check mark or a `x`

buildtest/cli/compilers.py
```diff
@@ -34,6 +40,42 @@ def compiler_cmd(args, configuration):
         bc.print_yaml()
 
 
+def compiler_test(args, configuration):
```
shahzebsiddiqui: please remove `args` and just pass the `debug` option

buildtest/cli/__init__.py
```diff
@@ -856,6 +856,11 @@ def config_menu(subparsers):
         help="Update configuration file with new compilers",
     )
 
+    subparsers_compiler.add_parser(
+        "test",
+        help="Test compilers",
```
shahzebsiddiqui: Say `Test each compiler instance by performing module load test`

buildtest/cli/compilers.py
```diff
@@ -13,7 +13,9 @@
 from buildtest.utils.tools import deep_get
 from lmod.module import Module
 from lmod.spider import Spider
+from rich.console import Console
```
shahzebsiddiqui: please remove this and call `from buildtest.defaults import console`

buildtest/cli/compilers.py
shahzebsiddiqui: we dont need to do this since we remove the argument

shahzebsiddiqui:
@Xiangs18 i think the only remaining item is to add a regression test for this method. I think what you should do is add a test in https://github.com/buildtesters/buildtest/blob/devel/tests/test_cori.py which will run a simple test to validate `buildtest bc compilers test` since only way you can run this test is if you have a module system.

Once this is done we can merge this MR

Xiangs18:
Hey @shahzebsiddiqui, please let me know if I need to add this new command to docs or summary table. Thank you!

Screen Shot 2022-08-01 at 3 47 02 PM

shahzebsiddiqui:
Yes might as well add it to docs and summary table. For the docs you may want to run it locally on Cori i dont think you can get it auto-generated but it could be worth a try to see if it works in readthedocs build.

Xiangs18:
You can visualize docs here: https://buildtest--1175.org.readthedocs.build/en/1175/configuring_buildtest/compilers.html#test-compilers-experimental-feature

PR #1188, opened 2022-08-08
change behavior of 'buildtest config compilers find'

### LINKED ISSUE #1184: [FEATURE]: change behavior of 'buildtest config compilers find'

Right now we print the output of the configuration file after running the `buildtest config compilers find`. Let's not print this large file after running command instead we should just print a list of compilers detected. Eventually one will want to update the configuration file so `buildtest config compilers find --update` will overwrite the configuration file. We don't need to see content of file because we know it will be fine.

### Post question in Slack

Xiangs18:


Xiangs18:
Hey @shahzebsiddiqui, please let me know if you prefer a table format. Thank you!

shahzebsiddiqui:
I think output is good as is.

PR #1193, opened 2022-08-11
Fixed CI test failing

### LINKED ISSUE #1191: [Bug]: CI test failing at `buildtest buildspec find invalid`
### Bug Description

see https://github.com/buildtesters/buildtest/runs/7794092682?check_suite_focus=true 

Problem is that we now get `buildtest bc find invalid` to raise SystemExit and this causes CI check to fail so we need to fix this 

The regression test is in 

https://github.com/buildtesters/buildtest/blob/3ead042cbb987d7854afa17ef187159c4ec5a35a/tests/cli/test_buildspec.py#L145-L152

```
            # return a non-zero returncode
            if table.row_count > 0:
                console.print(table)
>               sys.exit(1)
E               SystemExit: 1

```

### Version

devel

### Relevant log output

_No response_

### Post question in Slack

prathmesh4321:
The test passes now!

Screen Shot 2022-08-11 at 4 33 03 PM

### DIFF
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -874,6 +874,7 @@ class BuildspecCache:
 
         if error and terse:
             console.print("The --terse flag can not be used with the --error option")
+            sys.exit(1)
             return
 
         # implementation for machine readable format specified via --terse
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -882,6 +882,7 @@ class BuildspecCache:
                 print("buildspec")
             for buildspec in self.cache["invalids"].keys():
                 print(buildspec)
+            sys.exit(1)
             return
 
         # if --error is not specified print list of invalid buildspecs in rich table
@@ -907,6 +908,7 @@ class BuildspecCache:
         for buildspec, value in self.cache["invalids"].items():
             console.rule(buildspec)
             pprint(value)
+        sys.exit(1)
 
     @staticmethod
     def print_filter_fields():
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -875,7 +875,6 @@ class BuildspecCache:
         if error and terse:
             console.print("The --terse flag can not be used with the --error option")
             sys.exit(1)
-            return
 
         # implementation for machine readable format specified via --terse
         if terse:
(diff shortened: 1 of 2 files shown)


prathmesh4321:
Screen Shot 2022-08-12 at 10 54 49 AM

tests/cli/test_buildspec.py
```diff
@@ -144,12 +144,15 @@ def test_buildspec_maintainers():
...
-    cache.print_invalid_buildspecs(error=False, terse=True, header=True)
+
+    # testing buildtest buildspec find invalid. This will assert SystemExit exception raised by sys.exit
+    with pytest.raises(SystemExit):
+        cache.print_invalid_buildspecs(error=True)
+        cache.print_invalid_buildspecs(error=False)
+        cache.print_invalid_buildspecs(error=True, terse=True)
+        cache.print_invalid_buildspecs(error=False, terse=True)
+        cache.print_invalid_buildspecs(error=True, terse=True, header=True)
+        cache.print_invalid_buildspecs(error=False, terse=True, header=True)
```
shahzebsiddiqui: i think for this you should invoke each call in its own context manager via `with` since if we decide to change the functionality of this method we want to make sure that the test works as expected. Right now this code will execute the first invocation raise an exception and we can't determine if all other calls will raise exception.

I think in future we will want to raise exception only if there is one or more entry that is invalid otherwise dont return an exit 1.

prathmesh4321: Done !

prathmesh4321: Done! @shahzebsiddiqui

buildtest/cli/buildspec.py
```diff
@@ -874,6 +874,7 @@ def print_invalid_buildspecs(self, error=None, terse=None, header=None):
 
         if error and terse:
             console.print("The --terse flag can not be used with the --error option")
+            sys.exit(1)
             return
```
shahzebsiddiqui: we can remove this `return` statement since this statement will never be reached

buildtest/cli/buildspec.py
```diff
@@ -882,6 +883,7 @@ def print_invalid_buildspecs(self, error=None, terse=None, header=None):
                 print("buildspec")
             for buildspec in self.cache["invalids"].keys():
                 print(buildspec)
+            sys.exit(1)
             return
```
shahzebsiddiqui: same with this one, this `return` statement will never be reached

PR #1197, opened 2022-08-15
Fixed the case when there no invalid buildspecs at all

prathmesh4321:
**Case when there are no invalid buildspecs system exit is not raised.**

Screen Shot 2022-08-15 at 12 15 26 PM

**Case when there are invalid buildspecs** 

Screen Shot 2022-08-15 at 12 14 20 PM

### DIFF
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -882,7 +882,9 @@ class BuildspecCache:
                 print("buildspec")
             for buildspec in self.cache["invalids"].keys():
                 print(buildspec)
-            sys.exit(1)
+            if self.cache["invalids"]:
+                sys.exit(1)
+            return
 
         # if --error is not specified print list of invalid buildspecs in rich table
         if not error:
@@ -907,7 +909,9 @@ class BuildspecCache:
         for buildspec, value in self.cache["invalids"].items():
             console.rule(buildspec)
             pprint(value)
-        sys.exit(1)
+        if self.cache["invalids"]:
+            sys.exit(1)
+        return
 
     @staticmethod
     def print_filter_fields():
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -886,6 +886,7 @@ class BuildspecCache:
                 print("buildspec")
             for buildspec in self.cache["invalids"].keys():
                 print(buildspec)
+            sys.exit(1)
 
         # if --error is not specified print list of invalid buildspecs in rich table
         if not error:
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -876,15 +876,16 @@ class BuildspecCache:
             console.print("The --terse flag can not be used with the --error option")
             sys.exit(1)
 
+        if not self.get_invalid_buildspecs():
+            console.print("There are no invalid buildspecs in cache")
+            return
+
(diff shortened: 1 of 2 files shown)


tests/cli/test_buildspec.py
```diff
@@ -146,23 +146,35 @@ def test_buildspec_find_invalid():
...
+    try:
         cache.print_invalid_buildspecs(error=False)
+    except SystemExit:
+        assert True
 
-    with pytest.raises(SystemExit):
+    try:
         cache.print_invalid_buildspecs(error=True, terse=True)
+    except SystemExit:
+        assert True
```
shahzebsiddiqui: i think this is right thinking, what i would like to have is some getter method that can retrieve invalid specs when running `cache.print_invalid_ buildspecs`. Since we already have `self.cache["invalids"]` to get the invalids so if the key is empty we know there are no entry therefore we return from method without raising exception. What we don't have is a getter method. One has do something like this assuming `bc` is an object of type `BuildspecCache`

```
  if bc.cache['invalids']: 
    print("There are no invalid buildspecs")
```

We can have some method like `get_invalid_buildspecs` so one can just do `bc.get_invalid_buildspecs` to get a list of buildspecs that are invalid as a list.

shahzebsiddiqui: Also i think a `assert True` is not necessary but since you need a statement in the except you may try to assert the exception type or run `continue`. The `assert` statement is useless

prathmesh4321: I guess 'continue' keyword works only with loops. Instead adding 'pass' is fine ? 
`except SystemExit:
        pass`

shahzebsiddiqui: yea i think `pass` should be fine

prathmesh4321: I guess we already have that function here 
https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/buildspec.py#:~:text=def%20get_invalid_buildspecs(,%5D.keys()

buildtest/cli/buildspec.py
```diff
@@ -882,7 +882,9 @@ def print_invalid_buildspecs(self, error=None, terse=None, header=None):
                 print("buildspec")
             for buildspec in self.cache["invalids"].keys():
                 print(buildspec)
-            sys.exit(1)
+            if self.cache["invalids"]:
+                sys.exit(1)
```
shahzebsiddiqui: so i think we can refactor this codebase now that i think about it you can call at beginning of method  after these lines

```
        if error and terse:
            console.print("The --terse flag can not be used with the --error option")
            sys.exit(1)
```

The following check for any invalid buildspecs and print a message and just return. Something like this

```
if not self.get_invalid_buildspecs():
  console.print("There are no invalid buildspecs in cache")
  return
```

Now after this statement we know any subsequent command will imply there is an invalid buildspec in cache and we can call `sys.exit(1)` at end of method instead of doing a conditional check. Can you please refactor this method i hope i make myself clear.

prathmesh4321: Yes Right! I have refactored it. Also, I guess we donot require to check for the number of rows in the table in error=false condition so I have remove it.
`   # return a non-zero returncode
            if table.row_count > 0:
                console.print(table)
                sys.exit(1)
            return`

shahzebsiddiqui:
Looking good thanks for fixing this

PR #1205, opened 2022-08-23
Specify MODULEPATH when searching for compilers

### LINKED ISSUE #1185: specify MODULEPATH when searching for compilers
Currently buildtest will search for compilers (`buildtest cg compilers find`) via MODULEPATH and it will use whatever the system default MODULEPATH is set. We can allow this to by dynamically changed during the search since MODULEPATH can be a list of many directories and running `buildtest cg compilers find` or `buildtest cg compilers test` can test a very long time.


Therefore i propose the following we add an optional argument ``buildtest cg compilers --modulepath=<dir1>:<dir2>`` set MODULEPATH prior to finding compilers.

Secondly we can add this to configuration file where we set MODULEPATH. This could be set in configuration file in same level as `find` with keyword `modulepath` that accepts a list of directories. 

```
# compiler block                                                                                                                                                                                                                  
compilers:         
  modulepath: ['/tmp', '/var/']                                                                                                                                                                                                               
  find:
    gcc: "^(gcc)"                                                                                                                                              
```

In order to make this happen we need to do the following:

- [x] Update [settings.schema.json](https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/settings.schema.json) with support for 'modulepath' keyword
- [x] Add new option `buildtest cg compilers --modulepath` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/__init__.py
- [x] Implement logic in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/compilers.py. Note when using `Lmod` one will call the Spider class from Lmodule. 

https://github.com/buildtesters/buildtest/blob/a7996718142cb30a2a6bccdf2dbed3ada87fdf52/buildtest/cli/compilers.py#L203-L206

The __init__ method of `Spider` class will accept a list of directories from MODULEPATH to get the modules

https://github.com/buildtesters/lmodule/blob/e0242505dae800789beeb81e57a9698db2a83188/lmod/spider.py#L19-L39

This is then retrieved by calling `get_modules` method

https://github.com/buildtesters/lmodule/blob/e0242505dae800789beeb81e57a9698db2a83188/lmod/spider.py#L68-L99

Which is invoked by buildtest

https://github.com/buildtesters/buildtest/blob/a7996718142cb30a2a6bccdf2dbed3ada87fdf52/buildtest/cli/compilers.py#L208

As per `environment-modules` logic we use `module av -t` to get list of modules that is processed, we should change it such that it will process the files belonging to MODULEPATH. One way is to get this detail is to build a list of files by iterating over the directories in MODULEPATH. The method `_validate_modules` will eventually test all the files anyways we just need to pass the list

https://github.com/buildtesters/buildtest/blob/a7996718142cb30a2a6bccdf2dbed3ada87fdf52/buildtest/cli/compilers.py#L260-L289

Xiangs18:


buildtest/schemas/settings.schema.json
```diff
@@ -94,6 +94,7 @@
           "description": "Declare compiler section for defining system compilers that can be referenced in buildspec.",
           "additionalProperties": false,
           "properties": {
+            "modulepath": [],
```
shahzebsiddiqui: you need to set this to a string type

buildtest/cli/__init__.py
```diff
@@ -832,6 +832,13 @@ def config_menu(subparsers):
         action="store_true",
         help="List compiler details in YAML format",
     )
+    compilers.add_argument(
+        "-m",
+        "--modulepath",
+        type=str,
+        nargs="+",
+        help="List modulepath to search for compiler directories",
```
shahzebsiddiqui: Say `Specify a list of directories to search for modules via MODULEPATH to detect compilers`

buildtest/cli/compilers.py
```diff
@@ -211,18 +215,24 @@ def find_compilers(self):
         module_dict = {}
         console.print(f"MODULEPATH: {os.getenv('MODULEPATH')}")
 
+        print("self.moduletool: ", self.moduletool)
```
shahzebsiddiqui: I think this print statement was for debugging purposes right? can you remove it.

buildtest/cli/compilers.py
```diff
@@ -211,18 +215,24 @@ def find_compilers(self):
...
             if self.detailed:
                 print("Searching modules via Lmod Spider")
-            spider = Spider()
+            spider = Spider(tree=self.modulepath)
+
+            print("spider.get_modules(): ", spider.get_modules())
 
             spider_modules = list(spider.get_modules().values())
-            for name, module_regex_patttern in self.configuration.target_config[
+            print("spider_modules: ", spider_modules)
```
shahzebsiddiqui: same with this statement please remove it once you have finished debugging

Xiangs18:
Hi @shahzebsiddiqui, this MR is ready for review!

buildtest/settings/config.yml
```diff
@@ -31,6 +31,7 @@ system:
           shell: zsh
     # compiler block
     compilers:
+      modulepath: ['/tmp', '/var/']
```
shahzebsiddiqui: please remove this from the default configuration. It is good for testing but should not be in the configuration. 

What you can do is put these two lines in comment and provide a brief description such as

buildtest/cli/__init__.py
```diff
@@ -855,6 +855,13 @@ def config_menu(subparsers):
         action="store_true",
         help="Update configuration file with new compilers",
     )
+    compiler_find.add_argument(
+        "-m",
+        "--modulepath",
+        type=str,
+        nargs="+",
+        help="Specify a list of directories to search for modules via MODULEPATH to detect compilers",
+    )
```
shahzebsiddiqui, line 864: can we move this option under `buildtest cg compilers` for now and if it doesn't work out we can move it within each subcommand

bash_completion.sh
```diff
@@ -177,7 +177,7 @@ _buildtest ()
           local opts="--help --json --yaml -h -j -y find test"
           COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
           if [[ "${prev}" == "find" ]]; then
-            local opts="--detailed --help --update -d -h -u"
+            local opts="--detailed --help --modulepath --update -d -h -m -u"
```
shahzebsiddiqui, line 180: that will also impact this line

buildtest/cli/compilers.py
```diff
@@ -90,15 +95,15 @@ def compiler_test(configuration):
         console.print(table)
 
 
-def compiler_find(args, configuration):
+def compiler_find(configuration, modulepath, detailed, update):
```
shahzebsiddiqui: Can you please make `modulepath`, `detailed` and `update` as optional argument and document them in docstring

buildtest/settings/config.yml
```diff
@@ -31,7 +31,8 @@ system:
           shell: zsh
     # compiler block
     compilers:
-      modulepath: ['/tmp', '/var/']
+      # a list of directories to search for modules
+      # modulepath: ['/tmp', '/var/']
```
shahzebsiddiqui, line 35: can you add a new line between `modulepath` and the next statement so its clear the comments are not next to command.

PR #1207, opened 2022-08-26
Issue 1190 console color option

### LINKED ISSUE #1190: [FEATURE]: buildtest report summary --color to specify color dynamically

Let's add an option to dynamically change the color via `buildtest report summary --color=<color>`.

### Post question in Slack

JeffreyFG:
Added final documentation for --color

### DIFF
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -177,20 +177,10 @@ _buildtest ()
         return
       esac
       case "$prev" in summary)
-        local opts="-d -h --detailed --help --pager"
+        local opts="-d -h --detailed --help"
         COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
         return
       esac
-      if [[ " ${arr1[*]} " == *" $prev "* ]]; then 
-            delete=$delete+(--color)
-            COMPREPLY=( $( compgen -W "${opts/$delete}" -- $cur ) ) 
-            return
-      fi
-      if [[ " ${arr2[*]} " == *" $prev "* ]]; then 
-            delete=$delete+($prev)
-            COMPREPLY=( $( compgen -W "${opts/$delete}" -- $cur ) ) 
-            return
-      fi
     ;;
 
     config|cg)
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -2,6 +2,7 @@
 # For auto completion via compgen, options are sorted alphabetically in the format: <longoption> <shortoption> <subcommands>
 # For more details see https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html
 
+
 if test -n "${ZSH_VERSION:-}" ; then
   if [[ "$(emulate)" = zsh ]] ; then
     if ! typeset -f compdef >& /dev/null ; then
@@ -17,6 +18,13 @@ if test -n "${ZSH_VERSION:-}" ; then
     return # stop interpreting file
   fi
 fi
+
+
+function containsElement () 
+{
+  
+  return $result
+}
 
 # get list of available tags
 _avail_tags ()
@@ -159,16 +167,27 @@ _buildtest ()
       ;;
 
     report|rt)
-      local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
-      local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;a=' --color-'.join(list(ANSI_COLOR_NAMES.keys())) ;print(a)")
-      local allopts="${opts} ${copts} "
-      COMPREPLY=( $( compgen -W "$allopts" -- $cur ) )
-
+      local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+      COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
+      case ${COMP_WORDS[2]} in --color)
+        local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;print(' '.join(list(ANSI_COLOR_NAMES.keys())))")
+        
+        COMPREPLY=( $( compgen -W "$copts" -- $cur ) )
+        arr2=($opts)
+        local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+        if [[ " ${arr1[*]} " == *" $prev "* ]]; then
+            local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+            COMPREPLY=( $( compgen -W "${opts}" -- $cur ) ) 
+        fi
+        if [[ "$prev" == "summary" ]]; then
+              local opts="-d -h --detailed --help --pager"
+              COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
+        fi
+      esac
       case "${COMP_WORDS[2]}" in summary)
         local opts="-d -h --detailed --help --pager"
-        COMPREPLY=( $( compgen -W "$opts" -- $cur ) )
-      esac
-    ;;
+        COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
+      esac;;
     config|cg)
       local cmds="-h --help compilers edit executors path systems validate view"
 
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -173,7 +173,7 @@ _buildtest ()
         local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;print(' '.join(list(ANSI_COLOR_NAMES.keys())))")
         
         COMPREPLY=( $( compgen -W "$copts" -- $cur ) )
-        arr2=($opts)
+        arr1=($copts)
         local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
         if [[ " ${arr1[*]} " == *" $prev "* ]]; then
             local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -168,26 +168,31 @@ _buildtest ()
 
     report|rt)
       local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+      local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;print(' '.join(list(ANSI_COLOR_NAMES.keys())))")
+      arr1=($copts)
+      arr2=($opts)
       COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-      case ${COMP_WORDS[2]} in --color)
-        local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;print(' '.join(list(ANSI_COLOR_NAMES.keys())))")
-        
+      case ${prev} in --color)
         COMPREPLY=( $( compgen -W "$copts" -- $cur ) )
-        arr1=($copts)
-        local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
-        if [[ " ${arr1[*]} " == *" $prev "* ]]; then
-            local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
-            COMPREPLY=( $( compgen -W "${opts}" -- $cur ) ) 
-        fi
-        if [[ "$prev" == "summary" ]]; then
-              local opts="-d -h --detailed --help --pager"
-              COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-        fi
+        return
       esac
-      case "${COMP_WORDS[2]}" in summary)
+      case "$prev" in summary)
         local opts="-d -h --detailed --help --pager"
         COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-      esac;;
+        return
+      esac
+      if [[ " ${arr1[*]} " == *" $prev "* ]]; then 
+            delete=$delete+(--color)
+            COMPREPLY=( $( compgen -W "${opts/$delete}" -- $cur ) ) 
+            return
+      fi
+      if [[ " ${arr2[*]} " == *" $prev "* ]]; then 
+            delete=$delete+($prev)
+            COMPREPLY=( $( compgen -W "${opts/$delete}" -- $cur ) ) 
+            return
+      fi
+    ;;
+
     config|cg)
       local cmds="-h --help compilers edit executors path systems validate view"
 
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -159,19 +159,16 @@ _buildtest ()
       ;;
 
     report|rt)
-      local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"
-      COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-      case "${COMP_WORDS[2]}" in --color)  
-        local opts=$(python -c "from rich.color import ANSI_COLOR_NAMES; print(str(list(ANSI_COLOR_NAMES.keys())).replace(',',''))")
-        COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-      esac
+      local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+      local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;a=' --color-'.join(list(ANSI_COLOR_NAMES.keys())) ;print(a)")
+      local allopts="${opts} ${copts} "
+      COMPREPLY=( $( compgen -W "$allopts" -- $cur ) )
+
       case "${COMP_WORDS[2]}" in summary)
         local opts="-d -h --detailed --help --pager"
-        COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
-        ;;
+        COMPREPLY=( $( compgen -W "$opts" -- $cur ) )
       esac
-      ;;
-      
+    ;;
     config|cg)
       local cmds="-h --help compilers edit executors path systems validate view"
 
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -18,13 +18,6 @@ if test -n "${ZSH_VERSION:-}" ; then
   fi
 fi
 
-
-function containsElement () 
-{
-  
-  return $result
-}
-
 # get list of available tags
 _avail_tags ()
 {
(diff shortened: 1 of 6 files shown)


bash_completion.sh
```diff
@@ -2,6 +2,7 @@
 # For auto completion via compgen, options are sorted alphabetically in the format: <longoption> <shortoption> <subcommands>
 # For more details see https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html
 
+
```
shahzebsiddiqui: remove this line

bash_completion.sh
```diff
@@ -159,14 +167,22 @@ _buildtest ()
       ;;
 
     report|rt)
-      local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"
-      COMPREPLY=( $( compgen -W "$opts" -- $cur ) )
-      case "${COMP_WORDS[2]}" in summary)
+      local opts="--color --end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"     
+      local copts=$(python -c "from rich.color import ANSI_COLOR_NAMES;print(' '.join(list(ANSI_COLOR_NAMES.keys())))")
+      arr1=($copts)
+      arr2=($opts)
```
shahzebsiddiqui: i dont see where `arr2` is being used can we remove this. I can see `$opts` is used later down the code

buildtest/cli/__init__.py
```diff
@@ -953,6 +953,9 @@ def report_menu(subparsers):
         action="store_true",
         help="Print output in machine readable format",
     )
+    parser_report.add_argument(
+        "--color", type=str, action="append", help="change the color of the output font"
```
shahzebsiddiqui: can we add a `choices` field see https://docs.python.org/3/library/argparse.html#choices and call a method that returns a list of colors. You can do something like `type=rich_colors` where `rich_colors` is a method that returns all the colors

buildtest/cli/report.py
```diff
@@ -552,6 +560,7 @@ def print_report(
             root_disk_usage|PASS|0
             root_disk_usage|PASS|0
         """
+        consoleColor = checkColor(color) or "blue"
```
shahzebsiddiqui: you can remove the `or "blue"`. The method `checkColor` will return the default anyways

buildtest/cli/report.py
```diff
@@ -874,29 +886,38 @@ def print_report_summary_output(
...
+    consoleColor = checkColor(color)
+
+    console.print("Report File: ", report.reportfile(), style=consoleColor)
+    console.print("Total Tests:", len(report.get_testids()), style=consoleColor)
+    console.print("Total Tests by Names: ", len(report.get_names()), style=consoleColor)
+    console.print(
+        "Number of buildspecs in report: ",
+        len(report.get_buildspecs()),
+        style=consoleColor,
+    )
```
shahzebsiddiqui: as discussed can we not change the color of all text to the consoleColor

docs/gettingstarted/query_test_report.rst
```diff
@@ -268,6 +268,14 @@ where ``--count`` is number of rows that that get printed. Shown below is the ou
 The ``--count`` option also works with terse mode ``--terse``, shown below is the same output 
 
 .. command-output:: buildtest report --terse --count 2
+Color Output
+-------------
+
+If you would like to change the output color for either ``buildtest report`` and ``buildtest report summary`` you can use the ``--color`` option.
+This will print the output of the previous two commands in the provied color. The color must be one of the supported colors by rich console.
```
shahzebsiddiqui: spelling error `provied` -> `provided`

bash_completion.sh
JeffreyFG: done

bash_completion.sh
JeffreyFG: Removed both arr1 and arr2

buildtest/cli/report.py
JeffreyFG: removed

buildtest/cli/report.py
JeffreyFG: Can you please be more specific about which lines of output do you not want to be in the console color

bash_completion.sh
```diff
@@ -18,6 +18,13 @@ if test -n "${ZSH_VERSION:-}" ; then
   fi
 fi
 
+
+function containsElement ()
```
shahzebsiddiqui: what is the point of this method i dont see it being called

bash_completion.sh
```diff
@@ -159,14 +166,20 @@ _buildtest ()
       ;;
 
     report|rt)
-      local opts="--end --failure --filter --format --help --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"
-      COMPREPLY=( $( compgen -W "$opts" -- $cur ) )
-      case "${COMP_WORDS[2]}" in summary)
+      local opts="--color --end --failure --filter --format --help --helpcolor --helpfilter --helpformat --latest --no-header --oldest --pager --passed --start --terse  -e -f -h -n -p -s -t clear list summary"
```
shahzebsiddiqui: can you remove `--helpcolor` from this bash completion this feature is not implemented yet. Please add this in your subsequent PR

buildtest/cli/__init__.py
```diff
@@ -82,6 +83,35 @@ def positive_number(value):
     return int_val
 
 
+def supported_color(value):
+    """Checks if input is positive number and returns value as an int type.
```
shahzebsiddiqui: please change this docstring you must have copied this from another method

buildtest/cli/__init__.py
```diff
@@ -82,6 +83,35 @@ def positive_number(value):
...
+
+    Args:
+        value (str): Specify an input number
+
+    Returns:
+        str: Return value as str type
+
```
shahzebsiddiqui: again same thing please update this docstring accordingly

buildtest/cli/__init__.py
```diff
@@ -82,6 +83,35 @@ def positive_number(value):
     return int_val
 
 
+def supported_color(value):
```
shahzebsiddiqui: you might want to use a meaningful argument instead of `value` perhaps say `input_color` or `color_choice`

buildtest/cli/__init__.py
```diff
@@ -82,6 +83,35 @@ def positive_number(value):
...
+    if not isinstance(value, (str)):
+        raise argparse.ArgumentTypeError(
+            f"Input must be a string type, you have specified '{value}' which is of type {type(value)}"
+        )
+    try:
+        color_val = Color.parse(value)
+    except ColorParseError:
+        console.print(f"[red]Unable to convert {value} to a Color ")
+        console.print_exception()
+        return None
```
shahzebsiddiqui: note that `return None` is same as just doing a `return`.

I think if the exception is raised this is the correct behavior from this method. I was contemplating on using a default color if there is an error but i think its probably better we raise an exception on invalid color type instead of resorting to a default color.

buildtest/cli/__init__.py
```diff
@@ -953,6 +983,13 @@ def report_menu(subparsers):
         action="store_true",
         help="Print output in machine readable format",
     )
+    parser_report.add_argument(
+        "--color",
+        type=supported_color,
+        metavar="COLOR",
+        action="append",
```
shahzebsiddiqui: i think the `--color` option should not be appended multiple times meaning you can't do `buildtest rt --color blue --color red`. That being said please remove `action="append"`. The `type=supported_color` will handle this correctly which will be of type `str` that is a valid color

buildtest/cli/__init__.py
```diff
@@ -953,6 +983,13 @@ def report_menu(subparsers):
...
+        help="Prints reports and the subcommand summary in the provided COLOR, colors defined in Rich.Color",
```
shahzebsiddiqui: You can say `help="Print output of report table with the selected color.`

That way we are not too specific on the help message since implementation may change in future. We can elaborate on this in the user docs instead of the help message.

buildtest/cli/report.py
```diff
@@ -7,12 +7,29 @@
...
-from rich.color import Color, ColorParseError
+from rich.color import ANSI_COLOR_NAMES, Color, ColorParseError
 from rich.table import Table
 
 logger = logging.getLogger(__name__)
 
 
+def checkColor(colorArg):
+    """Checks the provided colorArg against the compatible colors from Rich.Color"""
+    if type(colorArg) is Color:
```
shahzebsiddiqui: i think you can say `if isinstance(colorArg, Color)` that would imply the object is an instance of type `Color`. See https://docs.python.org/3/library/functions.html#isinstance.

buildtest/cli/report.py
```diff
@@ -7,12 +7,29 @@
...
+        return colorArg.name
+
+    if colorArg and type(colorArg) is list:
```
shahzebsiddiqui: you can change this to `if isinstance(colorArg, list):` which is essentially the same statement

buildtest/cli/report.py
```diff
@@ -7,12 +7,29 @@
...
+        colorArg = colorArg[0]
+        return colorArg
+    if type(colorArg) is str:
```
shahzebsiddiqui: same here change this to `if isinstance(colorArg, str):`

buildtest/cli/report.py
```diff
@@ -7,12 +7,29 @@
...
+        try:
+            checkedColor = Color.parse(colorArg).name
+        except ColorParseError:
+            checkedColor = Color.default().name
+        return checkedColor
+    return None
```
shahzebsiddiqui: you dont need this `return None` statement. Python will implicitly return a `None` when it gets out of a method

buildtest/cli/report.py
```diff
@@ -59,6 +76,7 @@ class Report:
         "state": {"description": "Filter by test state", "type": "PASS/FAIL"},
         "tags": {"description": "Filter tests by tag name", "type": "STRING"},
     }
+    color_options = list(ANSI_COLOR_NAMES.keys())
```
shahzebsiddiqui: why is this line in there? this must be a typo

buildtest/cli/report.py
```diff
@@ -479,12 +500,23 @@ def process_report(self):
...
 
         console.print(table)
 
+    def print_color_help(self):
+        """Displays list of colors supported by the rich console which implements command ``buildtest report --helpcolor``"""
+        table = Table("Color", title="Supported Colors")
+        for color_option in self.color_options:
+            table.add_row(f"[{color_option}]{color_option}")
+        console.print(table)
+
```
shahzebsiddiqui: Can you please remove this statement. This should be in a subsequent PR.

buildtest/cli/report.py
```diff
@@ -576,14 +609,11 @@ def print_report(
         join_list = []
         title = title or f"Report File: {self.reportfile()}"
         table = Table(title=title, show_lines=True, expand=True)
-        consoleColor = checkColor(color=color)
-
         for field in self.display_table.keys():
             table.add_column(
-                f"[blue]{field}", overflow="fold", style=consoleColor
-            )  # change [blue] to console color if we want the entire console output to be uniform.
+                f"[{consoleColor}]{field}", overflow="fold", style=consoleColor
```
shahzebsiddiqui: now that i think about this i think the `style=consoleColor` will take care of coloring the field. See https://rich.readthedocs.io/en/stable/reference/table.html#rich.table.Table.add_column.

That being said i think the `f"[{consoleColor}]{field}"` could be changed to `table.add_column("field",overflow="fold", style=consoleColor)`.

Please try this out and if it works then please make the change otherwise leave it as is.

buildtest/cli/report.py
```diff
@@ -874,29 +910,22 @@ def print_report_summary_output(
...
+    consoleColor = checkColor(color)
+
+    console.print("Report File: ", report.reportfile(), style=consoleColor)
+    console.print("Total Tests:", len(report.get_testids()), style=consoleColor)
+    console.print("Total Tests by Names: ", len(report.get_names()), style=consoleColor)
+    console.print(
+        "Number of buildspecs in report: ",
+        len(report.get_buildspecs()),
+        style=consoleColor,
+    )
```
shahzebsiddiqui: for the time being let's not add `style=consoleColor` to each of these statements let it print out in the normal output whatever color that may be. However we do want to color the output of the tables for pass and fail tables so that looks right.

docs/gettingstarted/query_test_report.rst
```diff
@@ -20,6 +20,10 @@ with default format fields. To see a list of all format fields, click :ref:`here
 .. note::
    ``buildtest rt`` is an alias for ``buildtest report`` command.
 
+Color Options (``buildtest report --color``)
```
shahzebsiddiqui: Perhaps you can name this title `Color output of buildtest report (``buildtest report --color``)

docs/gettingstarted/query_test_report.rst
```diff
@@ -268,6 +272,14 @@ where ``--count`` is number of rows that that get printed. Shown below is the ou
...
 
 .. command-output:: buildtest report --terse --count 2
+Color Output
+-------------
+
+If you would like to change the output color for either ``buildtest report`` and ``buildtest report summary`` you can use the ``--color`` option.
+This will print the output of the previous two commands in the provided color. The color must be one of the supported colors by rich console.
+
+.. command-output:: buildtest report --color red 
+.. command-output:: buildtest report --color red summary
```
shahzebsiddiqui: Don't we want this section under the section you had previously defined `Color Options (``buildtest report --color``)` on line 23. Please move this up.

bash_completion.sh
JeffreyFG: removed

buildtest/cli/__init__.py
JeffreyFG: done

buildtest/cli/__init__.py
JeffreyFG: simplified

buildtest/cli/__init__.py
JeffreyFG: Updated input and return sections

buildtest/cli/__init__.py
JeffreyFG: Removed

buildtest/cli/__init__.py
JeffreyFG: simplified

buildtest/cli/__init__.py
JeffreyFG: changed

buildtest/cli/report.py
JeffreyFG: removed

buildtest/cli/report.py
JeffreyFG: removed

buildtest/cli/report.py
JeffreyFG: simplified

buildtest/cli/report.py
JeffreyFG: simplified

buildtest/cli/report.py
JeffreyFG: simplified

buildtest/cli/report.py
JeffreyFG: simplified

buildtest/cli/report.py
JeffreyFG: console color removed

buildtest/cli/report.py
JeffreyFG: So the only difference between those two is that with the original line(f"[{consoleColor}]{field}") the feild title text is also changed to the color. With "table.add_column("field",overflow="fold", style=consoleColor)" the feild name is in the default color tell me if thats the way you want it.

buildtest/cli/report.py
JeffreyFG: Ok so I removed the color from this section

docs/gettingstarted/query_test_report.rst
JeffreyFG: moved

docs/gettingstarted/query_test_report.rst
JeffreyFG: changed

buildtest/cli/report.py
shahzebsiddiqui: Yeah let's only change the rowa that being said I think the original code is making everything be blue so let's keep this as is. I think using `style=consoleColor`  means we don't need to color inside the fitting.

Just make sure the column heading are blue and only the table rows are changed based on --color option

shahzebsiddiqui: You can do table.add_column(field, ..., style=consoleColor)

JeffreyFG: done

buildtest/cli/__init__.py
```diff
@@ -83,14 +84,14 @@ def positive_number(value):
...
 
 
-def supported_color(value):
-    """Checks if input is positive number and returns value as an int type.
+def supported_color(input_color):
+    """Checks if input is a supported color and returns value as an Color type.
 
     Args:
-        value (str): Specify an input number
+        value (str): Specify an input color
```
shahzebsiddiqui: please change `value` to `input_color` these args in docstring need to match the method signature.

buildtest/cli/report.py
```diff
@@ -7,27 +7,27 @@
...
         return colorArg
-    if type(colorArg) is str:
+    if isinstance(colorArg, str):
         try:
             checkedColor = Color.parse(colorArg).name
         except ColorParseError:
             checkedColor = Color.default().name
         return checkedColor
-    return None
+    return
```
shahzebsiddiqui: you dont need this `return` statement since this is the last line in method and python will return from method. Only use `return` when needed specially when one needs to return out of method using an if condition.

docs/gettingstarted/query_test_report.rst
```diff
@@ -20,9 +20,14 @@ with default format fields. To see a list of all format fields, click :ref:`here
 .. note::
    ``buildtest rt`` is an alias for ``buildtest report`` command.
 
-Color Options (``buildtest report --color``)
+Color output of buildtest report (buildtest report --color)
 ---------------------------------------------
```
shahzebsiddiqui: you need the `----` to go all the way to length of the heading i think if you dont do that their is some rendering issues. Take a look at https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings

docs/gettingstarted/query_test_report.rst
```diff
@@ -20,9 +20,14 @@ with default format fields. To see a list of all format fields, click :ref:`here
...
 
+If you would like to change the output color for either ``buildtest report`` and ``buildtest report summary`` you can use the ``--color`` option.
+This will print the output of the previous two commands in the provided color. The color must be one of the supported colors by rich console.
+
+.. command-output:: buildtest report --color red 
+.. command-output:: buildtest report --color red summary
```
shahzebsiddiqui: lets not run these commands because readthedocs wouldnt render output of color in docs see https://buildtest--1207.org.readthedocs.build/en/1207/gettingstarted/query_test_report.html#color-output which is the build of this docs for this PR.

Instead let's put this as a block of code that is highlighted using the `::` with tab indentation so it would be something like this


```
The color must be one of the supported colors by rich console, show below is an example of coloring output of tables with red::

    $ buildtest report --color red
    $ buildtest report --color red summary --detailed
```

We need to use `--detailed` otherwise the output doesnt color since default output of `buildtest rt summary` is just text and not table.

buildtest/cli/report.py
JeffreyFG: removed

buildtest/cli/__init__.py
JeffreyFG: changed

docs/gettingstarted/query_test_report.rst
JeffreyFG: I think that this is the way that yout want it let me know

docs/gettingstarted/query_test_report.rst
JeffreyFG: -- added

docs/gettingstarted/query_test_report.rst
shahzebsiddiqui: No if you use `.. command-output::` it will run the command and that will not work correctly. Just add this as a blob of text tab indented as i shared in https://github.com/buildtesters/buildtest/pull/1207#discussion_r966112535

docs/gettingstarted/query_test_report.rst
```diff
@@ -268,6 +278,14 @@ where ``--count`` is number of rows that that get printed. Shown below is the ou
...
 
 .. command-output:: buildtest report --terse --count 2
+Color Output
+-------------
+
+If you would like to change the output color for either ``buildtest report`` and ``buildtest report summary`` you can use the ``--color`` option.
+This will print the output of the previous two commands in the provided color. The color must be one of the supported colors by rich console.
+
+.. command-output:: buildtest report --color red 
+.. command-output:: buildtest report --color red summary
```
shahzebsiddiqui: wait i think this section is duplicated with what i see in section on line 23. Can you please remove this section `Color Output` the one that uses `command-output` this should not be in docs 

```
Color output of buildtest report (buildtest report --color)
------------------------------------------------------------
```

PR #1239, opened 2022-09-15
Add positional argument to test specific compilers

### LINKED ISSUE #1231: [FEATURE]: add positional argument to 'buildtest config compilers test' to test specific compilers

@Xiangs18 

Let's add support to test specific compilers given input argument so `buildtest config compilers test <compiler1> <compiler2>`

Right now we test all compilers and this can take a long time. We can keep the positional argument as optional such that one can do `buildtest config compilers test` that will test all compilers but if an argument is specified then the specific compiler is tested.

Let's assume we have one compiler

```
(buildtest)  ~/Documents/github/buildtest/ [remove_timeline] buildtest cg compilers        
builtin_gcc
``` 

We can test this via `buildtest config compilers test builtin_gcc` the positional argument should be tab completed. If input argument is invalid test then we skip the test so one can do `buildtest cg compilers test gcc_11.2.0 gcc_10.2.0` to test multiple compilers

### Post question in Slack

Xiangs18:


buildtest/cli/__init__.py
```diff
@@ -925,10 +925,14 @@ def config_menu(subparsers):
         help="Specify a list of directories to search for modules via MODULEPATH to detect compilers",
     )
 
-    subparsers_compiler.add_parser(
+    compiler_test = subparsers_compiler.add_parser(
         "test",
         help="Test each compiler instance by performing module load test",
     )
+    compiler_test.add_argument(
+        "specific_compilers",
```
shahzebsiddiqui: just a nitpick, you can call these `compiler_names`

buildtest/cli/__init__.py
```diff
@@ -925,10 +925,14 @@ def config_menu(subparsers):
         help="Specify a list of directories to search for modules via MODULEPATH to detect compilers",
     )
 
-    subparsers_compiler.add_parser(
+    compiler_test = subparsers_compiler.add_parser(
         "test",
         help="Test each compiler instance by performing module load test",
     )
+    compiler_test.add_argument(
```
shahzebsiddiqui, line 958: can you add a help message something like `help="Specify a compiler name to test by running module test"`

buildtest/cli/compilers.py
```diff
@@ -43,7 +43,7 @@ def compiler_cmd(args, configuration):
         bc.print_yaml()
 
 
-def compiler_test(configuration):
+def compiler_test(configuration, specific_compilers=[]):
```
shahzebsiddiqui: please add a docstring describing the argument, can we change this to `compiler_names`.

shahzebsiddiqui: I think this empty list doesn't make sense, the default value of `args.specific_compilers` would be None if its not specified. I think we can make this `specific_compilers=None` so its an optional argument.

buildtest/cli/compilers.py
```diff
@@ -56,9 +56,12 @@ def compiler_test(configuration):
     fail_compilers = []
 
     compilers = configuration.target_config["compilers"]["compiler"]
+    target_compilers = set(specific_compilers)
 
     for name in compilers:
         for module in compilers[name]:
+            if target_compilers and module not in target_compilers:
+                continue
```
shahzebsiddiqui, line 68: it could be helpful to have a `console.print(f"Skipping test for compiler {target_compilers}")`  message

shahzebsiddiqui:
You can try this on Ascent system on OLCF see https://github.com/buildtesters/buildtest/blob/devel/tests/test_ascent.py. The configuration is in https://github.com/buildtesters/buildtest/blob/devel/tests/settings/ascent.yml and i have access to the system with Lmod so i just need to run `buildtest cg co find` and update this configuration so we can do something like `buildtest cg co test gcc/7.5.0`

```
[shahzebsiddiqui@login1.ascent ~]$ ml av gcc

---------------------------------------------------------------------------------------------------------------- /sw/ascent/spack-envs/base/modules/site/Core ----------------------------------------------------------------------------------------------------------------
   gcc/7.5.0    gcc/9.1.0 (D)    gcc/9.3.0    gcc/10.2.0    gcc/11.1.0    gcc/11.2.0

  Where:
   D:  Default Module

Use "module spider" to find all possible modules and extensions.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".

```

buildtest/cli/__init__.py
shahzebsiddiqui: fixed

buildtest/cli/compilers.py
shahzebsiddiqui: fixed

PR #1260, opened 2022-10-05
Add subcommand "buildtest report path" to print path to the report file being used.

### LINKED ISSUE #1256: [FEATURE]: show path to report file via 'buildtest report path'

we should add a new subcommand `buildtest report path` that will print the full path to report file that is being used. 

We can see list of report files via 
```
(buildtest)  ~/Documents/github/buildtest/ [buildtest_buildspec_find_count] buildtest rt list
/Users/siddiq90/Documents/github/buildtest/var/report.json

```

Note that one can clear the report file hence there wont be any report so running any `buildtest rt` command or subcommand should just raise an exception

```
(buildtest)  ~/Documents/github/buildtest/ [buildtest_buildspec_find_count] buildtest rt clear
Removing report file: /Users/siddiq90/Documents/github/buildtest/var/report.json
(buildtest)  ~/Documents/github/buildtest/ [buildtest_buildspec_find_count] buildtest rt      
Unable to find report please check if /Users/siddiq90/Documents/github/buildtest/var/report.json is a file or run a test via 'buildtest build' to generate report file
```

### Post question in Slack

prathmesh4321:
**buildtest rt path**

buildtest_report

**buildtest -r <path> rt path**

builtest_report_try

**"buildtest rt path" raising exception when there is no report file**

buildtes report clear

### DIFF
--- a/buildtest/cli/__init__.py
+++ b/buildtest/cli/__init__.py
@@ -4,7 +4,6 @@ interact with a global configuration for buildtest.
 """
 import argparse
 import datetime
-from ast import alias
 
 from buildtest import BUILDTEST_COPYRIGHT, BUILDTEST_VERSION
 from buildtest.defaults import console
--- a/buildtest/cli/__init__.py
+++ b/buildtest/cli/__init__.py
@@ -4,6 +4,7 @@ interact with a global configuration for buildtest.
 """
 import argparse
 import datetime
+from ast import alias
 
 from buildtest import BUILDTEST_COPYRIGHT, BUILDTEST_VERSION
 from buildtest.defaults import console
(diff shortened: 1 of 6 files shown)


buildtest/cli/__init__.py
```diff
@@ -4,6 +4,7 @@
 """
 import argparse
 import datetime
+from ast import alias
```
shahzebsiddiqui: i dont think we need this import statement. It not being used must be an accident on your end. Note we already have builtin `aliases` as a argument to `add_parser` which is available from [argparse](https://docs.python.org/3/library/argparse.html) library

shahzebsiddiqui (changes requested):
This look perfect great job, can you please remove the import statement and we should be all set to merge

buildtest/cli/__init__.py
prathmesh4321: Done ! @shahzebsiddiqui

shahzebsiddiqui (approved):

PR #1277, opened 2022-10-14
Add option buildtest --helpcolor to show available color options in table format

### LINKED ISSUE #1206: [FEATURE]: Add buildtest -helpcolor flag to

The amount of different colors can be confusing and it would before great if a use could have ``buildtest --helpcolor`` to find out which color options there are. 

### Post question in Slack

prathmesh4321:


buildtest/cli/helpcolor.py
```diff
@@ -0,0 +1,15 @@
+from buildtest.defaults import console
+from rich.color import ANSI_COLOR_NAMES
+from rich.table import Table
+
+
+def print_available_colors():
+    """Print the available color options in a table format on background of the color option."""
+    table = Table(
+        "[blue]Number",
+        "[blue]Color",
```
shahzebsiddiqui: you can replace the `[blue]` with argument `header_style='blue'`. See https://rich.readthedocs.io/en/stable/reference/table.html#rich.table.Table. I believe this will color the headers based on the color. Try it out i believe it should work

shahzebsiddiqui (changes requested):
One last thing can you please add a section for documentation. Perhaps you can mention this in https://buildtest.readthedocs.io/en/devel/features.html

buildtest/cli/helpcolor.py
prathmesh4321: Done !

prathmesh4321:
> One last thing can you please add a section for documentation. Perhaps you can mention this in https://buildtest.readthedocs.io/en/devel/features.html

Done ! @shahzebsiddiqui

shahzebsiddiqui:
@prathmesh4321 question about the image you uploaded about the `buildtest --helpcolor` if we want to add the picture we should probably add the command prompt `$ buildtest --helpcolor` or we dont include the image or run the command. If we include the image then it becomes one extra step to update the image if the command option changes or the output behavior.

I was looking at the generated docs https://buildtest--1277.org.readthedocs.build/en/1277/features.html#listing-available-color-options-buildtest-helpcolor and it looks nice but i am thinking we shouldnt have this output. Let's follow up on slack and would like to get your opinion on the docs

prathmesh4321:
> @prathmesh4321 question about the image you uploaded about the `buildtest --helpcolor` if we want to add the picture we should probably add the command prompt `$ buildtest --helpcolor` or we dont include the image or run the command. If we include the image then it becomes one extra step to update the image if the command option changes or the output behavior.
> 
> I was looking at the generated docs https://buildtest--1277.org.readthedocs.build/en/1277/features.html#listing-available-color-options-buildtest-helpcolor and it looks nice but i am thinking we shouldnt have this output. Let's follow up on slack and would like to get your opinion on the docs

@shahzebsiddiqui I have added the command instead of the picture. Let me know if that looks right.

shahzebsiddiqui (approved):
LGTM. I rephrased one sentence but this looks good.

PR #1295, opened 2022-11-14
Color output for filter and format fileds in buildspec cache

### LINKED ISSUE #1255: color output for filter and format fields in buildspec cache
we should be able to use `--color` option for the following option

```
  --helpfilter         Show Filter fields for --filter option for filtering
                       buildspec cache output
  --helpformat         Show Format fields for --format option for formatting
                       buildspec cache output
  --filterfields       Print raw Filter fields for --filter option for
                       filtering builspec cache output
  --formatfields       Print raw Format fields for --format option for
                       formatting builspec cache output

```

prathmesh4321:
Screen Shot 2022-11-14 at 2 54 57 PM

### DIFF
--- a/buildtest/cli/buildspec.py
+++ b/buildtest/cli/buildspec.py
@@ -930,8 +930,7 @@ class BuildspecCache:
             pprint(value)
         sys.exit(1)
 
-    @staticmethod
-    def print_filter_fields():
+    def print_filter_fields(self):
         """This method prints filter fields available for buildspec cache. This
         method implements command ``buildtest buildspec find --helpfilter``
         """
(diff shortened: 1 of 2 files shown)


buildtest/cli/buildspec.py
```diff
@@ -973,14 +971,14 @@ def print_raw_filter_fields(self):
         method implements command ``buildtest buildspec find --filterfields``
         """
         for field in self.filter_fields:
-            console.print(field)
+            console.print(f"[{self.color}]{field}")
```
shahzebsiddiqui: i think this statement you can do `console.print(field,style=self.color)`.

The only time you may want to wrap this around fstring is when you need to color part of a word like `console.print("[red]red text[/red] and normal text")`. Sometimes data can't be represented with `console.print(data)` where data is not printed properly and therefore one needs to do `console.print(f"{data}")`. This is true for some data-types like class instances

buildtest/cli/buildspec.py
```diff
@@ -973,14 +971,14 @@ def print_raw_filter_fields(self):
...
 
     def print_raw_format_fields(self):
         """This method prints the raw format fields available for buildspec cache. This
         method implements command ``buildtest buildspec find --formatfields``
         """
         for field in self.format_fields:
-            console.print(field)
+            console.print(f"[{self.color}]{field}")
```
shahzebsiddiqui: same here i think you can do `console.print(field,style=self.color)` since each field is a str type

buildtest/cli/buildspec.py
prathmesh4321: @shahzebsiddiqui Okay got it, fixed this !

shahzebsiddiqui:
Lgtm

shahzebsiddiqui (approved):

PR #1299, opened 2022-11-17
Add coverage for buildtest report clear

### LINKED ISSUE #1281: add regression coverage for buildtest report clear
see https://app.codecov.io/gh/buildtesters/buildtest/blob/devel/buildtest/cli/report.py

Screen Shot 2022-10-20 at 3 57 27 PM

I suppose since we need the report files we should consider creating backups of file during test just so we can remove it and then copy the files back since other test may rely on the report and having it removed can cause some tests to fail.

prathmesh4321:
Screen Shot 2022-11-17 at 2 25 04 PM

### DIFF
--- a/tests/cli/test_report.py
+++ b/tests/cli/test_report.py
@@ -302,6 +302,7 @@ def test_report_clear():
     with pytest.raises(SystemExit):
         report_cmd(args)
 
+    assert not is_file(BUILD_REPORT)
     # move back the backe-up files since report_cmd() function removes the files BUILD_REPORT and BUILDTEST_REPORTS
     shutil.move(backupfile_report.name, BUILD_REPORT)
 
--- a/tests/cli/test_report.py
+++ b/tests/cli/test_report.py
@@ -7,6 +7,7 @@ import pytest
 from buildtest.cli.report import Report, report_cmd, report_summary
 from buildtest.defaults import BUILD_REPORT, BUILDTEST_REPORTS, BUILDTEST_ROOT
 from buildtest.exceptions import BuildTestError
+from buildtest.utils.file import is_file
 from rich.color import Color
 
 
@@ -261,8 +262,8 @@ def test_report_list():
 
     report_cmd(args)
 
-    backupfile = BUILDTEST_REPORTS + ".bak"
-    shutil.copy2(BUILDTEST_REPORTS, backupfile)
+    backupfile = tempfile.NamedTemporaryFile()
+    shutil.copy2(BUILDTEST_REPORTS, backupfile.name)
 
     # now removing report summary it should print a message
     os.remove(BUILDTEST_REPORTS)
@@ -270,8 +271,9 @@ def test_report_list():
     with pytest.raises(SystemExit):
         report_cmd(args)
 
-    shutil.move(backupfile, BUILDTEST_REPORTS)
-    assert BUILDTEST_REPORTS
+    # move back the removed BUILDTEST_REPORTS file
+    shutil.move(backupfile.name, BUILDTEST_REPORTS)
+    assert is_file(BUILDTEST_REPORTS)
 
 
 @pytest.mark.cli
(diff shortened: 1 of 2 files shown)


tests/cli/test_report.py
```diff
@@ -261,12 +261,18 @@ class args:
...
+    shutil.copy2(BUILDTEST_REPORTS, backupfile)
+
     # now removing report summary it should print a message
     os.remove(BUILDTEST_REPORTS)
 
     with pytest.raises(SystemExit):
         report_cmd(args)
 
+    shutil.move(backupfile, BUILDTEST_REPORTS)
+    assert BUILDTEST_REPORTS
```
shahzebsiddiqui: these assert statements dont do anything. I would either do `is_file(BUILDTEST_REPORTS)` which will check if file exists.

tests/cli/test_report.py
```diff
@@ -282,14 +288,22 @@ class args:
...
 
     # buildtest report clear will raise an error since file doesn't exist
     with pytest.raises(SystemExit):
         report_cmd(args)
 
-    shutil.move(backupfile, BUILD_REPORT)
+    shutil.move(backupfile_report, BUILD_REPORT)
+
+    shutil.move(backupfile_list_report, BUILDTEST_REPORTS)
+
```
shahzebsiddiqui: Before running these commands i would check if the files were removed by the `report_cmd` method since it will remove the files so having a check like

`assert not is_file(BUILD_REPORT)` would be useful

tests/cli/test_report.py
```diff
@@ -282,14 +288,22 @@ class args:
...
+    shutil.copy2(BUILDTEST_REPORTS, backupfile_list_report)
+
+    report_cmd(args)
 
     # buildtest report clear will raise an error since file doesn't exist
     with pytest.raises(SystemExit):
         report_cmd(args)
 
-    shutil.move(backupfile, BUILD_REPORT)
+    shutil.move(backupfile_report, BUILD_REPORT)
```
shahzebsiddiqui: it would be helpful to put some comments here on why we moving back files.

tests/cli/test_report.py
```diff
@@ -282,14 +288,22 @@ class args:
         no_header = None
         color = None
 
-    backupfile = BUILD_REPORT + ".bak"
-    shutil.copy2(BUILD_REPORT, backupfile)
+    backupfile_report = BUILD_REPORT + ".bak"
```
shahzebsiddiqui: please use tempfile https://docs.python.org/3/library/tempfile.html to create temporary file and then just copy the content into the tempfile

shahzebsiddiqui: or take a look at https://stackoverflow.com/questions/6587516/how-to-concisely-create-a-temporary-file-that-is-a-copy-of-another-file-in-pytho

tests/cli/test_report.py
prathmesh4321: @shahzebsiddiqui I have added this. Could you please check and let me know if that is the way you meant it should be added ?

tests/cli/test_report.py
```diff
@@ -282,15 +290,25 @@ class args:
...
 
-    shutil.move(backupfile, BUILD_REPORT)
-    assert BUILD_REPORT
+    assert not is_file(BUILD_REPORT)
+    # move back the backe-up files since report_cmd() function removes the files BUILD_REPORT and BUILDTEST_REPORTS
+    shutil.move(backupfile_report.name, BUILD_REPORT)
+
+    shutil.move(backupfile_list_report.name, BUILDTEST_REPORTS)
+
+    assert is_file(BUILD_REPORT)
```
shahzebsiddiqui, line 311: could you also add `assert is_file(BUILDTEST_REPORTS)` just to be consistent since we are moving both files back

shahzebsiddiqui: Once we have this i think this looks good and we can merge

prathmesh4321: @shahzebsiddiqui Done! You can merge this now.

shahzebsiddiqui (approved):
LGTM

PR #1339, opened 2023-01-12
Assert less than and equal status check

### LINKED ISSUE #1314: Status Check for Assert Less Than Equal - assert_le
In https://github.com/buildtesters/buildtest/pull/1313 we added support for assert greater equal `assert_ge` for status check. Let's add an operation `assert_le` that follows the same format. We can try the same test.

```
buildspecs:
  stream_test:
    type: script
    executor: generic.local.bash
    description: Run stream test with metrics example using assert greater equal
    env:
      OMP_NUM_THREADS: 4
    run: |
      wget https://raw.githubusercontent.com/jeffhammond/STREAM/master/stream.c
      gcc -openmp -o stream stream.c
      ./stream
    metrics:
      copy:
        type: float
        regex:
          exp: 'Copy:\s+(\S+)\s+.*'
          stream: stdout
          item: 1
      scale:
        type: float
        regex:
          exp: 'Scale:\s+(\S+)\s+.*'
          stream: stdout
          item: 1
      add:
        type: float
        regex:
          exp: 'Add:\s+(\S+)\s+.*'
          stream: stdout
          item: 1
      triad:
        type: float
        regex:
          exp: 'Triad:\s+(\S+)\s+.*'
          stream: stdout
          item: 1
    status:
      assert_le:
        - name: copy
          ref: 5000
        - name: scale
          ref: 5500
        - name: add
          ref: 6000
        - name: triad
          ref: 6500
```

Tasks:
- [x] Add support for key `assert_le` in  [definitions.schema.json](https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/definitions.schema.json)
- [x] Add logic for `assert_le` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/builders/base.py
- [x] Add regression test in https://github.com/buildtesters/buildtest/blob/devel/tests/builders/test_builders.py with an example buildspec in same directory
- [x] Add documentation

Xiangs18:
Hi @shahzebsiddiqui, please take a look at MR and provide feedback. Thank you!

Screen Shot 2023-01-11 at 6 20 44 PM

### DIFF
--- a/buildtest/buildsystem/checks.py
+++ b/buildtest/buildsystem/checks.py
@@ -260,7 +260,7 @@ def assert_ge_check(builder):
 
     Args:
         builder (buildtest.builders.base.BuilderBase): An instance of BuilderBase class used for printing the builder name
-        
+
     Returns:
         bool: True or False for performance check ``assert_ge``
     """
--- a/buildtest/buildsystem/checks.py
+++ b/buildtest/buildsystem/checks.py
@@ -258,6 +258,9 @@ def assert_ge_check(builder):
     One can specify multiple assert checks to check each metric with its reference value. When multiple items are specified, the operation is a logical AND and all checks
     must be ``True``.
 
+    Args:
+        builder (buildtest.builders.base.BuilderBase): An instance of BuilderBase class used for printing the builder name
+        
     Returns:
         bool: True or False for performance check ``assert_ge``
     """
@@ -320,6 +323,9 @@ def assert_le_check(builder):
     """Perform check on assert less than and equal when ``assert_le`` is specified in buildspec. The return is a boolean value that determines if the check has passed.
     One can specify multiple assert checks to check each metric with its reference value. When multiple items are specified, the operation is a logical AND and all checks
     must be ``True``.
+
+    Args:
+        builder (buildtest.builders.base.BuilderBase): An instance of BuilderBase class used for printing the builder name
 
     Returns:
         bool: True or False for performance check ``assert_le``
--- a/buildtest/buildsystem/checks.py
+++ b/buildtest/buildsystem/checks.py
@@ -316,6 +316,69 @@ def assert_ge_check(builder):
     return all(assert_check)
 
 
+def assert_le_check(builder):
+    """Perform check on assert less than and equal when ``assert_le`` is specified in buildspec. The return is a boolean value that determines if the check has passed.
+    One can specify multiple assert checks to check each metric with its reference value. When multiple items are specified, the operation is a logical AND and all checks
+    must be ``True``.
+
+    Returns:
+        bool: True or False for performance check ``assert_le``
+    """
+
+    # a list containing booleans to evaluate reference check for each metric
+    assert_check = []
+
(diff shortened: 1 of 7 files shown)


buildtest/buildsystem/checks.py
```diff
@@ -316,6 +316,69 @@ def assert_ge_check(builder):
     return all(assert_check)
 
 
+def assert_le_check(builder):
+    """Perform check on assert less than and equal when ``assert_le`` is specified in buildspec. The return is a boolean value that determines if the check has passed.
+    One can specify multiple assert checks to check each metric with its reference value. When multiple items are specified, the operation is a logical AND and all checks
+    must be ``True``.
+
```
shahzebsiddiqui, line 326: can you please add the `Args` for `builder` so it shows up in docstring. See other methods in the same file

docs/buildspecs/performance_checks.rst
```diff
@@ -88,6 +88,41 @@ Let's run ``buildtest inspect query -o stream_test`` to retrieve the test detail
...
+
+.. dropdown:: ``buildtest build -b tutorials/perf_checks/assert_le.yml``
+
+    .. command-output:: buildtest build -b tutorials/perf_checks/assert_le.yml
+
+
+.. dropdown:: ``buildtest inspect query -o assert_le_example``
+
+    .. command-output:: buildtest inspect query -o assert_le_example 
+
```
shahzebsiddiqui: i think this text was a copy of the another section in docs in that case i am leaning on having one example that shows all the keys `assert_le`, `assert_lt`, `assert_gt` and `assert_ge` so we dont have to rewrite docs for each section.

tutorials/perf_checks/assert_le.yml
```diff
@@ -0,0 +1,46 @@
...
+    status:
+      assert_le:
+        - name: copy
+          ref: 5000
+        - name: scale
+          ref: 5500
+        - name: add
+          ref: 6000
+        - name: triad
+          ref: 6500
```
shahzebsiddiqui, line 46: i think for now let's keep these files but i think we may want to merge this one test into one single test such that we can have all the checks in one test

buildtest/buildsystem/checks.py
Xiangs18: I already added on both assert_ge_check and assert_le_check

docs/buildspecs/performance_checks.rst
Xiangs18: If that's the case, we probably need to change ``Assert Greater Equal`` title to something like ``Performance Check`` and then put 4 tests below.

shahzebsiddiqui: yeah i agree with you, let's try to update that part of docs as a separate PR until we get all the implementation added for all the relevant checks that way we can just document once for the comparison checks `>`, `>=`, `<=`, `<`.

shahzebsiddiqui:
@Xiangs18 thanks this PR looks good. I will merge this in

shahzebsiddiqui (approved):
LGTM.

PR #1378, opened 2023-02-08
add a cron schedule to regression test workflow

### LINKED ISSUE #1347: run nightly regression test
Currently we have the [regression workflow](https://github.com/buildtesters/buildtest/blob/devel/.github/workflows/regression.yml) run on every PR which is okay for development. 

Let's extend the CI check to run nightly regression test on the latest release which is the `master` branch. 

See https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule for some tips on how to run scheduled workflows 

@prathmesh4321 this could be a good task for you to implement if you are new to github workflows.

prathmesh4321:
Hi @shahzebsiddiqui. I have added the nightly regression test. Please review the PR.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -22,15 +22,7 @@ jobs:
         python-version: ['3.7', '3.8', '3.9', '3.10']
 
     steps:
-    - name: Run actions/checkout@v3
-      if: ${{ github.event_name != 'schedule' }}
-      uses: actions/checkout@v3
-
-    - name: Run nightly regression test on master branch
-      if: ${{ github.event_name == 'schedule' }}
-      uses: actions/checkout@v3
-      with:
-        ref: master
+    - uses: actions/checkout@v3
 
     - name: Setup Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4.5.0
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -7,6 +7,9 @@ on:
       - 'buildtest/**'
       - 'tests/**'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/regression.yml
```diff
@@ -19,7 +22,15 @@ jobs:
...
-    - uses: actions/checkout@v3
+    - name: Run actions/checkout@v3
+      if: ${{ github.event_name != 'schedule' }}
+      uses: actions/checkout@v3
+
+    - name: Run nightly regression test on master branch
+      if: ${{ github.event_name == 'schedule' }}
+      uses: actions/checkout@v3
+      with:
+        ref: master
```
shahzebsiddiqui: okay how about let's focus on testing the `devel` branch of the code on nightly basis. I particularly dont like the fact we have this `if` condition.

prathmesh4321: @shahzebsiddiqui Okay done. Just out of curiosity, what issues do you think the `if` condition can cause ?

shahzebsiddiqui: yeah i think the workflow could get hard to troubleshoot considering there are now 2 more steps and only one step will run. Originally i was envisioning a new workflow file (`.yml`) but this works better.

shahzebsiddiqui: the only benefit of a new workflow file is that we can get a new badge and be able to see the nightly builds. Right now this badge is used for this workflow which typically corresponds to the check run from the PR
Screen Shot 2023-02-10 at 9 57 02 AM

shahzebsiddiqui (approved):

PR #1384, opened 2023-02-10
Created new workflow for nightly ci

### LINKED ISSUE #1347: run nightly regression test

prathmesh4321:
Hi @shahzebsiddiqui . I have the new workflow for nightly regression. Please review.

### DIFF
--- a/.github/workflows/nightly_regression.yml
+++ b/.github/workflows/nightly_regression.yml
@@ -1,12 +1,9 @@
 name: Nightly Regression Test
 
 on:
-  push:
-    branches: [devel]
-  pull_request:
-    branches: [devel]
+ 
   schedule:
-    - cron: '*/5 * * * *'
+    - cron: '0 0 * * *'
 
 jobs:
 
--- a/.github/workflows/nightly_regression.yml
+++ b/.github/workflows/nightly_regression.yml
@@ -7,12 +7,10 @@ on:
     branches: [devel]
   schedule:
     - cron: '*/5 * * * *'
-  workflow_dispatch:
-
 
 jobs:
 
-  buildtest_regtest:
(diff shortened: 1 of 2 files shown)


shahzebsiddiqui:
please add the ci badge to `README` file and then we can try to clean up the `on` event to just limit to `schedule`. We should let it run every 5min and see if it works. If it doesn't we may need to have

```
on:
  pull_requests:
    branches: [devel]
  push:
    branches: [devel]
```

.github/workflows/nightly_regression.yml
```diff
@@ -0,0 +1,53 @@
+name: Nightly Regression Test
+
+on:
+  push:
+    branches: [devel]
+  pull_request:
+    branches: [devel]
```
shahzebsiddiqui: lets remove this and then we can see how this goes tomorrow

.github/workflows/nightly_regression.yml
```diff
@@ -0,0 +1,53 @@
...
+  schedule:
+    - cron: '*/5 * * * *'
```
shahzebsiddiqui: change this back to once a day.

prathmesh4321: @shahzebsiddiqui Okay

PR #1385, opened 2023-02-10
add yamllinter  and shell linter check 

shahzebsiddiqui:


shahzebsiddiqui:
@prathmesh4321 @Xiangs18 i will be adding this `yamllint` stylecheck to the stylecheck this will cause several files to start failing. I tried to fix several files but didnt do all of them. When you do get chance please do fix them in subsequent PRs.

Issue #1388, opened 2023-02-13
[FEATURE]: add global option for `--pager`


shahzebsiddiqui:
### Please describe your feature request

currently we have `--pager` option as a sub-command for several commands including new features #1210 and #1211 for adding paging support for `buildtest history` and `buildtest inspect`.

Let's create a top-level option `buildtest --pager` and remove the options in each subcommand.

### Post question in Slack

- [X] I agree that I posted my question in [slack](https://hpcbuildtest.slack.com/) before creating this issue

### Is there an existing issue

- [X] I confirm there is no existing [issue](https://github.com/buildtesters/buildtest/issues) for this issue

shahzebsiddiqui:
@jscook2345 following up on your recommendation in #1387 for `--row-count` we could probably do the same for `--pager` support. Does that make sense?

Currently we put options in each subcommand since option is not applicable for all subcommands but `--pager` is used with several commands so duplicating the option in each subcommand was more work and creating one option can help.

PR #1392, opened 2023-02-16
add support for global pager option 

shahzebsiddiqui:
@prathmesh4321 @Xiangs18  @jscook2345. 

This will address #1388

This PR will do the following
1. add global option `buildtest --pager` and remove `--pager` option from each subcommands in particular `buildtest config view`, `buildtest buildspec find`, `buildtest buildspec summary`, `buildtest report`
2. Update bash completion script 
3. Update documentation 
4. Refactor some tests by adding comments on what command is being tested
5. Remove some test pertaining to `buildtest rt sm` where paging with color was being tested which is not necessary since paging output doesn't show color.
6. Update table `buildtest help` to account for `--pager` option

With this change, we will remove the duplication of `--pager` option when building out the command line options however user may not know when `--pager` will work. Not all commands support paging, it only works with tables. Perhaps we have a brief section on paging and show some commands where paging is supported.

Can you please review this PR

### DIFF
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -195,7 +195,7 @@ _buildtest ()
       ;;
 
     report|rt)
-      local opts="--end --fail --filter --filterfields --format --formatfields --help --helpfilter --helpformat --latest --no-header --oldest --pager --pass --row-count --start --terse  -e -f -h -n -p -s -t c clear l list p path sm summary"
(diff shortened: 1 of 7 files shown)


prathmesh4321 (approved):
Hi @shahzebsiddiqui , LGTM !

shahzebsiddiqui:
thanks @prathmesh4321

PR #1394, opened 2023-02-16
Add status check for Symlink check

### LINKED ISSUE #1330: symlink check for status
similar to #1327 where we check for files we can have a check for symbolic link checks.

This could be done something like this with a field name `is_symlink` which expects a list of values that are checked for symbolic link

```
buildspecs:
 symlink_test:
   type: script
   executor: generic.local.bash
   description: status check based on symbolic link
   run: |
      ln -s /tmp scratch
   status:
     is_symlink:
       - scratch
```

prathmesh4321:


buildtest/buildsystem/checks.py
```diff
@@ -151,6 +152,34 @@ def regex_check(builder):
...
+    Returns:
+        bool: A boolean for is_symlink status check
+    """
+    assert_exists = all(os.path.islink(file) for file in builder.status["is_symlink"])
+    console.print(
+        f"[blue]{builder}[/]: Check all items:  {builder.status['is_symlink']}  for symbolic links"
+    )
+    for sym_link in builder.status["is_symlink"]:
+        resolved_sym_link = resolve_path(sym_link)
+        if os.path.islink(sym_link):
```
shahzebsiddiqui: wait don't you want to use `resolve_sym_link` when using with `os.path.islink` since this will return a resolved path. Note `resolve_path` can return `None` if file doesn't exist. 

I am pretty sure if you do a symlink with directory expansion (`$HOME/abc`) or shell expansion  (`~/abc`) this will fail

tutorials/test_status/is_symlink.yml
```diff
@@ -0,0 +1,10 @@
+buildspecs:
+  symlink_test:
+   type: script
+   executor: generic.local.bash
+   description: status check based on symbolic link
+   run: |
+      ln -s /tmp scratch
```
shahzebsiddiqui, line 7: can you try to add shell expansion and variable expansion so 
```
ln -s $HOME/.bashrc $HOME/bashrc
```

tutorials/test_status/is_symlink.yml
```diff
@@ -0,0 +1,10 @@
...
+   status:
+     is_symlink:
+       - scratch
```
shahzebsiddiqui: over here try to add  the following
```yaml
is_symlink:
  - scratch
  - $HOME/bashrc
  - ~/bashrc
```

buildtest/buildsystem/checks.py
prathmesh4321: @shahzebsiddiqui Okay , I have updated the function to handle directory expansion, shell expansion and broken symlinks. In below screenshot bashrc is the broken symlink since .bashrc file does not exist on my system.

Screen Shot 2023-02-16 at 7 21 56 PM

buildtest/buildsystem/checks.py
```diff
@@ -161,23 +161,30 @@ def is_symlink_check(builder):
...
     """
-    assert_exists = all(os.path.islink(file) for file in builder.status["is_symlink"])
+    assert_exists = []
     console.print(
         f"[blue]{builder}[/]: Check all items:  {builder.status['is_symlink']}  for symbolic links"
     )
     for sym_link in builder.status["is_symlink"]:
+
+        expanded_sym_link = os.path.expandvars(sym_link)
+        expanded_sym_link = os.path.expanduser(expanded_sym_link)
```
shahzebsiddiqui: These commands are not necessary because this is already covered in resolve_path which takes into account variable and shell expansion see https://github.com/buildtesters/buildtest/blob/devel/buildtest/utils/file.py#L148-L150 

So what i would do is create another method in https://github.com/buildtesters/buildtest/blob/devel/buildtest/utils/file.py such as `is_symlink` which is similar to `is_file` but will return `True` or `False` if file is symlink.

Then what you can do is just call 

```
  if is_symlink(sym_link):
    console.print(...)
```

This way we have a portable method `is_symlink` that can be called in the codebase when needed.

prathmesh4321: Yeah, makes sense to have a portable method. I think the above commands (expandvars, expanduser) are required because `os.path.islink` requires shell and variable expanded path to the symlink. The `resolve_path` function along with shell and variable expansion resolves the symlink as well and returns full path to the location the symlink is pointing to (`os.path.realpath` in the function resolves the symlink), which doesn't satisfy the input requirements of `os.path.islink`. Additionally, `os.path.islink` doesn't take care of broken symlinks, it returns `true` if the symlink is broken. So I have used the fully resolved symlink to check that.

buildtest/buildsystem/checks.py
```diff
@@ -167,13 +167,9 @@ def is_symlink_check(builder):
...
 
-        expanded_sym_link = os.path.expandvars(sym_link)
-        expanded_sym_link = os.path.expanduser(expanded_sym_link)
-        resolved_sym_link = resolve_path(sym_link)
-
-        if os.path.islink(expanded_sym_link) and resolved_sym_link:
+        if is_symlink(sym_link):
             console.print(
-                f"[blue]{builder}[/]: item: {sym_link} is a symbolic link to {resolved_sym_link}"
+                f"[blue]{builder}[/]: item: {sym_link} is a symbolic link to {resolve_path(sym_link)}"
```
shahzebsiddiqui: please remove `item`

buildtest/utils/file.py
```diff
@@ -57,6 +57,28 @@ def is_dir(dirname):
     return os.path.isdir(dirname)
 
 
+def is_symlink(sym_link):
```
shahzebsiddiqui: could change to `filename` to avoid confusion with method `is_symlink`

buildtest/utils/file.py
```diff
@@ -57,6 +57,28 @@ def is_dir(dirname):
...
+    """Check if the given link is a symlink and return ``True``, if it is not a symlink or the symlink is broken return ``False``.
+
+    Args:
+        sym_link (str): input link to check.
```
shahzebsiddiqui: Could change this to `filename (str): Input filename to check if its a symbolic link`

buildtest/utils/file.py
```diff
@@ -57,6 +57,28 @@ def is_dir(dirname):
...
+
+    Returns:
+        bool: True if input link is a symlink and is not broken otherwise return False.
```
shahzebsiddiqui: change to ` bool: True if filename is a symbolic link otherwise return False if its not a symbolic link or broken link`

Remember to use the full word `symbolic link` when referring to documentation (docstring) as pose to `symlink`

buildtest/utils/file.py
```diff
@@ -57,6 +57,28 @@ def is_dir(dirname):
...
+    """
+
+    # apply shell expansion  when file includes something like $HOME/example
+    expanded_sym_link = os.path.expandvars(sym_link)
```
shahzebsiddiqui: could rename this variable to `expanded_filepath`

buildtest/buildsystem/checks.py
```diff
@@ -169,12 +169,12 @@ def is_symlink_check(builder):
 
         if is_symlink(sym_link):
```
shahzebsiddiqui: can you change the `sym_link` variable to `fname` or `filename` when u are going through the for loop. It will help with code readability. There is a variable `sym_link` then there is a method `is_symlink` and then we also have this method `is_symlink_check` so it is already getting quite confusing so best to use meaningful variable names.

shahzebsiddiqui:
@prathmesh4321 apart from this please try to add the following:
1. Documentation
2. Test Coverage
3. Add valid example on `is_symlink` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/examples/script.schema.json/valid/examples.yml

So far looking pretty good

buildtest/buildsystem/checks.py
prathmesh4321: Okay done

prathmesh4321:
> @prathmesh4321 apart from this please try to add the following:
> 
> 1. Documentation
> 2. Test Coverage
> 3. Add valid example on `is_symlink` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/examples/script.schema.json/valid/examples.yml
> 
> So far looking pretty good

Okay, I have added the test coverage. Yet to add the documentation and example.

Screen Shot 2023-02-17 at 6 20 31 PM

buildtest/buildsystem/checks.py
shahzebsiddiqui: Thanks for clarification. Yeah you are right, i saw the api and we will need to keep this logic

tests/utils/test_file.py
```diff
@@ -35,6 +36,38 @@ def test_checking_file():
     assert is_file("/bin/bash")
 
 
+@pytest.mark.utility
+def test_is_symlink():
+    # Target path of the symbolic link
+    link_target = os.path.join("$HOME", "test_dir")
+    # creating directory
+    create_dir(link_target)
```
shahzebsiddiqui: please don't do this instead create a temporary directory via https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp and then you can create a symbolic link to `$HOME/test_dir_symlink` to name of directory.


Just testing this out

```
>>> import tempfile
>>> tmpdir = tempfile.mkdtemp()
>>> tmpdir 
'/var/folders/1m/_jjv09h17k37mkktwnmbkmj0002t_q/T/tmpxgcap2qt'

>>> import os
>>> os.path.isdir(tmpdir)
True

```

docs/buildspecs/buildspec_overview.rst
```diff
@@ -447,6 +448,28 @@ Let's build the test and see the output.
 
    .. command-output:: buildtest build -b tutorials/test_status/file_and_dir_check.yml
 
+.. _symlink_check:
+
+Passing test based on Symbolic Link
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+buildtest can configure PASS/FAIL of test based on the status of symbolic link. This can be useful if your test will create a symbolic link to a file or directory and 
+will pass if the symbolic link is not broken.
```
shahzebsiddiqui: could replace `and will pass if the symbolic link is not broken` -> `and test will pass if the symbolic link is present.`

docs/buildspecs/buildspec_overview.rst
```diff
@@ -447,6 +448,28 @@ Let's build the test and see the output.
...
+.. _symlink_check:
+
+Passing test based on Symbolic Link
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+buildtest can configure PASS/FAIL of test based on the status of symbolic link. This can be useful if your test will create a symbolic link to a file or directory and 
+will pass if the symbolic link is not broken.  
+
+You can use the ``is_symlink`` property which expects a list of values that are checked for symbolic links.
+In the example below, the test will pass as all the values are valid symbolic links and are not broken. Note that variable and shell expansion
```
shahzebsiddiqui: Either add a new line to start a new paragraph or remove the new line. Right now it looks like a new line is added in middle of a paragraph

prathmesh4321: @shahzebsiddiqui Okay, I have changed this.

tests/utils/test_file.py
```diff
@@ -39,15 +39,13 @@ def test_checking_file():
 @pytest.mark.utility
 def test_is_symlink():
     # Target path of the symbolic link
-    link_target = os.path.join("$HOME", "test_dir")
-    # creating directory
-    create_dir(link_target)
+    link_target = tempfile.mkdtemp()
 
     # Symbolic link path
     link_path = os.path.join("$HOME", "test_dir_symlink")
```
shahzebsiddiqui: please try creating a temporary file in $HOME directory as pose to hardcoding this path.

What you can do is create a temporary directory something like this

```
>>> f = tempfile.TemporaryDirectory(dir = os.path.expanduser('~'))
>>> f
<TemporaryDirectory '/Users/siddiq90/tmpyn5w9lz1'>
```

Then create a symbolic link to let's say `/tmp`.

It would be better to put this in a context manager so file is closed and deleted upon termination see https://docs.python.org/3/library/tempfile.html#examples

something like this

```
# create a temporary directory using the context manager
with tempfile.TemporaryDirectory() as tmpdirname:
    print('created temporary directory', tmpdirname)
```

prathmesh4321: So if we try to create directory/file and then use `os.symlink()` then it throws 'FileExists' error. Seems like os.symlink() requires a filename which does not already exists.

Screen Shot 2023-02-21 at 1 19 51 PM

tests/utils/test_file.py
```diff
@@ -39,31 +39,36 @@ def test_checking_file():
...
+    link_path = tempfile.NamedTemporaryFile(dir=os.path.expanduser("~"))
+
+    link_path.close()
 
     # Create symbolic link
-    os.symlink(link_target, os.path.expandvars(link_path))
+    os.symlink(link_target.name, link_path.name)
+
+    # get filename from the path
+    filename = os.path.split(link_path.name)[-1]
```
shahzebsiddiqui: i think this should work but it is a bit cryptic using arrays. I think this is equivalent to running [os.path.basename](https://docs.python.org/3/library/os.path.html#os.path.basename) you can do `os.path.basename(link_path.name)`. I think these two statements are equivalent but a bit more safer.

shahzebsiddiqui:
thanks @prathmesh4321 the CI checks are passing now that symlink is removed. Great job!

PR #1408, opened 2023-03-02
Add code coverage for subprocesses

### LINKED ISSUE #1398: investigate regression test coverage for checks.py
@prathmesh4321 

currently we see https://app.codecov.io/github/buildtesters/buildtest/blob/devel/buildtest/buildsystem/checks.py showing a 5% coverage despite the fact we are attempting to test each status check in https://github.com/buildtesters/buildtest/blob/devel/tests/builders/test_builders.py 
Screen Shot 2023-02-22 at 2 29 38 PM

According to test coverage it is showing all methods have no coverage so something is going wrong w.r.t to line coverage. We should look into why this is happening since test are running but not impacting test coverage.

Screen Shot 2023-02-22 at 2 30 50 PM

Currently, we are invoking the BuildTest class see https://github.com/buildtesters/buildtest/blob/0141fa2d87d713db8f8b3650b359c6201ab2d482/buildtest/cli/build.py#L468-L499 to emulate the `buildtest build` command which should run the test when invoking the `build` method. Based on output of regression test it is running the checks but line-coverage is not hitting each method

prathmesh4321:
@shahzebsiddiqui Looks like code coverage works now. Check this - https://app.codecov.io/gh/buildtesters/buildtest/blob/coverage_subprocesses/buildtest/buildsystem/checks.py

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,9 @@
 [run]
 branch = True
+parallel = True
+concurrency = multiprocessing
 source = buildtest
+
 omit =
     buildtest/main.py
     buildtest/defaults.py
--- a/buildtest/tools/unittests.py
+++ b/buildtest/tools/unittests.py
@@ -72,6 +72,7 @@ def run_unit_tests(pytestopts=None, sourcefiles=None, enable_coverage=False):
 
     if enable_coverage:
         cov.stop()
+        cov.combine()
         cov.html_report(title="buildtest unittests coverage report", directory=html_dir)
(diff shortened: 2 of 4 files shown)


prathmesh4321:
@shahzebsiddiqui Looks like code coverage works now. Check this - https://app.codecov.io/gh/buildtesters/buildtest/blob/coverage_subprocesses/buildtest/buildsystem/checks.py

shahzebsiddiqui:
@prathmesh4321 great job. Yeah i see there is a big change in coverage +5.6% that huge.

shahzebsiddiqui (approved):
i think this looks good

PR #1419, opened 2023-03-08
Build by multiple tags by providing tagnames as comma seperated list

### LINKED ISSUE #1409: [FEATURE]: specify multiple tagnames as comma separated list

currently if one wants to specify multiple tags one has to do `buildtest bd -t TAG1 -t TAG2`. Let's keep the append logic for now it would be nice if we can do `buildtest bd -t TAG1,TAG2` which is equivalent to two tag names.

We should still be able to preserve append feature so it should be possible to do `buildtest bd -t TAG1,TAG2 -t TAG3`.

TODO:
- [x] Add implementation 
- [x] Add regression test in https://github.com/buildtesters/buildtest/blob/devel/tests/cli/test_build.py
- [x] Add documentation in in https://buildtest.readthedocs.io/en/devel/gettingstarted/buildingtest.html#building-by-tags
- [x] Update `buildtest help build`

### Post question in Slack

prathmesh4321:
Hi @shahzebsiddiqui . Please review the PR.

Screen Shot 2023-03-08 at 12 58 36 PM

Screen Shot 2023-03-08 at 12 58 46 PM

### DIFF
--- a/buildtest/cli/build.py
+++ b/buildtest/cli/build.py
@@ -308,14 +308,16 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
     buildspecs_by_tags = {}
 
     buildspecs = []
-    multiple_taglist = []
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
-  
-    for name in tagnames:
-        multiple_taglist.extend(name.split(","))
+    tagnames = [
+        tag.strip() for tagname in tagnames for tag in tagname.split(",") if tag != ""
+    ]
 
-    for name in multiple_taglist:
+    if not tagnames:
+        return buildspecs, buildspecs_by_tags
+
+    for name in tagnames:
         buildspecs_by_tags[name] = set()
 
         for buildspecfile in buildspec_cache["buildspecs"].keys():
--- a/buildtest/cli/build.py
+++ b/buildtest/cli/build.py
@@ -308,10 +308,14 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
     buildspecs_by_tags = {}
 
     buildspecs = []
+    multiple_taglist = []
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
-
+  
     for name in tagnames:
+        multiple_taglist.extend(name.split(","))
+
+    for name in multiple_taglist:
         buildspecs_by_tags[name] = set()
 
         for buildspecfile in buildspec_cache["buildspecs"].keys():
--- a/buildtest/cli/build.py
+++ b/buildtest/cli/build.py
@@ -311,7 +311,7 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
     tagnames = [
-        tag.strip() for tagname in tagnames for tag in tagname.split(",") if tag != ""
+        tag.strip() for tagname in tagnames for tag in tagname.split(",") if tag.strip()
     ]
 
     if not tagnames:
--- a/buildtest/cli/help.py
+++ b/buildtest/cli/help.py
@@ -26,8 +26,8 @@ def print_build_help():
         "Build buildspecs by tagname 'pass' and 'python'",
     )
     table.add_row(
-        "buildtest build -t pass,fail -t network",
-        "Build buildspecs by tagnames that are specified as comma sperated list"
+        "buildtest build -t pass,fail",
+        "Build buildspecs by tagnames that are specified as comma sperated list",
     )
     table.add_row(
         "buildtest build -e <executor1> -e <executor2>",
--- a/buildtest/cli/help.py
+++ b/buildtest/cli/help.py
@@ -26,6 +26,10 @@ def print_build_help():
         "Build buildspecs by tagname 'pass' and 'python'",
     )
     table.add_row(
+        "buildtest build -t pass,fail -t network",
+        "Build buildspecs by tagnames that are specified as comma sperated list"
+    )
+    table.add_row(
         "buildtest build -e <executor1> -e <executor2>",
         "Building buildspecs by executor",
     )
--- a/docs/gettingstarted/buildingtest.rst
+++ b/docs/gettingstarted/buildingtest.rst
@@ -138,8 +138,8 @@ tags by running ``buildtest buildspec find --tags``.
 .. Note:: The ``--tags`` is used for discovering buildspec file and not filtering tests
    by tag.
 
-You can also build by multiple tags by providing tagnames as comma seperated list. In the below
-example we build all tests with tag names ``pass``, ``fail`` and ``network``. Note that you can still specify ``--tags`` multiple times.
+You can also build by multiple tags by providing tagnames as comma seperated list. In the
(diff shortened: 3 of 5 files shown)


docs/gettingstarted/buildingtest.rst
```diff
@@ -138,6 +138,13 @@ tags by running ``buildtest buildspec find --tags``.
 .. Note:: The ``--tags`` is used for discovering buildspec file and not filtering tests
    by tag.
 
+You can also build by multiple tags by providing tagnames as comma seperated list. In the below
```
shahzebsiddiqui: change `In the below example` -> `In the example below`

docs/gettingstarted/buildingtest.rst
```diff
@@ -138,6 +138,13 @@ tags by running ``buildtest buildspec find --tags``.
...
+example we build all tests with tag names ``pass``, ``fail`` and ``network``. Note that you can still specify ``--tags`` multiple times.
```
shahzebsiddiqui: I think you can remove the `Note that you can still ...` this was already mentioned above.

buildtest/cli/help.py
```diff
@@ -25,6 +25,10 @@ def print_build_help():
         "buildtest build -t pass -t python",
         "Build buildspecs by tagname 'pass' and 'python'",
     )
+    table.add_row(
+        "buildtest build -t pass,fail -t network",
```
shahzebsiddiqui: i think you should remove the `-t network` since we do the appending example above. This will match the description for this row entry

tests/cli/test_build.py
```diff
@@ -46,6 +46,14 @@ def test_build_by_tags():
     )
     cmd.build()
 
+    #  testing multiple tags as comma seperated list:  buildtest build --tags fail,pass --tags python,ping --tags network
+    cmd = BuildTest(
+        configuration=configuration,
+        tags=["fail,pass", "python,ping", "network"],
+        buildtest_system=system,
+    )
+    cmd.build()
```
shahzebsiddiqui: we can cut down on number of tags to avoid lots of builds so it can be sufficient to just do  `tags = ["fail,python", "network"]`.

It would be good exercise to also test for edge cases such as comma separated without any tags or a tagname with a trailing comma

```
buildtest build --tags ,,
buildtest build --tags python,
buildtest build --tags ,python
```

buildtest/cli/build.py
```diff
@@ -308,10 +308,14 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
...
     buildspecs = []
+    multiple_taglist = []
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
-
+  
     for name in tagnames:
+        multiple_taglist.extend(name.split(","))
+
+    for name in multiple_taglist:
```
shahzebsiddiqui: i think if you run  `buildtest build -t ,, ` there is no need to go through the entire for loop to discover the buildspecs by tags. 

We should expect to see `tagnames=",,"` and the value of `multiple_taglist` would be an empty list if that is the case we may need to return 
That being said i think we can add a condition like this prior to doing the `for name in multiple_taglist`

```
if not multiple_taglist:
   return buildspecs_by_tags, buildspecs
```

buildtest/cli/build.py
```diff
@@ -308,10 +308,14 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
     buildspecs_by_tags = {}
 
     buildspecs = []
+    multiple_taglist = []
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
-
+  
     for name in tagnames:
+        multiple_taglist.extend(name.split(","))
```
shahzebsiddiqui: can we avoid declaring a new variable  `multiple_taglist` instead update tagnames variable as follows

```
tagnames = [tag.strip() for tagname in tagnames for tag in tagname.split(",")]
```

Just tried this and it does work

```
>>> tagnames
['fail, pass', 'python', 'network']
>>> tagnames = [tag.strip() for tagname in tagnames for tag in tagname.split(",")]
>>> tagnames
['fail', 'pass', 'python', 'network']
```

It works in the normal usecase where no comma is found in element of list.

```
>>> tagnames = ['python','fail']
>>> tagnames = [tag.strip() for tagname in tagnames for tag in tagname.split(",")]
>>> tagnames
['python', 'fail']

```

buildtest/cli/build.py
```diff
@@ -310,6 +310,12 @@ def discover_buildspecs_by_tags(buildspec_cache, tagnames):
     buildspecs = []
     # query all buildspecs from BUILDSPEC_CACHE_FILE for tags keyword and
     # if it matches input_tag we add buildspec to list
+    tagnames = [
+        tag.strip() for tagname in tagnames for tag in tagname.split(",") if tag != ""
```
shahzebsiddiqui: i dont like the `if tag != ""` in the list comprehension and it doesnt work all the time

i tried this with a few use-cases where i have spaces between commas and the strip was working

```
>>> tagnames = ['fail, pass', 'python', 'network', " ,  "]
>>> b =  [tag.strip() for tag in tagnames for tag in tag.split(',') if tag.strip()]
>>> b
['fail', 'pass', 'python', 'network']


>>> tagnames = ['fail, pass', 'python', 'network', " , ab ,  "]
>>> b =  [tag.strip() for tag in tagnames for tag in tag.split(',') if tag.strip()]
>>> b
['fail', 'pass', 'python', 'network', 'ab']

```

However this same example with your if condition returns empty characters

```
>>> tagnames = ['fail, pass', 'python', 'network', " , ab ,  "]
>>> b = [ tag.strip() for tagname in tagnames for tag in tagname.split(",") if tag != "" ] 
>>> b
['fail', 'pass', 'python', 'network', '', 'ab', '']

```

docs/gettingstarted/buildingtest.rst
```diff
@@ -138,6 +138,13 @@ tags by running ``buildtest buildspec find --tags``.
 .. Note:: The ``--tags`` is used for discovering buildspec file and not filtering tests
    by tag.
 
+You can also build by multiple tags by providing tagnames as comma seperated list. In the
```
shahzebsiddiqui: spelling error `seperated` -> `separated`.

Also you can change this sentence `You can also build by multiple tags by providing ...` --> `You can specify multiple tag names as a comma separated list.`.

tests/cli/test_build.py
```diff
@@ -46,6 +46,36 @@ def test_build_by_tags():
...
+        configuration=configuration,
+        tags=[",,"],
+        buildtest_system=system,
+    )
+    with pytest.raises(SystemExit):
+        cmd.build()
+
+    cmd = BuildTest(
+        configuration=configuration,
+        tags=["python,"],
```
shahzebsiddiqui: now that i think about this, lets change this to something like this, which should capture all the edge cases.  `tags=["  ,python,  fail,   ,,"]`.

This way we can remove the following statement

```
    cmd = BuildTest(
        configuration=configuration,
        tags=[",python"],
        buildtest_system=system,
    )
    cmd.build()
```

There are few edge cases that i can think of 
1. comma at start `buildtest bd -t ,python`
2. comma at end `buildtest bd -t python,`
3. comma with spaces `buildtest bd -t python,  fail,network  ,`

buildtest/cli/build.py
prathmesh4321: @shahzebsiddiqui Good catch! Have fixed this now.

shahzebsiddiqui (approved):
looks good!

PR #1420, opened 2023-03-09
buildtest config compilers find --file

### LINKED ISSUE #1375: [FEATURE]: buildtest config compilers find --file

current we have `buildtest config compilers find --update` write the content of configuration into the same configuration file, it would be nice to have an option to specify an alternative filepath where configuration file will be run

For instance if someone wants to write file to `/tmp/buildtest.config.yml`
 
 `buildtest config compilers find --file /tmp/buildtest.config.yml -u`

### Post question in Slack

Xiangs18:


buildtest/cli/compilers.py
```diff
@@ -127,6 +128,7 @@ def compiler_find(configuration, modulepath=None, detailed=None, update=None):
         modulepath (List, optional): An instance of List, a list of directories to search for modules via MODULEPATH to detect compilers
         detailed (bool, optional): An instance of bool, flag for printing a detailed report.
         update (bool, optional): An instance of bool, flag for updating configuration file with new compilers
+        filepath (str, optional): An instance of str, an option to specify an alternative filepath where configuration file will be run
```
shahzebsiddiqui: I think we can remove `An instance of` from all the variables. 

Also can you change `run` -> `written`. Since the configuration file is written.

buildtest/cli/compilers.py
```diff
@@ -151,14 +153,17 @@ def compiler_find(configuration, modulepath=None, detailed=None, update=None):
     # print out all compilers from existing configuration file
     # run buildtest config compilers find --update to update existing configuration file
 
+    if filepath and filepath.split(".")[-1] != ".yml":
+        raise ValueError("alternative file specified should be a yml file")
+
```
shahzebsiddiqui: i think we should also check for edge case and raise exception when `--file` specifies a directory path `i.e buildtest cg co find -u /tmp` should raise an exception.

As per the configuration file we are not forcing `.yml` extension right now so i am not sure if we should force this check.

buildtest/cli/compilers.py
```diff
@@ -151,14 +153,18 @@ def compiler_find(configuration, modulepath=None, detailed=None, update=None):
     # print out all compilers from existing configuration file
     # run buildtest config compilers find --update to update existing configuration file
 
+    # check for edge case and raise exception when --file specifies a directory path
+    if filepath and os.path.isdir(filepath):
```
shahzebsiddiqui: i think you should use `is_dir(filepath)` instead of `os.path.isdir` because we also take into account variable expansion. that being said i think you may want to do a little bit more.

do the following

```
resolved_filepath = resolve_path(filepath, exist=False)

if is_dir(resolved_filepath):
  raise BuildTestError(f"The file: {resolved_filepath} is a directory, please specify a file path")
```

You may want to wrap this around a

```
if filepath:
```

I am not sure what happens when `None` is specified to `resolve_path`. 

Alternatively you can specify above your code something like this

```
dest_filepath = os.path.join(os.path.dirname(configuration.file), fname)

if filepath:
  resolved_filepath = resolve_path(filepath, exist=False)

  if is_dir(resolved_filepath):
    raise BuildTestError(f"The file: {resolved_filepath} is a directory, please specify a file path")
  dest_filepath = resolved_filepath
```

buildtest/cli/__init__.py
```diff
@@ -1006,6 +1006,12 @@ def config_menu(subparsers, parent_parser):
         action="store_true",
         help="Update configuration file with new compilers",
     )
+    compiler_find.add_argument(
+        "-f",
+        "--file",
+        action="store_true",
+        help="Wrtie configuration file to a new file",
+    )
```
shahzebsiddiqui: can you please use [parent parser](https://docs.python.org/3/library/argparse.html#parents) for defining `--file` option so we can potentially use this option in other subcommands.

shahzebsiddiqui: create it as a separate variable see https://github.com/buildtesters/buildtest/blob/869ab652bf5ef2cd5ea77232559574aa3449f82a/buildtest/cli/__init__.py#L226-L229 i think we need to have several parent parsers to help organize where options are inherited. I think we will likely see a clash with option `-f` so rule of thumb when using parent parser only have long option for now, later we can figure out which ones have short option such as `-f`

buildtest/cli/__init__.py
```diff
@@ -1027,6 +1021,11 @@ def config_menu(subparsers, parent_parser):
     compiler_test.add_argument(
         "compiler_names", nargs="*", help="Specify compiler name to test"
     )
+    # use parent parser for defining --file option
+    file_parser = argparse.ArgumentParser(parents=[parent_parser])
+    file_parser.add_argument(
+        "--file", action="store_true", help="Wrtie configuration file to a new file"
+    )
```
shahzebsiddiqui: can you please define the `file_parser` next to https://github.com/buildtesters/buildtest/blob/ff6de89df077fad36df6e212d45d296843d5edd2/buildtest/cli/__init__.py#L226 I think we should probably have some variable `self.parent_parsers` as a list and then we can pass this around to each method where needed.

Have something like this 

```
self.parent_parsers['file'] = argparse.ArgumentParser(add_help=False)
self.parent_parsers['file'].add_argument("--file", action="store_true", help="Wrtie configuration file to a new file")
```

Then you can add the parent parser in the following line

https://github.com/buildtesters/buildtest/blob/ff6de89df077fad36df6e212d45d296843d5edd2/buildtest/cli/__init__.py#L993-L996

by doing 

```
    compiler_find = subparsers_compiler.add_parser(
        "find",
        help="Find compilers",
        parents=[self.parent_parsers['file']]
    )
```

Does that make sense?.

Xiangs18: I assume I will have the following in the get_parser function:

```python
parent_parser = {}
parent_parser["pager"] = argparse.ArgumentParser(add_help=False)
parent_parser["pager"].add_argument(
        "--pager", action="store_true", help="Enable PAGING when viewing result"
    )
parent_parser["file"] = argparse.ArgumentParser(add_help=False)
parent_parser["file"].add_argument(
        "--file", action="store_true", help="Wrtie configuration file to a new file"
```

And then pass parent_parser around to each menu method where needed. 

If my understanding is correct, only in compiler_find will `parent = [self.parent_parsers['file']]`.
All others that used to have `parents=[parent_parser]` params will become` parent = [self.parent_parsers["pager"]]`? 

Thank you!!

shahzebsiddiqui: yes except change `parent_parser` to `self.parent_parser` that way we dont need to pass variable around each method. Yes i think you understand the logic. 

Depending on the subcommand you will have `parents=[self.parent_parsers['pager']` where `--pager` option is inherited. Likewise, where `--file` is inherited you can do `parents=[self.parent_parsers['file']]`. In event multiple arguments need to be inherited to a sub-command we should be able to index the dict so `parents=[self.parent_parsers['pager'], [self.parent_parsers['file']]`.

Also a tip when copying code in markdown you can use triple backticks to format code. If you want to format python code you can do ```python. Github supports syntax highlighting for several languages you should take a look at https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting

buildtest/cli/__init__.py
```diff
@@ -149,7 +149,22 @@ def valid_time(value):
     return dt_object
 
 
-def get_parser():
+class ParentParser:
```
shahzebsiddiqui: I don't see a point in having a class if all it does it holds a single variable. Can't we just declare this dictionary anywhere even a method would do which just returns the parent parser object.

The main thing is we need the parent parser object to be accessible in get_parser and any of the sub commands where we build the command menu

shahzebsiddiqui: okay i think i now see what is going on lets make this `BuildtestParser` as the main class since `get_parser` is a method and not a method of a class.

This could require some other changes for instance we can move all the methods like `build_menu`, `report_menu`, etc... as methods in the class. No need to pass arguments around we can use `self`. This will require changing the invocation in main.py https://github.com/buildtesters/buildtest/blob/ddc9dc0781a671b8d4a3b3ca5032cdae2a57bacd/buildtest/main.py#L66

Xiangs18:
Hi @shahzebsiddiqui, running command below keeps getting unrecognized arguments error. Can you take a look? 
```
buildtest config compilers find --file /tmp/buildtest.config.yml -u
```
usage: buildtest [options] [COMMANDS]
buildtest: error: unrecognized arguments: /tmp/buildtest.config.yml

buildtest/cli/__init__.py
```diff
@@ -223,12 +223,24 @@ def get_parser():
...
+        parent_parser = {}
+
+        parent_parser["pager"] = argparse.ArgumentParser(add_help=False)
+        parent_parser["pager"].add_argument(
+            "--pager", action="store_true", help="Enable PAGING when viewing result"
+        )
+
+        parent_parser["file"] = argparse.ArgumentParser(add_help=False)
+        parent_parser["file"].add_argument(
+            "--file", action="store_true", help="Wrtie configuration file to a new file"
```
shahzebsiddiqui: yes the issue is right here i missed this, please remove `action="store_true"` we need argparse to hold value of the file not a `true` or `false`. This should fix the issue

shahzebsiddiqui:
> Hi @shahzebsiddiqui, running command below keeps getting unrecognized arguments error. Can you take a look?
> 
> ```
> buildtest config compilers find --file /tmp/buildtest.config.yml -u
> ```
> 
> usage: buildtest [options] [COMMANDS] buildtest: error: unrecognized arguments: /tmp/buildtest.config.yml

see my message https://github.com/buildtesters/buildtest/pull/1420/files#r1157445437 the add_argument is not setup correctly. Let me know if this fix. the issue

shahzebsiddiqui:
@Xiangs18 i saw the gitlab runner was stuck so i restarted it now your CI job is running on Perlmutter https://software.nersc.gov/NERSC/buildtest/-/jobs/202208 hopefully we will see the test coverage results in codecov.

Xiangs18:
Screen Shot 2023-04-06 at 8 19 59 PM

PR #1428, opened 2023-03-21
Add pagination support for subcommands in "buildtest inspect" command

### LINKED ISSUE #1210: [FEATURE]: add pagination for 'buildtest inspect' command

Let's add pagination support for **buildtest inspect** command that will be done via option `--pager` and short option `-p` to be consistent.

The pagination will be done via rich see https://rich.readthedocs.io/en/stable/console.html#paging. 

See related implementation in `buildtest report` or `buildtest buildspec` for pagination in files: https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/report.py and https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/buildspec.py  

# Action Item
- [x] Add `--pager` and `-p` option in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/__init__.py 
- [x] Add support of pagination for commands `buildtest inspect buildspec`, `buildtest inspect name`, `buildtest inspect query` and `buildtest inspect list` in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/inspect.py 
- [x] Add option to bash completion script https://github.com/buildtesters/buildtest/blob/devel/bash_completion.sh
- [x] Add regression test for pagination in https://github.com/buildtesters/buildtest/blob/devel/tests/cli/test_inspect.py
- [x] Add a row entry in `buildtest help inspect` pertaining to pagination in https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/help.py

### Post question in Slack

prathmesh4321:
Hi @shahzebsiddiqui I have added pagination for buildtest inspect subcommands. I have also refactored the function `inspect_list()` as the terse option in it was using hardcoded column names without a color option. Please review the PR.

Screen Shot 2023-03-21 at 1 29 26 PM

Screen Shot 2023-03-21 at 1 35 12 PM

### DIFF
--- a/buildtest/cli/help.py
+++ b/buildtest/cli/help.py
@@ -310,8 +310,13 @@ def print_inspect_help():
         "Display all test names, ids, and corresponding buildspec file",
     )
     table.add_row("buildtest inspect list -t", "Show output in terse format")
+    table.add_row("buildtest inspect list --pager", "Paginate output of inspect list")
     table.add_row(
         "buildtest inspect name hello", "Display last run for test name 'hello'"
+    )
+    table.add_row(
+        "buildtest inspect name hello --pager",
+        "Paginate output of last run for test name 'hello'",
     )
     table.add_row(
         "buildtest inspect name hello/9ac bar/ac9",
@@ -322,8 +327,16 @@ def print_inspect_help():
         "Fetch latest runs for all tests in buildspec file 'tutorials/vars.yml'",
     )
     table.add_row(
+        "buildtest inspect buildspec tutorials/vars.yml --pager",
+        "Paginate output of test runs for the tests in file 'tutorials/vars.yml'",
+    )
+    table.add_row(
         "buildtest inspect query -o exit1_fail",
         "Display content of output file for latest run for test name 'exit1_fail'",
+    )
+    table.add_row(
+        "buildtest inspect query -o exit1_fail --pager",
+        "Paginate content of output file for latest run for test name 'exit1_fail'",
     )
     table.add_row(
         "buildtest inspect query -e hello",
--- a/buildtest/cli/help.py
+++ b/buildtest/cli/help.py
@@ -315,10 +315,6 @@ def print_inspect_help():
         "buildtest inspect name hello", "Display last run for test name 'hello'"
     )
     table.add_row(
-        "buildtest inspect name hello --pager",
-        "Paginate output of last run for test name 'hello'",
-    )
-    table.add_row(
         "buildtest inspect name hello/9ac bar/ac9",
         "Display record for test 'hello/9ac' and 'bar/ac9'. Will find first match for each test ID",
     )
@@ -327,16 +323,8 @@ def print_inspect_help():
         "Fetch latest runs for all tests in buildspec file 'tutorials/vars.yml'",
     )
     table.add_row(
-        "buildtest inspect buildspec tutorials/vars.yml --pager",
-        "Paginate output of test runs for the tests in file 'tutorials/vars.yml'",
-    )
-    table.add_row(
         "buildtest inspect query -o exit1_fail",
         "Display content of output file for latest run for test name 'exit1_fail'",
-    )
-    table.add_row(
-        "buildtest inspect query -o exit1_fail --pager",
-        "Paginate content of output file for latest run for test name 'exit1_fail'",
     )
     table.add_row(
         "buildtest inspect query -e hello",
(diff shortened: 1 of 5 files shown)


buildtest/cli/help.py
```diff
@@ -310,9 +310,14 @@ def print_inspect_help():
...
     )
     table.add_row("buildtest inspect list -t", "Show output in terse format")
+    table.add_row("buildtest inspect list --pager", "Paginate output of inspect list")
     table.add_row(
         "buildtest inspect name hello", "Display last run for test name 'hello'"
     )
+    table.add_row(
+        "buildtest inspect name hello --pager",
+        "Paginate output of last run for test name 'hello'",
+    )
```
shahzebsiddiqui: yeah i think we dont need to document every single `--pager` option for every sub-command. I think its sufficient to just have one like `buildtest inspect list --pager`. Otherwise we will end up having too many outputs.

buildtest/cli/inspect.py
```diff
@@ -88,65 +94,118 @@ def fetch_test_names(report, names):
...
+        report (str): Path to report file
+    """
+
+    builders = report.builder_names()
+    for name in builders:
+        console.print(name)
+
+
+def print_terse(table, no_header=None, consoleColor=None):
+    """This method prints the buildtest inspect list in terse mode which is run via command ``buildtest inspect list --terse``
```
shahzebsiddiqui: I think you can remove the first reference `buildtest inspect list` and just say `This method prints output of builders in terse mode via command ...`. It is redundant because we have `buildtest inspect list --terse`

buildtest/cli/inspect.py
```diff
@@ -88,65 +94,118 @@ def fetch_test_names(report, names):
...
+    Args:
+        report (str): Path to report file
+    """
+
+    builders = report.builder_names()
+    for name in builders:
+        console.print(name)
+
+
+def print_terse(table, no_header=None, consoleColor=None):
```
shahzebsiddiqui: oh we might as well be consistent with the variable names and not do camel case `consoleColor` this just doesn't look right. I think this was done by Jeff a while ago. Perhaps rename this variable to all lower case either `consolecolor` or `console_color`

buildtest/cli/inspect.py
```diff
@@ -88,65 +94,118 @@ def fetch_test_names(report, names):
...
+
+    Args:
+        table (dict): Table with columns required for the ``buildtest inspect list`` command.
+        no_header (bool, optional): Determine whether to print header in terse format.
+        consoleColor (bool, optional): Select desired color when displaying results
+    """
+
+    row_entry = [table[key] for key in table.keys()]
+    transpose_list = [list(i) for i in zip(*row_entry)]
+    header = "|".join(table.keys())
```
shahzebsiddiqui: we could probably remove this variable declaration `header` since its used once and just do 
`console.print('|'.join(table.keys()), style=consoleColor)`

buildtest/cli/inspect.py
```diff
@@ -88,65 +94,118 @@ def fetch_test_names(report, names):
...
+
+    # We print the table columns if --no-header is not specified
+    if not no_header:
+        console.print(header, style=consoleColor)
+
+    for row in transpose_list:
+        line = "|".join(row)
```
shahzebsiddiqui: same with this line can we just move this operation into `console.print`

buildtest/cli/inspect.py
```diff
@@ -88,65 +94,118 @@ def fetch_test_names(report, names):
...
-    for identifier in test_ids.keys():
-        table.add_row(
-            f"{identifier}",
-            f"{test_ids[identifier]['name']}",
-            f"{test_ids[identifier]['buildspec']}",
-        )
-    console.print(table)
+    inspect_table.add_column("id")
+    inspect_table.add_column("name")
+    inspect_table.add_column("buildspec")
```
shahzebsiddiqui: could you rewrite this as follows: 

```python
for column in table.keys():
    inspect_table.add_column(column)
```

I think both statements are equivalent but less code and also makes sure table columns are added based on keys in `table` variable

shahzebsiddiqui (changes requested):
Overall this looks pretty good i made a few suggestions to the code-base. Please do check the code coverage to see if coverage results.

prathmesh4321:
> Overall this looks pretty good i made a few suggestions to the code-base. Please do check the code coverage to see if coverage results.

@shahzebsiddiqui Yes, I have checked the code coverage it looks good.

shahzebsiddiqui (approved):
LGTM.

PR #1429, opened 2023-03-22
support for logical AND in status check via 'mode'

### LINKED ISSUE #1418: status check support with logical AND
currently each status check is evaluated separately and we perform a logical OR on all the status check to evaluate the final `state` of test https://github.com/buildtesters/buildtest/blob/6458ab865c5a7f680c0d2ae98f0314584801804c/buildtest/builders/base.py#L1052-L1076

What we need is an option to control between logical OR or AND operation we could do this by adding a property let's say `mode` which could be an optional field. Valid values are [`or`, `and`]. If field is not specified we will assume `or`. In the following test, we perform status check with `returncode` and `file_exists`. The first test will fail while second test will pass. Note both test do the example same assertion except for the field `mode`.

```yaml
buildspecs:
   status_check_logical_and:
     type: script
     executor: generic.local.bash
     description: logical and assertion on status check
     run: mkdir foo
     status:
       mode: 'and'
       returncode: 1
       file_exists:
         - foo
   status_check_logical_or:
     type: script
     executor: generic.local.bash
     description: logical and assertion on status check
     run: mkdir foo
     status:
       mode: 'or'
       returncode: 1
       file_exists:
         - foo                 
```

shahzebsiddiqui:


buildtest/schemas/definitions.schema.json
```diff
@@ -573,6 +572,11 @@
         "state": {
           "$ref": "#/definitions/state",
           "description": "explicitly mark state of test regardless of status calculation"
+        },
+        "mode": {
+          "type": "string",
+          "description": "Determine how the status check is resolved, for instance it can be logical AND or OR",
+          "enum": ["or", "and"]
```
shahzebsiddiqui: @wyphan so i would like to get your input we can basically have any keyword i proposed

we have `mode: [or|and]` but if you want to have other values we could do `OR` `AND` or `|` and `&` though i dont want to have too many values it will make it quite difficult to keep track of it. It could also be `any` `and` which map to the python built-in.

Do you have any preference or you are okay with the proposed solution.

wyphan: I think `any` and `all` might be good labels too.

shahzebsiddiqui: so remove `or` `and` instead of `any` and `all`?

shahzebsiddiqui: or do you want all of them `or`, `and`, `any`, `all`

shahzebsiddiqui: okay cool. Ill change this up

buildtest/builders/base.py
```diff
@@ -985,128 +1006,94 @@ def check_test_state(self):
...
-            returncode_match = returncode_check(self)
+            # if 'state' property is specified explicitly honor this value regardless of what is calculated
+            if self.status.get("state"):
+                self.metadata["result"]["state"] = self.status["state"]
+                return
 
-            # check regex against output or error stream based on regular expression
-            # defined in status property. Return value is a boolean
-            regex_match = regex_check(self)
+            # returncode_match is boolean to check if reference returncode matches return code from test
```
prathmesh4321: @shahzebsiddiqui You can remove this comment since we are not using the variable "returncode_match".

prathmesh4321 (changes requested):
@shahzebsiddiqui Please take a look at the above comment. Other than this the PR looks good to me !

PR #1432, opened 2023-03-31
Fix Issue with tab completion for subcommands with top level options

### LINKED ISSUE #1261: [Bug]: issue with tab completion
### Bug Description

There is an issue with tab completion with some sub-commands especially when one uses top-level options like `buildtest <option> <subcommand>` then the bash completion doesn't work properly

Not working, bash completion showing top-level argument instead of options of `bd`

```
(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest -d bd 
bc           build        cd           cg           config       debugreport  h            history      info         it           report       schema       stats        stylecheck   unittests                                     
bd           buildspec    cdash        clean        debug        docs         help         hy           inspect      path         rt           schemadocs   style        test                     
```

This is working 

```
(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest bd 
--buildspec               --executor-type           --module-purge            --procs                   --stage                   -b                        -m                        -x                                                
--disable-executor-check  --filter                  --modules                 --remove-stagedir         --tags                    -e                        -s                                                                        
--exclude                 --helpfilter              --nodes                   --rerun                   --timeout                 -et                       -t                                                                        
--executor                --maxpendtime             --pollinterval            --retry                   --unload-modules          -f                        -u                                                                        
```

Tab completion not working for `buildtest cg executors`, it is showing options for `buildtest cg` instead of the options for subcommand

```
(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest cg executors
--help     -h         co         compilers  e          edit       ex         executors  p          path       systems    v          validate   view    
```

The tab completion for `buildtest cg co` is working however for subcommands like `test` it doesn't work after tab completion. Note that buildtest tab completion worked for `buildtest cg co test` which shows *builtin_gcc* however positional argument after the argument now shows the tab completion for parent command `buildtest cg co`

```
(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest cg co find
--help     -h         co         compilers  e          edit       ex         executors  p          path       systems    v          validate   view     

(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest cg co test builtin_gcc
--help     -h         co         compilers  e          edit       ex         executors  p          path       systems    v          validate   view   

(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest cg co     
builtin_gcc
```

Tab completion for subcommands inside `buildtest rt` such as `buildtest rt c`, `buildtest rt list` or `buildtest rt sm` are not working it is showing tab completion for `buildtest rt` command such as one below

```
(buildtest)  ~/Documents/github/buildtest/ [fix_ci_issue_1257] buildtest rt c 
--color         --filter        --formatfields  --helpformat    --oldest        --row-count     -e              -n              -t              l               summary                                                   
--end           --filterfields  --help          --latest        --pager         --start         -f              -p              c               list                                                                    
--fail          --format        --helpfilter    --no-header     --pass          --terse         -h              -s              clear           sm            
```

### Version

devel

### Relevant log output

_No response_

### Post question in Slack

prathmesh4321:
Hi @shahzebsiddiqui . I have fixed the issue with tab completion. Please review the PR.

Screen Shot 2023-03-31 at 1 20 50 PM

### DIFF
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -128,11 +128,33 @@ _buildtest ()
 {
   local cur="${COMP_WORDS[COMP_CWORD]}"
   local prev="${COMP_WORDS[COMP_CWORD-1]}"
+  local next=
+  declare -i offset="0"
 
   COMPREPLY=()   # Array variable storing the possible completions.
 
-  next=${COMP_WORDS[1]}
+  declare -a buildtest_opts=("--color" "--config" "--debug" "--editor" "--help" "--helpcolor" "--logpath" "--loglevel" "--print-log" "--no-color" "--report" "--version" "--view-log" "-c" "-d" "-h" "-l" "-p" "-r" "-V") 
+  
+  for command in "${COMP_WORDS[@]}"
+  do
+    for element in "${buildtest_opts[@]}"
+    do
 
+        if [[ "$command" == "$element" ]]; then
+          
+          if [[ "$command" == "--color" ]] || [[ "$command" == "--config" ]] || [[ "$command" == "-c" ]] || [[ "$command" == "--report" ]] || [[ "$command" == "-r" ]];
+          then
+            ((offset+=2))
+          else
+            ((offset+=1))
+          fi
+        fi
+
+    done
+
+  done
+
+  next=${COMP_WORDS[1+offset]}
 
   case "$next" in
   #case "${prev}" in
@@ -205,7 +227,7 @@ _buildtest ()
         COMPREPLY=( $( compgen -W "$(_avail_report_formatfields)" -- $cur ) )
         return
       esac
-      case "$prev" in summary)
+      case "$prev" in summary|sm)
         local opts="-d -h --detailed --help"
         COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
         return
@@ -218,7 +240,7 @@ _buildtest ()
       COMPREPLY=( $( compgen -W "${cmds}" -- $cur ) )
       # handle completion logic for 'buildtest config <subcommand>' based on subcommands
 
-      case "${COMP_WORDS[COMP_CWORD-2]}" in
+      case "${COMP_WORDS[2+offset]}" in
         compilers|co)
           local opts="--help --json --yaml -h -j -y find test"
           COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
@@ -226,7 +248,7 @@ _buildtest ()
             local opts="--detailed --help --modulepath --update -d -h -m -u"
             COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
           fi
-          if [[ "${prev}" == "test"  ]]; then
+          if [[ "${prev}" == "test" ]]; then
             COMPREPLY=( $( compgen -W "$(_avail_compilers)" -- $cur ) )
           fi
           ;;
@@ -253,7 +275,7 @@ _buildtest ()
       COMPREPLY=( $( compgen -W "${cmds}" -- $cur ) )
 
       # case statement to handle completion for buildtest inspect [name|id|list] command
-      case "${COMP_WORDS[2]}" in
+      case "${COMP_WORDS[2+offset]}" in
         list|l)
           local opts="--builder --help --no-header --pager --terse -b -h -n -t"
           COMPREPLY=( $( compgen -W "${opts}" -- $cur ) );;
@@ -292,9 +314,9 @@ _buildtest ()
       COMPREPLY=( $( compgen -W "${cmds}" -- $cur ) )
 
       # switch based on 2nd word 'buildtest buildspec <subcommand>'
-      case ${COMP_WORDS[2]} in
+      case ${COMP_WORDS[2+offset]} in
       find|f)
-         case ${COMP_WORDS[3]} in
+         case ${COMP_WORDS[3+offset]} in
          # completion for 'buildtest buildspec find invalid'
          invalid)
            local opts="--error --help -e -h"
@@ -318,7 +340,7 @@ _buildtest ()
          esac
         ;;
       summary|sm)
-         case ${COMP_WORDS[3]} in
+         case ${COMP_WORDS[3+offset]} in
          # completion for rest of arguments
          *)
            local longopts="--help --pager"
@@ -358,7 +380,7 @@ _buildtest ()
         local opts="--breakdown --list --help --terse --no-header -b -h -l -n find"
         COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
 
-        case ${COMP_WORDS[3]} in
+        case ${COMP_WORDS[3+offset]} in
         find)
           COMPREPLY=( $( compgen -W "$(_avail_maintainers)" -- $cur ) );;
         esac
@@ -383,7 +405,7 @@ _buildtest ()
       local cmds="--help --pager -h list query"
       COMPREPLY=( $( compgen -W "${cmds}" -- $cur ) )
 
-      case ${COMP_WORDS[2]} in
+      case ${COMP_WORDS[2+offset]} in
       list)
         local opts="--help --no-header --terse -h -n -t"
         COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
--- a/bash_completion.sh
+++ b/bash_completion.sh
@@ -156,9 +156,7 @@ _buildtest ()
   done
 
   local next=${COMP_WORDS[1+offset]}
-  local offset_by_two=2+offset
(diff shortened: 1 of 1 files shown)


bash_completion.sh
```diff
@@ -128,11 +128,32 @@ _buildtest ()
...
+  declare -a buildtest_opts=("--color" "--config" "--debug" "--editor" "--help" "--helpcolor" "--logpath" "--loglevel" "--print-log" "--no-color" "--report" "--version" "--view-log" "-c" "-d" "-h" "-l" "-p" "-r" "-V") 
+  
+  for command in "${COMP_WORDS[@]}"
+  do
+    for element in "${buildtest_opts[@]}"
+    do
+
+        if [[ "$command" == "$element" ]]; then
+          
+          if [[ "$command" == "--color" ]] || [[ "$command" == "--config" ]] || [[ "$command" == "-c" ]] || [[ "$command" == "--report" ]] || [[ "$command" == "-r" ]];
```
shahzebsiddiqui: i think i understand what this is trying to do its catching the edge case where we are offset by 2 due to input argument

```
buildtest --color <COLOR> <sub-command>
buildtest --report <report_file> sub-command>
```

There are few other edge-cases that haven't been accounted for, in this one we could have more offset

```
buildtest -d --loglevel <LOG_LEVEL> <sub-command>
buildtest --editor <editor> bc edit
```

shahzebsiddiqui: tab completion works for `--loglevel` 
Screen Shot 2023-04-07 at 10 34 19 AM

But this doesn't work in conjuction with a sub-command
Screen Shot 2023-04-07 at 10 34 45 AM

prathmesh4321: Thanks @shahzebsiddiqui for catching that. I have added the edge cases now.

Screen Shot 2023-04-07 at 9 38 00 AM

bash_completion.sh
```diff
@@ -128,11 +128,32 @@ _buildtest ()
...
+  declare -a buildtest_opts=("--color" "--config" "--debug" "--editor" "--help" "--helpcolor" "--logpath" "--loglevel" "--print-log" "--no-color" "--report" "--version" "--view-log" "-c" "-d" "-h" "-l" "-p" "-r" "-V") 
+  
+  for command in "${COMP_WORDS[@]}"
+  do
+    for element in "${buildtest_opts[@]}"
+    do
+
+        if [[ "$command" == "$element" ]]; then
+          
+          if [[ "$command" == "--color" ]] || [[ "$command" == "--config" ]] || [[ "$command" == "-c" ]] || [[ "$command" == "--report" ]] || [[ "$command" == "-r" ]] || [[ "$command" == "--loglevel" ]] || [[ "$command" == "-l" ]] || [[ "$command" == "--editor" ]];
```
shahzebsiddiqui: this long if condition looks messy, i believe you can do a check for variable in a list of values something like this. I am not sure if this will work.


see https://techieroop.com/check-if-array-contains-value-in-shell-script/

I just got this code-snippet from chatGPT. 


```bash
commands=( "--color" "--config" "-c" "--report" "-r" "--loglevel" "-l" "--editor" )

if [[ "${commands[*]}" =~ "$command" ]]; then
  # code to be executed if command is found in the list
fi
```

shahzebsiddiqui: here is another rewrite of the code from chatGPT it even smaller though i am not sure if the array assignment make sense at first until you read the code-base later on

```bash
declare -A buildtest_opts=(
  ["--color"]=2
  ["--config"]=2
  ["--debug"]=1
  ["--editor"]=2
  ["--help"]=1
  ["--helpcolor"]=1
  ["--logpath"]=2
  ["--loglevel"]=2
  ["--print-log"]=1
  ["--no-color"]=1
  ["--report"]=2
  ["--version"]=1
  ["--view-log"]=1
  ["-c"]=2
  ["-d"]=1
  ["-h"]=1
  ["-l"]=2
  ["-p"]=1
  ["-r"]=2
  ["-V"]=1
)

offset=0
for command in "${COMP_WORDS[@]}"
do
  if [[ -n "${buildtest_opts[$command]}" ]]; then
    ((offset+=buildtest_opts[$command]))
  else
    ((offset+=1))
  fi
done
```

prathmesh4321: @shahzebsiddiqui Yeah, I had tried with `declare -A` before, it is for associative arrays but it only works for bash version 4.

prathmesh4321: Replacing the long if condition with a check for variable in a list works. 

Screen Shot 2023-04-11 at 1 10 05 PM

bash_completion.sh
```diff
@@ -128,11 +128,34 @@ _buildtest ()
...
   local cur="${COMP_WORDS[COMP_CWORD]}"
   local prev="${COMP_WORDS[COMP_CWORD-1]}"
+  declare -i offset="0"
 
   COMPREPLY=()   # Array variable storing the possible completions.
 
-  next=${COMP_WORDS[1]}
+  declare -a buildtest_opts=("--color" "--config" "--debug" "--editor" "--help" "--helpcolor" "--logpath" "--loglevel" "--print-log" "--no-color" "--report" "--version" "--view-log" "-c" "-d" "-h" "-l" "-p" "-r" "-V")
 
+  commands=( "--color" "--config" "-c" "--report" "-r" "--loglevel" "-l" "--editor" )   # Array variable storing commands which require an input argument from the user.
```
shahzebsiddiqui: the bash completion doesnt work with `--color` 

Screen Shot 2023-04-14 at 11 56 58 AM

shahzebsiddiqui: can you please test this out with several commands 

- buildtest --color blue bd 
- buildtest --color blue --report python.json bd 
- buildtest --loglevel 
- buildtest --loglevel bd
- buildtest -c /path/to/config bd

shahzebsiddiqui: can you rename these variables it is confusing when using `commands` and then the `for command in ${COMP_WORDS[@]}"`

prathmesh4321: @shahzebsiddiqui Done. I have added bash completion for `--editor` as well.

Screen Shot 2023-04-17 at 2 26 52 PM

bash_completion.sh
```diff
@@ -128,11 +128,36 @@ _buildtest ()
...
+          fi
+        fi
+
+    done
+
+  done
+
+  local next=${COMP_WORDS[1+offset]}
+  local offset_by_two=2+offset
+  local offset_by_three=3+offset
```
shahzebsiddiqui: i find this bit confusing having variables `offset_by_two` and `offset_by_tree` since this can lead to endless amount of local variables.

bash_completion.sh
```diff
@@ -218,15 +243,15 @@ _buildtest ()
       COMPREPLY=( $( compgen -W "${cmds}" -- $cur ) )
       # handle completion logic for 'buildtest config <subcommand>' based on subcommands
 
-      case "${COMP_WORDS[COMP_CWORD-2]}" in
+      case "${COMP_WORDS[offset_by_two]}" in
```
shahzebsiddiqui: looking at the commit diff the `COMP_CWORDS` does get the total words and thats what we set for current pointer and previous.

```bash
local cur="${COMP_WORDS[COMP_CWORD]}"
local prev="${COMP_WORDS[COMP_CWORD-1]}"
```  

I found this https://devmanual.gentoo.org/tasks-reference/completion/index.html that explains about the different variables we can use including COMP_LINE, COMP_POINT. I think that we need to get this logic correct and inspect the variable COMP_WORDS to make sure we index the correct tab completion. 

I apologize for my last suggestion, the variable `offset` makes sense now that i think about it but having `offset_by_two` and `offset_by_three` doesnt make sense. Can you review this logic and try again.

prathmesh4321: @shahzebsiddiqui I had tried earlier using `COMP_CWORD` but getting the index based on the total words will not work in every case. The index would be incorrect if there is another subcommad/option ahead. For instance, doing `COMP_WORDS[COMP_CWORD-1]` would makes sense for `buildtest cg co` it gives the index of `co` option but if we now do `buildtest cg co find` then `COMP_WORDS[COMP_CWORD-1]` would give the index of `find` option instead of the index of `co`. So the code will never enter into any of the parent case statements for previous commands.

shahzebsiddiqui: i understand. I think for now this is a slight improvement from the current setup since we couldn't get bash completion working when top-level options were specified. That being said your latest change which reverted the changes back look better so i am fine with merging this in.

Thanks

PR #1436, opened 2023-04-13
remove codecov from pypi dependency 

shahzebsiddiqui:
@prathmesh4321 and @Xiangs18 just a heads up. I recently noticed `codecov` is not available in pypi so i reported an issue https://github.com/codecov/uploader/issues/1019 this was caused by CI failure in https://github.com/buildtesters/buildtest/pull/1435 in particular it was the docs build

Screen Shot 2023-04-13 at 5 27 35 PM

So i am going to remove this apparently if you have a python environment setup with `codecov` you may still have the binary 

```
(buildtest)  ~/Documents/github/buildtest/ [remove_codecov_from_deps] pip list | grep codecov 
codecov                        2.1.12
```

I think they want us to use the new uploader described in https://docs.codecov.com/docs/codecov-uploader which is to download via `curl`.

### DIFF
--- a/.gitlab/ascent.yml
+++ b/.gitlab/ascent.yml
@@ -27,7 +27,9 @@ ascent_pr_regression_test:
     - pip install -r docs/requirements.txt
     - python $BUILDTEST_ROOT/buildtest/tools/unittests.py -c
     - echo $?
+    - curl -Os https://uploader.codecov.io/latest/linux/codecov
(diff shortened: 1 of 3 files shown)


shahzebsiddiqui:
looks like the docs were just built https://readthedocs.org/projects/buildtest/builds/20109758/ so this is good sign.

PR #1463, opened 2023-05-04
global option for --row-count

### LINKED ISSUE #1387: global option for --row-count
Rather than adding this as an option to all of your print functions I would consider instead checking for the argument in the main function and than printing the table there since all of the code is the same, but copied to different places.

_Originally posted by @jscook2345 in https://github.com/buildtesters/buildtest/issues/1377#issuecomment-1426191362_

Xiangs18:
Screen Shot 2023-05-04 at 12 59 58 AM


tests/cli/test_buildspec.py
```diff
@@ -231,6 +232,13 @@ def test_buildspec_find_invalid():
     except SystemExit:
         pass
 
+    try:
+        cache.print_invalid_buildspecs(
+            error=False, terse=False, header=False, row_count=True
+        )
+    except SystemExit:
+        pass
+
```
shahzebsiddiqui: i think you can replace this section with. Its the same thing but using `try/except` with a `pass` is not a good practice. We know this method will raise `sys.exit`. I do see other parts of this code-base uses `try/except` which could also be changed but thats not relevant for this PR.

```
with pytest.raises (SystemExit):
           cache.print_invalid_buildspecs(
            error=False, terse=False, header=False, row_count=True
        )
```

shahzebsiddiqui (changes requested):
okay this looks pretty good. I just have one suggestion for changing the `try/except` with the `with`.

shahzebsiddiqui:
thanks @Xiangs18 this looks good. Great job!

PR #1466, opened 2023-05-05
refactor codebase of 'buildtest inspect' 

shahzebsiddiqui:


shahzebsiddiqui:
@Xiangs18 based on your last PR #1387 which turned out we needed to refactor some of the code-base. I did this basically removing the entry method `inspect_cmd` and invoke the method directly. This caused issues with regression tests as arguments had to be passed over to several methods such as `inspect_cmd` -> `inspect_list` just to invoke a certain test. This also avoids having us to define `args`.

PR #1528, opened 2023-07-07
changing black version in github workflow

### LINKED ISSUE #1526: change black version in github workflow
the black version we are using is https://github.com/buildtesters/buildtest/blob/1bc8c895b4be13308422898b13d8b954b13132d9/requirements.txt#L12

Also the one in pre-commit is 

https://github.com/buildtesters/buildtest/blob/1bc8c895b4be13308422898b13d8b954b13132d9/.pre-commit-config.yaml#L13-L17

But the one used in github workflow is https://github.com/buildtesters/buildtest/blob/1bc8c895b4be13308422898b13d8b954b13132d9/.github/workflows/style.yml#L43

We should change the version in github workflow. See https://black.readthedocs.io/en/stable/integrations/github_actions.html#usage for more detail i think we should still continue with hardcoded version as pose to using `@stable`.

kingxleyy:
I changed the version in github workflow. from "~= 22.0" to "==23.3.0"


requirements.txt
```diff
@@ -8,7 +8,7 @@ rich
 coverage
 pytest
 isort
-black==23.3.0
```
shahzebsiddiqui, line 11: please revert this change back to 23.3.0

hello.txt
```diff
@@ -0,0 +1 @@
+Hello
```
shahzebsiddiqui: remove this file

requirements.txt
shahzebsiddiqui: @kingxleyy i didnt ask you to delete the file i asked to simply revert the change in `requirements.txt` i **dont** want you to update the black version in requirements.txt. Please try again or simply open another PR.

PR #1536, opened 2023-07-11
updating the github version workflow to 23.3.0

### LINKED ISSUE #1526: change black version in github workflow

kingxleyy:
I changed the version in github workflow.  to "==22.0"


shahzebsiddiqui:
Thanks @kingxleyy this looks good!

PR #1557, opened 2023-07-16
Add Coverage to check runtime status

### LINKED ISSUE #1449: Add coverage for runtime status check.

Add coverage in the runtime status check for the case when only max runtime is specified and for the case when both max and min runtime are specified in the buildspec.
See: https://app.codecov.io/gh/buildtesters/buildtest/blob/devel/buildtest/buildsystem/checks.py

Screen Shot 2023-04-19 at 11 38 58 PM

_No response_

### Additional Information

_No response_

### Post question in Slack

szuananwar:
This test will add coverage in the runtime status check for the case when only max runtime is specified and for the case when both max and min runtime are specified in the buildspec.


tutorials/test_status/runtime_status_test.yml
```diff
@@ -49,3 +49,24 @@ buildspecs:
...
+  timelimit_min_max_specified:
+    type: script
+    executor: generic.local.sh
+    description: "Run a sleep job for 2 seconds and test pass if it's within the specified minimum and maximum time"
+    tags: ["tutorials"]
+    run: sleep 2
+    status:
+      runtime:
+        min: 1.5
+        max: 3.5
```
shahzebsiddiqui: i dont understand why we have these two tests. The above 5 test cover this already and can be used as part of the regression test. Please remove these tests unless i am missing something. 

Also if test needs to be added to buildspec for regression test we should not add them to the file where we created symbolic link because this will affect the user documentation. Instead create a separate test in the same directory where test file is.

szuananwar: Noted, thanks!

szuananwar: @shahzebsiddiqui fixed.

PR #1569, opened 2023-07-20
Added support for "spack load" in spack schema

### LINKED ISSUE #1533: [FEATURE]: Ability to load specs in spack schema

We should add support for `spack load` this can be useful for running tests or simply loading the package once its installed. We should add the following. Let's say if we were to test `m4` we can load the package prior to testing it. This could be done as follows

```yaml

buildspecs:
  spack_load_example:
    type: spack
    executor: generic.local.bash
    description: "Run spack load for m4 package and run tests"
    tags: [spack]
    spack:
      root: $HOME/spack
      load:
        options: '--only package'
        specs: [m4]
      test:
        run:
          specs: ['m4']
        results:
          option: "-l"
```

In the above test, the `load` refers to `spack load` and `options` refer to options that can be passed to `spack load` command with a list of specs defined by `specs` property. The below command would translate to `spack load --only package m4`  

```yaml
      load:
        options: '--only package'
        specs: [m4]
```

The order of where `spack load` command must be written in test must be done after `spack install` (if present) and before `spack test` (if present).

# TODO

- [ ] Update [spack schema](https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/spack.schema.json) to add support for `load` 
- [ ] Add implementation of emitting `spack load` command in https://github.com/buildtesters/buildtest/blob/devel/buildtest/builders/spack.py. This would be in method `generate_script` see https://github.com/buildtesters/buildtest/blob/58f91fdc1c1d909d12166daefd642c3ddb78c045/buildtest/builders/spack.py#L48
- [ ] Add example buildspec to demonstrate this example in https://github.com/buildtesters/buildtest/tree/devel/examples/spack
- [ ] Add valid and invalid example for testing with regression test see https://github.com/buildtesters/buildtest/tree/devel/buildtest/schemas/examples/spack.schema.json please put a valid and invalid example so we can test this works during regression test
- [ ] Add documentation in https://buildtest.readthedocs.io/en/devel/buildspecs/spack.html add this at bottom of file

_No response_

### Additional Information

_No response_

### Post question in Slack

Mariamajib:
- [x] Update [spack schema](https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/spack.schema.json) to add support for load
- [x] Add implementation of emitting spack load command in https://github.com/buildtesters/buildtest/blob/devel/buildtest/builders/spack.py. 
- [x] Add example buildspec to demonstrate this example in https://github.com/buildtesters/buildtest/tree/devel/examples/spack
- [x] Add valid and invalid example for testing with regression test see https://github.com/buildtesters/buildtest/tree/devel/buildtest/schemas/examples/spack.schema.json please put a valid and invalid example so we can test this works during regression test
- [x]  Add documentation in https://buildtest.readthedocs.io/en/devel/buildspecs/spack.html add this at bottom of file


buildtest/schemas/examples/spack.schema.json/invalid/spack_load_invalid_example.yml
```diff
@@ -0,0 +1,16 @@
...
+    description: "Run spack load for m4 package and run tests"
+    tags: [spack]
+    spack:
+      root: $HOME/spack
+      load:
+        options: '--only package'
+        specs: [m4]
+      test:
+        run:
+          specs: ['m4', 1]
```
shahzebsiddiqui: remove the `1` from here instead you should put the `1` under `specs` section for `load`. You want to make test invalid and it needs to be something specific to the `load` section.

shahzebsiddiqui:
@Mariamajib so i would like you to add the documentation.

Please add your two commands `buildtest build` and `buildtest inspect query` in this section inorder to autogenerate the two examples https://github.com/buildtesters/buildtest/blob/eeae0183c068cbdaee368fd82f2e6f7fa3cbfbf8/buildtest/tools/docs.py#L69-L94 

Next please get into the container the command to run is `docker run -it -v  $BUILDTEST_ROOT:/home/spack/buildtest ghcr.io/buildtesters/buildtest_spack:latest` but you should check https://buildtest.readthedocs.io/en/devel/contributing/build_documentation.html#generating-documentation-examples-for-buildtest-tutorial

Next run `buildtest tutorial-examples` and then only add the two files that are relevant for the PR. This command will autogenerate the entire documentation so dont add all the files.

Once the files are generated please write the documentation

PR #1581, opened 2023-08-04
Assign Vars value to metrics

### LINKED ISSUE #1427: metrics assignment from variables
we should be able to assign metrics from variables defined in tests. This could look as follows. We can have a property name `vars` which will look for key under the `vars` section of the test. If the key is found we will assign the value to the metric. In this example below we can perform an assertion equality.

```yaml
buildspecs:
  metric_variables:
    executor: generic.local.bash
    type: script
    description: capture metrics from variables
    vars:
       FOO: BAR
    run: echo $FOO
    tags: tutorials
    metrics: 
      foo:
        type: str
        vars: foo 
    status:
      assert_eq:
        - name: foo
          ref: BAR
```

TODO:

- [ ] Update [JSON Schema](https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/definitions.schema.json) to add support for `vars`
- [ ] Add implementation for metrics with vars in https://github.com/buildtesters/buildtest/blob/devel/buildtest/builders/base.py
- [ ] Add example buildspec and regression test for metrics 
- [ ] Add documentation in https://buildtest.readthedocs.io/en/devel/writing_buildspecs/performance_checks.html#defining-metrics.

szuananwar:
This pull is to assign metrics from variables defined in tests. A property name vars is defined with value for vars.

### DIFF
--- a/buildtest/schemas/definitions.schema.json
+++ b/buildtest/schemas/definitions.schema.json
@@ -200,7 +200,7 @@
           "$ref": "#/definitions/regex"
         },
         "vars": {
-          "$ref": "#/definitions/vars"
+          "$ref": "#/definitions/env"
         },
         "file_regex": {
           "$ref": "#/definitions/file_regex_in_metrics"
@@ -221,25 +221,7 @@
       "additionalProperties": {
         "$ref": "#/definitions/metrics_field"
       }
-    },
-    "vars": {
-      "type": "object",
-      "additionalProperties": false,
-      "required": [
-        "value"
-      ],
-      "properties": {
-        "type": {
-          "type": "string",
-          "description": "Specify the type of the variable",
-          "enum": [
-            "str",
-            "int",
-            "float"
-          ]
-        }
-      }
-    },
+    }
     "state": {
       "type": "string",
       "description": "explicitly mark state of test regardless of status calculation",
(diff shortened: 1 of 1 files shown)


buildtest/schemas/definitions.schema.json
```diff
@@ -218,6 +222,24 @@
...
+          "type": "string",
+          "description": "Specify the type of the variable",
+          "enum": [
+            "str",
+            "int",
+            "float"
+          ]
+        }
+      }
+    },
```
shahzebsiddiqui: Please remove this we have a definition for environment variables that work same as variables. This is defined in same file 

```json
    "env": {
      "type": "object",
      "description": "One or more key value pairs for an environment (key=value)",
      "minItems": 1,
      "items": {
        "type": "object",
        "minItems": 1,
        "propertyNames": {
          "pattern": "^[A-Za-z_][A-Za-z0-9_]*$"
        }
      }
    },
```

buildtest/schemas/definitions.schema.json
```diff
@@ -199,13 +199,17 @@
         "regex": {
           "$ref": "#/definitions/regex"
         },
+        "vars": {
+          "$ref": "#/definitions/vars"
```
shahzebsiddiqui: change this to `#/definitions/env` that should fix the issue.

PR #1585, opened 2023-08-08
removing profiles automatically via buildtest config profiles remove

### LINKED ISSUE #1491: [FEATURE]: remove a profile via command line

adding onto feature #1490 we should have a command option `buildtest config profiles remove` that can remove an existing profile from configuration file. This can be useful way to remove a profile as pose to manually editing the file and removing the content.

This can be done with a positional argument which expects the name of profile, so let's say i want to remove `python-prof` i would do  `buildtest config profiles remove python-prof`

We can have an alias for `remove` as `rm` for accessibility. If upon removal there are no profiles we should remove the top-level key `profile` from configuration file. We can have let's say an option `-y` to accept `yes` so user can automatically confirm `yes`. If its not specified then user will be prompted to select an option like `yes/no` to remove profile. 

# TODO

- [ ]  Add command menu in argparse https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/__init__.py
- [ ] Add implementation https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/config.py
- [ ] Add entry in buildtest help config https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/help.py
- [ ] Add option in [bash completion](https://github.com/buildtesters/buildtest/blob/devel/bash_completion.sh) script
- [ ] Add regression test in https://github.com/buildtesters/buildtest/blob/devel/tests/cli/test_config.py
- [ ] Add documentation in https://buildtest.readthedocs.io/en/devel/configuring_buildtest.html

_No response_

### Additional Information

_No response_

### Post question in Slack

kingxleyy:


bash_completion.sh
```diff
@@ -239,7 +239,7 @@ _buildtest ()
     config|cg)
 
 
-      local cmds="-h --help co compilers e edit ex executors p path profiles systems validate v view"
+      local cmds="-h --help co compilers e edit ex executors p path profiles rm remove systems validate v view"
```
shahzebsiddiqui: The `rm` and `remove` does not go here this bash completion is for command `buildtest config` but your command is `buildtest config profiles remove`. Please remove these two words from line 242

bash_completion.sh
```diff
@@ -273,10 +273,10 @@ _buildtest ()
           esac
         ;;
         profiles)
-          local opts="--help -h list"
+          local opts="--help -h --remove -rm list"
```
shahzebsiddiqui: these will be commands so change `--remove` --> `remove` and `--rm` --> `rm`. Also make it alphabetically sorted

bash_completion.sh
```diff
@@ -273,10 +273,10 @@ _buildtest ()
...
           COMPREPLY=( $( compgen -W "${opts}" -- $cur ) )
           if [[ "${prev}" == "list" ]]; then
-            local opts="--help  --theme --yaml -h -y"
+            local opts="--help  --theme --remove --yaml -h -rm -y"
```
shahzebsiddiqui: remove these words `--remove` and `--rm` from here they dont belong in tab completion in `list`

buildtest/cli/__init__.py
```diff
@@ -1020,6 +1020,10 @@ def config_menu(subparsers, parent_parser):
     )
 
     subparsers_config.add_parser("systems", help="List all available systems")
+    #buildtest config profile remove
+    subparsers_profile_list.add_argument(
+        "-y", "--yaml", "rm", "remove", action="store_true", help="List Profile details in YAML Format"
```
shahzebsiddiqui: @kingxleyy this is not correct. Please refer to [add_argument](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument) library for reference you can't pass multiple arguments. 

Please remove `rm` and `remove`. These need to be a subparser similar to list where you can create a command `buildtest config profiles remove`.

This command should accept a positional argument that is the name of profile so if you want to delete a profile name `python` you can do

```
buildtest config profiles rm python
```

It should technically work with multiple arguments where you can remove two profiles (profile1, profile2)

```
buildtest config profiles rm profile1 profile2
```

buildtest/cli/config.py
```diff
@@ -28,6 +28,10 @@ def config_cmd(args, configuration, editor, system):
         if args.profiles == "list":
             list_profiles(configuration, theme=args.theme, print_yaml=args.yaml)
 
+    elif args.config in ["remove", "rm"]:
+        if args.profiles == "list":
+            remove_profiles(configuration, theme=args.theme, print_yaml=args.yaml)
```
shahzebsiddiqui: Remove the `elif` statement 

The `args.profile` should contain value of `remove` or `rm` if you print `args`. That being said you should have the following section 

```python
    elif args.config in ["profiles"]:
        if args.profiles == "list":
            list_profiles(configuration, theme=args.theme, print_yaml=args.yaml)
         elif args.profiles in ['remove', 'rm']
            remove_profiles(configuration, ...)
```

Also please note that method `remove_profiles` is suppose to remove profiles from configuration file so there is no need for `theme` and `print_yaml`

buildtest/cli/config.py
```diff
@@ -183,7 +187,20 @@ def list_profiles(configuration, theme=None, print_yaml=None):
     for profile_name in configuration.target_config["profiles"].keys():
         print(profile_name)
 
+def remove_profiles(configuration, theme=None, print_yaml=None):
```
shahzebsiddiqui: please remove `theme` and `print_yaml` method they are not required

buildtest/cli/config.py
```diff
@@ -183,7 +187,20 @@ def list_profiles(configuration, theme=None, print_yaml=None):
...
+    if print_yaml:
+        profile_configuration = yaml.dump(
+            configuration.target_config["remove"], indent=2
+        )
+        syntax = Syntax(profile_configuration, "yaml", theme=theme or "monokai")
+        console.print(syntax)
+        return
+
+    for profile_name in configuration.target_config["profiles remove"].keys():
+        print(profile_name)
```
shahzebsiddiqui: This whole section is wrong, you just copied the code for listing the profiles

I want you to do a lookup into the `configuration` dictionary which can be done via `configuration.target_config` and find the profile.

For example if we have a profile name `pass` that we want to remove if one runs `buildtest config profiles rm pass` then we need to remove the section `pass`

```yaml
    profiles:
      pass:
        tags:
        - pass
        testdir: /Users/siddiq90/Documents/github/buildtest/var/tests
```

You should only remove the profile if there is a match, if no match then you can print a message saying something along the lines `Unable to remove profile <profile>`

Ultimately i want you to do update the buildtest configuration file and write this back to same file. 

You will want to look at https://github.com/buildtesters/buildtest/blob/d0aff023be7a10c349bbb3bc6c63efcdc04d24b5/buildtest/cli/compilers.py#L122-L188 which is the implementation for `buildtest config compilers find` which shows how configuration file is updated once updated.

The main section that is important for writing the content back is

```python
    # if --update is specified we update existing configuration file and write backup in same directory
    if update:
        fname = (
            "buildtest_"
            + datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
            + ".yml"
        )
        backup_file = resolved_filepath or os.path.join(
            os.path.dirname(configuration.file), fname
        )
        copyfile(configuration.file, backup_file)
        print("Writing backup configuration file to: ", backup_file)
        print(f"Updating configuration file: {configuration.file}")
        with open(configuration.file, "w") as fd:
            yaml.safe_dump(
                configuration.config, fd, default_flow_style=False, sort_keys=False
            )

```

tests/cli/test_config.py
```diff
@@ -78,6 +79,16 @@ def test_config_profile():
...
+@pytest.mark.cli
+def test_config_profile_remove():
+    # buildtest config profiles remove
+    remove_profiles(configuration)
+
+    # buildtest config profiles --theme emacs --yaml --rm
+    remove_profiles(configuration, theme="emacs", print_yaml=True)
+
+    # buildtest config profiles --yaml
+    remove_profiles(configuration, print_yaml=True)
```
shahzebsiddiqui: Please remove these

tests/cli/test_config.py
```diff
@@ -78,6 +79,16 @@ def test_config_profile():
     # buildtest config profiles --yaml
     list_profiles(configuration, print_yaml=True)
 
+@pytest.mark.cli
+def test_config_profile_remove():
+    # buildtest config profiles remove
+    remove_profiles(configuration)
```
shahzebsiddiqui: You will want to create a dummy profile first before removing it. The `remove_profiles` should take an input profile.  

The method signature should be something like

```
def remove_profiles(configuration, names)
```

The `names` should be a list of profile names to remove.

Basically you will want to do a for loop over all profiles and remove them from the configuration if they are found.

tests/cli/test_config.py
kingxleyy: done

bash_completion.sh
kingxleyy: done

bash_completion.sh
kingxleyy: done

bash_completion.sh
kingxleyy: done

buildtest/cli/config.py
kingxleyy: done

buildtest/cli/config.py
kingxleyy: done

tests/cli/test_config.py
```diff
@@ -80,16 +80,10 @@ def test_config_profile():
     list_profiles(configuration, print_yaml=True)
 
 @pytest.mark.cli
-def test_config_profile_remove():
+def remove_profiles(configuration, names):
```
shahzebsiddiqui: This is a regression test method it needs to start with `test_` so name it `test_profiles_remove`. 

Remove the arguments it should simply be 

```python
def test_profiles_remove()
```

bash_completion.sh
```diff
@@ -239,7 +239,7 @@ _buildtest ()
     config|cg)
 
 
-      local cmds="-h --help co compilers e edit ex executors p path profiles systems validate v view"
```
shahzebsiddiqui, line 242: please get rid of `remove` from this line. This is tab completion for `buildtest config` command. This command is under `buildtest config profiles remove`

bash_completion.sh
```diff
@@ -278,7 +278,7 @@ _buildtest ()
           esac
         ;;
         profiles)
-          local opts="--help -h list"
```
shahzebsiddiqui, line 289: please sort this alphabetically. So move `list` before `remove`. Keep the options as is. They need to be declared first

buildtest/cli/__init__.py
```diff
@@ -1020,6 +1020,10 @@ def config_menu(subparsers, parent_parser):
     )
 
     subparsers_config.add_parser("systems", help="List all available systems")
+    #buildtest config profile remove
+    subparsers_profile_list.add_argument(
+        "-y", "--yaml", action="store_true", help="List Profile details in YAML Format"
+    )
```
shahzebsiddiqui: why are you declaring this. You already have this declared above in lines 1006. You need to declare a subparser `buildtest config profiles remove` and this code is not doing that. Please look at original issue and make sure you get this working correctly.

PR #1576, opened 2023-07-25
Remove short option "-t" for "--theme"

braeeast2001:
To complete this issue the following will be done:
- [x] Update [bash completion](https://github.com/buildtesters/buildtest/blob/devel/bash_completion.sh) script and remove -t option wherever its used for --theme
- [x] Remove -t from documentation wherever its mentioned for --theme also check all .. command-output where its -t and replace with --theme
- [x] Update any docstring or comment throughout the code base

PR #1599, opened 2023-08-25
Add short option -t for  --testpath

### LINKED ISSUE #1572: short option for 'buildtest inspect query'
Once https://github.com/buildtesters/buildtest/issues/1571 is addressed let's add short option `-t` for `--testpath

Let's make this change in the following

- [x] [bash completion](https://github.com/buildtesters/buildtest/blob/devel/bash_completion.sh) script
- [x] argparse https://github.com/buildtesters/buildtest/blob/devel/buildtest/cli/__init__.py
- [x] in https://github.com/buildtesters/buildtest/blob/devel/buildtest/tools/docs.py change `--testpath` to `-t`

shahzebsiddiqui:


jscook2345:
For documentation, I like to use long options so the user knows better what the option means, but I think this is fine to.

shahzebsiddiqui:
yeah i think either way works. This PR was addressing an issue where we had conflict with short option `-t` for `--theme` that was removed in https://github.com/buildtesters/buildtest/pull/1576 so we reintroduced the `-t` for `--testpath` again. For consistency yeah we gotta have docs use long options and that would be a significant rewrite to docs. It may be a good thing. For now i used short option considering we were using `-o` for `--output`

PR #1601, opened 2023-08-28
buildtest regression test run in unique directory for every CI run

shahzebsiddiqui:


jscook2345:
@shahzebsiddiqui It looks like the checks did not all pass due to the link checker. Can you fix that?

shahzebsiddiqui:
> @shahzebsiddiqui It looks like the checks did not all pass due to the link checker. Can you fix that?

thanks @jscook2345 yeah 2 of them are cdash will need to fix these apparently cdash doesn't preserve old builds i think they delete content after a while so we have this broken link. Either i exclude this url pattern from check that might be better approach since we will face this issue again.

jscook2345:
Yeah I would exclude the pattern or just don't use cdash url's if you can avoid it

PR #1635, opened 2023-09-29
adding regression test workflow in container

shahzebsiddiqui:


Xiangs18:
@shahzebsiddiqui Regression test is at least running in container now. Still need to debug on failing tests if any.

shahzebsiddiqui:
> Details



> @shahzebsiddiqui Regression test is at least running in container now. Still need to debug on failing tests if any.

thanks for getting this running for now, we should try getting checkout to work inside the container. I also merged #1789 which should address issue with tcsh, zsh not being in container. I recently added a PR to install pip automatically #1792

Xiangs18:
Hi @shahzebsiddiqui, I added check repo and now is running as expected. I downgraded from v4 to v3 to avoid `version GLIBC_2.28' not found` bug.

Still need to fix some failing tests but that is out of scope of this PR. Also why do we need tmate session step?

Let's remove `tmate` step

.github/workflows/regressiontest_container.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  pull_request:
+    branches: [ devel ]
+    
+jobs:
+  buildtest_tutorial_container:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/buildtesters/buildtest_spack
```
shahzebsiddiqui: can you try `ghcr.io/buildtesters/buildtest_spack:spack-sc23` for whatever reason its not finding `csh` and it should. You can try running `which csh zsh` prior to running regtest for debugging purposes

Xiangs18: is kinda strange that `csh` is not detected as one of system shells. I have to manually add `/bin/csh` into valid_shells for that test to pass. Something is wrong.

shahzebsiddiqui: its in the container i just checked see this line https://github.com/buildtesters/buildtest/actions/runs/9692887989/job/26747175463#step:4:17

shahzebsiddiqui: its here

Screenshot 2024-06-28 at 11 50 47 AM

Xiangs18: I know, but the test failed: https://github.com/buildtesters/buildtest/actions/runs/9692887989/job/26747175463#step:4:164

shahzebsiddiqui: ur right this is out of scope we will need to fix this for future. I will merge this in.

It appears that `/usr/bin/csh` is not available shell for whatever reason we are `cat /etc/shells` which contains a list of shell. Dont know why its installed as `/usr/bin/csh` there is typically a symlink to `/bin/csh`.

shahzebsiddiqui:
thanks @Xiangs18 for working on this.

PR #1735, opened 2024-03-22
Extract line number in metrics declaration

### LINKED ISSUE #1725: [FEATURE]: extract line number in metrics declaration

Buildtest should have ability to extract line from stdout/stderr that can be assigned to a metric which can be used later for status check. For instance let's take the following results from OSU bandwidth test result and we want to extract a particular line 

```
# OSU MPI Bandwidth Test v7.3                                                                                                                                                                                                                                              
# Size      Bandwidth (MB/s)                                                                                                                                                                                                                                               
# Datatype: MPI_CHAR.                                                                                                                                                                                                                                                      
1                       5.65                                                                                                                                                                                                                                               
2                       3.57                                                                                                                                                                                                                                               
4                       9.26                                                                                                                                                                                                                                               
8                      33.72                                                                                                                                                                                                                                               
16                     20.76                                                                                                                                                                                                                                               
32                    161.53                                                                                                                                                                                                                                               
64                    133.98                                                                                                                                                                                                                                               
128                   705.10                                                                                                                                                                                                                                               
256                   534.20                                                                                                                                                                                                                                               
512                   708.01                                                                                                                                                                                                                                               
1024                  390.97                                                                                                                                                                                                                                               
2048                  899.16                                                                                                                                                                                                                                               
4096                 5240.26                                                                                                                                                                                                                                               
8192                 3275.13                                                                                                                                                                                                                                               
16384                4527.77                                                                                                                                                                                                                                               
32768                6407.88                                                                                                                                                                                                                                               
65536                9635.86                                                                                                                                                                                                                                               
131072              15361.67                                                                                                                                                                                                                                               
262144              10873.81                                                                                                                                                                                                                                               
524288              18213.85                                                                                                                                                                                                                                               
1048576             10275.43                                                                                                                                                                                                                                               
2097152             11507.50                                                                                                                                                                                                                                               
4194304              4837.92     
```

At the moment we have support for regex in metrics that can be used as comparison which works fine.

```yaml
buildspecs:
  osu_bandwidth_perf:
    type: script
    executor: generic.local.bash
    description: Run OSU Bandwidth Performance Test
    run:  mpirun -np 2 osu_bw
    metrics:
      osu_bw:
        type: float
        regex:
          exp: '^16384\s+([\d.]+)$'
          stream: stdout
          item: 1
    status:
      assert_ge:
        comparisons:
          - name: osu_bw
            ref: 10000
```

This regex is applied on entire content of file and not on a particular line. 

For this feature, we would like to add a key something like `linenum` that will be used to extract a particular line where regex is applied. To extend this example we can simply do the following

```yaml
buildspecs:
  osu_bandwidth_perf:
    type: script
    executor: generic.local.bash
    description: Run OSU Bandwidth Performance Test
    run:  mpirun -np 2 osu_bw
    metrics:
      osu_bw:
        type: float
        regex:
          exp: '^16384\s+([\d.]+)$'
          stream: stdout
          linenum: 10
          item: 1
    status:
      assert_ge:
        comparisons:
          - name: osu_bw
            ref: 10000
```

The `linenum` must be an integer and it can be any value including negative values. The idea is that we simply extract the line from file and apply regex on the specific line. `linenum: -1` would mean last line of file and `linenum: 0` would be first line of file.

_No response_

### Additional Information

_No response_

### Post question in Slack

Xiangs18:


buildtest/builders/base.py
```diff
@@ -930,6 +930,23 @@ def _get_variables(self, variables):
...
+        Args:
+            regex (dict): regex section from the metrics field
+        """
+
+        stream = regex.get("stream")
+        content = self._output if stream == "stdout" else self._error
+
+        linenum = regex.get("linenum")
+        if linenum is not None and content:
+            split = content.split("\n")
```
shahzebsiddiqui: this is confusing please dont use variable `split` when doing `content.split("\n")`

instead we should use a more meaningful variable such as `line`

tests/builders/metrics_regex_with_linenum.yml
```diff
@@ -0,0 +1,29 @@
...
+        regex:
+          exp: '(\d+\.\d+)$'
+          stream: stdout
+          linenum: 0
+      hpcg_state_stream:
+        type: str
+        regex:
+          exp: '(SUPERB)'
+          stream: stdout
+          linenum: -1
```
shahzebsiddiqui: this example is not very apparent since we are using `re.search` as the default if you simply remove `linenum` you will still get a match.

Since re.search will search in entire string i think metrics will still be defined. 

I would suggest you have some duplicate perhaps the content is something like

```
This is line: 1
This is line: 2
This is line: 3
This is line: 4
```

Then apply a regular expression to and specify the line number.

To showcase line number just do something like following, because the expression `This is a line` will just find the first occurence but using `linenum` you will get the specific line. The metrics `second_line` should get the string `This is a linenum: 1`

```yaml
metrics:
  second_line:
    type: str
    regex:
      exp: 'This is a line:'
      stream: stdout
      linenum: 1
  without_linenum:
    type: str
    regex:
      exp: 'This is a line:'
      stream: stdout
```

You should also test when applying regex to `stderr` in this case there is nothing written to stderr so metric should be undefined

shahzebsiddiqui: Can you please put this example in https://github.com/buildtesters/buildtest/tree/devel/tutorials/metrics and put a symlink in this location

Xiangs18: 👍

buildtest/builders/base.py
Xiangs18: 👍

buildtest/builders/base.py
```diff
@@ -930,6 +930,26 @@ def _get_variables(self, variables):
...
+        content = self._output if stream == "stdout" else self._error
+
+        linenum = regex.get("linenum")
+        if linenum is not None and content:
+            lines = content.split("\n")
+            try:
+                content = lines[:-1][linenum] if lines[-1] == "" else lines[linenum]
+            except Exception as e:
+                raise ValueError("Failed to index linenum from the content") from e
+        return content
```
shahzebsiddiqui: i dont really follow this logic for `content = lines[-1][linenum] ...`

I think what we need is a `try/except` block to fetch index into `lines` and retrieve, if there is an IndexError then return an empty string since metrics would be undefined.

we should change `Exception` to `IndexError` since that is what would get raised when you access an index.

Also we should not raise `ValueError` since this will prevent the test from running, rather we should set the value to empty string `""`.

That being said i think the code should be something like

```python
try:
  content = lines[linenum]
except IndexError as e
  content = ''
  logger.error(e)
return content
```

shahzebsiddiqui: @Xiangs18 Please try my suggestion and try running the examples to see if it works. The `logger` variable should be accessible, it might be `self.logger` also so try running the invalid example and then check the log output for error message.

shahzebsiddiqui (approved):

PR #1751, opened 2024-04-12
Extract linenum from file_regex for metrics

### LINKED ISSUE #1726: [FEATURE]: extract linenum from file_regex for metrics

This feature is similar to #1725 but for `file_regex` where we want to extract a line number from file instead of stdout/stderr. 

Here is an example of metrics declaration for `file_regex` that will simply write one line to file.

```yaml
buildspecs:
  metric_file_regex:
    executor: generic.local.bash
    type: script
    description: capture result metric from file path
    run: echo "HPCG result is VALID with a GFLOP/s rating of=63.6515" > hpcg.txt
    tags: tutorials
    metrics:
      hpcg_rating_file:
        type: float
        file_regex:
          exp: '(\d+\.\d+)$'
          file: hpcg.txt
      hpcg_state_file:
        type: str
        file_regex:
          exp: '(VALID)'
          file: hpcg.txt
```

We can add a field `linenum` that will extract the line from file and apply regex so we can do the following

```yaml
  metric_file_regex:
    executor: generic.local.bash
    type: script
    description: capture result metric from file path
    run: echo "HPCG result is VALID with a GFLOP/s rating of=63.6515" > hpcg.txt
    tags: tutorials
    metrics:
      hpcg_rating_file:
        type: float
        file_regex:
          exp: '(\d+\.\d+)$'
          linenum: 0
          file: hpcg.txt
      hpcg_state_file:
        type: str
        file_regex:
          exp: '(VALID)'
          file: hpcg.txt
```

If there is an index error due to invalid line number the metrics would be undefined. This would be the default behavior of any metric if regex is not found then it will be not defined.

_No response_

### Additional Information

_No response_

### Post question in Slack

### LINKED ISSUE #1761: [FEATURE]: extract line from file_regex and regex status check

We should be able add support for extracting line when applying regular expression check in status check for `file_regex` and `regex`. 

So take for example the following example, we have introduced `linenum` that can be applied to extract a line from file or stream. In this example we will extract the 1st and last line of `count.txt` which should evaluate to the correct expression which is `1,` and `10` from the file. The `regex` check will pass for first test the expression `'Line\s\d+'` should match any of the lines regardless of the line number we pick

In the next test, we should get a failure since we will extracting last line of stream which is `********` therefore the expression should fail.

```yaml

buildspecs:
  extract_line_from_file_and_regex:
    type: script
    executor: generic.local.bash
    description: "Extract line from file and stream and apply regular expression"
    run: |
      for i in {1..10}; do      
        echo "Line $i"
      done
    status:
      file_regex:
        - file: count.txt
          exp: '^1$'
          linenum: 0
        - file: count.txt
          exp: '^10$'
          linenum: -1
      regex:
        stream: stdout
        exp: 'Line\s\d+'
  extract_line_failure:
    type: script
    executor: generic.local.bash
    description: Regular expression failure due to incorrect line
    run: |
      for i in {1..10}; do
        echo "Line $i" >> count.txt
        echo "Line $i"
      done
      echo "*********"
    status:
      regex:
        stream: stdout
        exp: 'Line\s\d+'
        linenum: -1
```

I ran this example without adding `linenum` to give you idea the 2nd test output should have ****** at end


```
(buildtest) ☁  general_tests [devel] ⚡  buildtest it query -o extract_line_failure 
──────────────────────────────────────────────────────────────────────────────────── extract_line_failure/5e3dbd09-4fb2-4d23-b628-b508108e04c6 ────────────────────────────────────────────────────────────────────────────────────
Executor: generic.local.bash
Description: Regular expression failure due to incorrect line
State: PASS
Returncode: 0
Runtime: 0.012094 sec
Starttime: 2024/04/26 12:45:15
Endtime: 2024/04/26 12:45:15
Command: bash extract_line_failure_build.sh
Test Script: /Users/siddiq90/Documents/buildtest/var/tests/generic.local.bash/a/extract_line_failure/5e3dbd09/extract_line_failure.sh
Build Script: /Users/siddiq90/Documents/buildtest/var/tests/generic.local.bash/a/extract_line_failure/5e3dbd09/extract_line_failure_build.sh
Output File: /Users/siddiq90/Documents/buildtest/var/tests/generic.local.bash/a/extract_line_failure/5e3dbd09/extract_line_failure.out
Error File: /Users/siddiq90/Documents/buildtest/var/tests/generic.local.bash/a/extract_line_failure/5e3dbd09/extract_line_failure.err
Log File: /Users/siddiq90/Documents/buildtest/var/logs/buildtest_t68eidr1.log
───────────────────────────────────────────── Output File: /Users/siddiq90/Documents/buildtest/var/tests/generic.local.bash/a/extract_line_failure/5e3dbd09/extract_line_failure.out ──────────────────────────────────────────────
Line 1                                                                                                                                                                                                                             
Line 2                                                                                                                                                                                                                             
Line 3                                                                                                                                                                                                                             
Line 4                                                                                                                                                                                                                             
Line 5                                                                                                                                                                                                                             
Line 6                                                                                                                                                                                                                             
Line 7                                                                                                                                                                                                                             
Line 8                                                                                                                                                                                                                             
Line 9                                                                                                                                                                                                                             
Line 10                                                                                                                                                                                                                            
*********                                 
```
# TODO
- [ ] Update schema https://github.com/buildtesters/buildtest/blob/devel/buildtest/schemas/definitions.schema.json
- [ ] Add example buildspec in https://github.com/buildtesters/buildtest/tree/devel/tutorials/test_status
- [ ] Add implementation, see methods for `regex_check` and `file_regex` https://github.com/buildtesters/buildtest/blob/140aabd11cac349c139f9d1965298d281cfedfa5/buildtest/buildsystem/checks.py#L97 and https://github.com/buildtesters/buildtest/blob/140aabd11cac349c139f9d1965298d281cfedfa5/buildtest/buildsystem/checks.py#L161. These are called in method `check_test_state` see https://github.com/buildtesters/buildtest/blob/140aabd11cac349c139f9d1965298d281cfedfa5/buildtest/builders/base.py#L1080
- [ ] Add regression test in https://github.com/buildtesters/buildtest/blob/devel/tests/builders/test_builders.py and create symlink for file
- [ ] Add symlink for valid buildspec example in https://github.com/buildtesters/buildtest/tree/devel/buildtest/schemas/examples/script.schema.json/valid
- [ ] Add an invalid test in https://github.com/buildtesters/buildtest/tree/devel/buildtest/schemas/examples/script.schema.json/invalid you can try to specify an invalid type for `linenum` such as specifying a floating point value or using a list so you can do something like `linenum: 1.0` or `linenum: [1, 3]`. Anything that makes the schema fail.   

_No response_

### Additional Information

_No response_

### Post question in Slack

Xiangs18:


Xiangs18:
self note: need to add documentation

buildtest/builders/base.py
```diff
@@ -932,17 +932,15 @@ def _get_variables(self, variables):
...
         Args:
-            regex (dict): regex section from the metrics field
+            regex (dict): regex or file_regex section from the metrics field
+            content (str): content to be extracted from
         """
 
-        stream = regex.get("stream")
-        content = self._output if stream == "stdout" else self._error
-
         linenum = regex.get("linenum")
```
shahzebsiddiqui: i am thinking since `regex` is used once just to check this assignment`linenum = regex.get('linenum')`, why not we do this outside the method. IF the `linenum` property is not defined we just need to set to empty quotes.

tests/builders/metrics_file_regex_with_invalid_linenum.yml
```diff
@@ -0,0 +1 @@
+/Users/sean/Documents/LBNL/code/buildtest/tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
```
shahzebsiddiqui: please fix symlink

tests/builders/metrics_file_regex_with_linenum.yml
```diff
@@ -0,0 +1 @@
+/Users/sean/Documents/LBNL/code/buildtest/tutorials/metrics/metrics_file_regex_with_linenum.yml
```
shahzebsiddiqui: fix symlink

tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
```diff
@@ -0,0 +1,15 @@
+buildspecs:
+  metric_file_regex:
+    executor: generic.local.bash
+    type: script
+    description: capture result metric from file path
+    run: echo -e "HPCG result is INVALID with a GFLOP/s rating of=28.1215\n
```
shahzebsiddiqui: see if this works, i know it works in shell.

You should use the `|` symbol see https://yaml.org/spec/1.2.2/#literal-style which allows one to literal block symbol

```
run: |
  cat <<EOF > hpcg.txt
  HPCG result is INVALID with a GFLOP/s rating of=28.1215 
  HPCG result is VALID with a GFLOP/s rating of=68.9888
  EOF

File: tutorials/metrics/metrics_file_regex_with_linenum.yml [hunk 5]
```diff
@@ -0,0 +1,27 @@
+buildspecs:
+  metric_file_regex:
+    executor: generic.local.bash
+    type: script
+    description: capture result metric from file path
+    run: echo -e "HPCG result is INVALID with a GFLOP/s rating of=28.1215\n
```
shahzebsiddiqui: please change this also to using `run: |`

tutorials/metrics/metrics_file_regex_with_linenum.yml
```diff
@@ -0,0 +1,27 @@
...
+        file_regex:
+          exp: '(\d+\.\d+)$'
+          file: hpcg.txt
+    status:
+      assert_eq:
+        comparisons:
+        - name: second_line
+          ref: 68.9888
+        - name: without_linenum
+          ref: 28.1215
```
shahzebsiddiqui: if i understand both of these checks will pass and the point of this is to show that regex with and without line-number using same regular expression can impact since it finds the first occurence of pattern when using the expression `(\d+\.\d+)`

tutorials/metrics/metrics_file_regex_with_linenum.yml
```diff
@@ -0,0 +1,27 @@
...
+    description: capture result metric from file path
+    run: echo -e "HPCG result is INVALID with a GFLOP/s rating of=28.1215\n
+                  HPCG result is VALID with a GFLOP/s rating of=68.9888" > hpcg.txt
+    tags: tutorials
+    metrics:
+      second_line:
+        type: float
+        file_regex:
+          exp: '(\d+\.\d+)$'
+          linenum: 1
```
shahzebsiddiqui: i wonder if we did `-1` it would be the same as `1` in this example since content of file is only two lines. This could be good way perhaps changing the text `second_line` to `last_line`


0:  first line of file
1:  second line
-1:  last line of file (2nd line)
-2: 2nd last line of file (first line)

shahzebsiddiqui: It would also allow us to explain that negative index also work and you have a duplicate example with `regex` with `linenum` so this might complement well.

tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
shahzebsiddiqui: here you can do it this way

```yaml
run: |
  echo -e "HPCG result is INVALID with a GFLOP/s rating of=28.1215" > hpcg.txt 
  echo -e "HPCG result is VALID with a GFLOP/s rating of=68.9888 >> hpcg.txt
```

buildtest/builders/base.py
shahzebsiddiqui: with this approach we will cut down on number of arguments that will be passed it would be something like this since if `linenum` is not present we dont extract any line and set `content` to empty string and it doesnt need to call the method

```python
if not regex.get('linenum'):
  content = ''
else:
  content = self._extract_content(content_input)
```

tests/builders/metrics_file_regex_with_invalid_linenum.yml
shahzebsiddiqui: you should use relative paths to create symlinks take a look below

```
☁  builders [lsf_queue_check_in_executors] ⚡  ls -l
total 40
-rw-r--r--@ 1 siddiq90  siddiq90     0 Mar 20 14:41 __init__.py
lrwxr-xr-x@ 1 siddiq90  siddiq90    41 Mar 20 14:41 assert_eq.yml -> ../../tutorials/perf_checks/assert_eq.yml
lrwxr-xr-x@ 1 siddiq90  siddiq90    41 Mar 20 14:41 assert_ge.yml -> ../../tutorials/perf_checks/assert_ge.yml
lrwxr-xr-x@ 1 siddiq90  siddiq90    41 Mar 20 14:41 assert_gt.yml -> ../../tutorials/perf_checks/assert_gt.yml
```

tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
```diff
@@ -0,0 +1,15 @@
+buildspecs:
+  metric_file_regex:
```
shahzebsiddiqui: change this test name to `metric_file_regex_with_linenum`

tutorials/metrics/metrics_file_regex_with_linenum.yml
```diff
@@ -0,0 +1,27 @@
...
+          file: hpcg.txt
+      without_linenum:
+        type: float
+        file_regex:
+          exp: '(\d+\.\d+)$'
+          file: hpcg.txt
+    status:
+      assert_eq:
+        comparisons:
+        - name: second_line
```
shahzebsiddiqui: change this to `last_line` if you use `-1`

buildtest/builders/base.py
Xiangs18: ah, I am confused again. If you just pass in content_input, how do you know which linenum to extract.
I must be missing something important.

Xiangs18: Also, it regex.get("linenum") is not None, content = content_input instead of ""

tutorials/metrics/metrics_file_regex_with_linenum.yml
Xiangs18: 👍

tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
Xiangs18: changed both bd names

tutorials/metrics/metrics_file_regex_with_linenum.yml
Xiangs18: 👍

tutorials/metrics/metrics_file_regex_with_linenum.yml
Xiangs18: right, anything you think I need to fix for this comment?

tutorials/metrics/metrics_file_regex_with_linenum.yml
Xiangs18: 👍

tutorials/metrics/metrics_file_regex_with_invalid_linenum.yml
Xiangs18:
```
run: |
  echo -e "HPCG result is INVALID with a GFLOP/s rating of=28.1215" > hpcg.txt 
  echo -e "HPCG result is VALID with a GFLOP/s rating of=68.9888" >> hpcg.txt
```
works

tests/builders/metrics_file_regex_with_linenum.yml
Xiangs18: 👍

tests/builders/metrics_file_regex_with_invalid_linenum.yml
Xiangs18: 👍

buildtest/builders/base.py
```diff
@@ -932,20 +932,20 @@ def _get_variables(self, variables):
...
         if linenum is not None and content:
             lines = content.split("\n")
+            if lines[-1] == "":
```
shahzebsiddiqui: can you please put comments above this line why we are doing this. I know we discussed but it will make it easier to understand code later.

I think if i recall this was used to remove last line since file will end in newline character `/n` so we dont want to view that as the last line rather the actual text.

buildtest/builders/base.py
```diff
@@ -932,20 +932,20 @@ def _get_variables(self, variables):
 
         return lines
 
-    def _extract_content(self, regex):
-        """Extract content based on the stream and linenum properties
-        from the regex field and return it as a string.
+    def _extract_line(self, regex, content):
```
shahzebsiddiqui: okay so i think this method should just take two arguments `linenum` and `content` that all that it needs. Since this is extracted a line from content we need both parameters. This way we can extend this method in future when extending multiple lines. The `regex` is not needed in this method and we can do this outside the method. 

This method would be typically called like 

```python
linenum = regex.get('linenum')
content = self._extract_line(linenum=linenum, content=input_content)
```

buildtest/builders/base.py
shahzebsiddiqui: okay so please take a look at https://github.com/buildtesters/buildtest/pull/1751/files#r1572440374 to help address your confusion. I think we do need 2 arguments for this method which are  `linenum` and `content` and `regex` is not needed. We should compute the linenum outside this method

buildtest/builders/base.py
shahzebsiddiqui: The benefit of computing linenum outside the method is we can call this method multiple times if needed like this



```python
linenums=[3, 6, 9]
for i in linenums:
  content += self._extract_line(linenum=i, content=content)  
```

Xiangs18: 👍

buildtest/builders/base.py
Xiangs18: 👍

buildtest/builders/base.py
Xiangs18: 👍

shahzebsiddiqui (approved):
Thanks @Xiangs18  i think this is ready to merge. I appreciate your help.

PR #1784, opened 2024-06-04
Add option --display to display output and test in buildtest build 

shahzebsiddiqui:


shahzebsiddiqui:
@Xiangs18 fyi i went ahead and implemented the feature in our slack conversation https://hpcbuildtest.slack.com/archives/D06LQLSU0SX/p1716566246526019 and decided to go with `--display` and make this an opt-in feature so default is not to print the output and test to screen since it added too much output to console.

